PR #29, opened 2022-09-06
Ecg tutorial

rene-bucchia:
Batched prior and model saving architecture as discussed on slack


rene-bucchia:
I realized I deleted some of my own changes before, so the version to evaluate for merging really is current state of the branch including the final push I made (18d0611). Sorry about the chaotic timeline of this request.

stefanradev93 (approved):
This is a nice and clean change. We could also think of a helper class in future which takes care of all I/O operations with one-liners.

PR #30, opened 2022-09-11
Ecg tutorial

rene-bucchia:
Exception-handling for KDE difficulties as discussed on slack.


stefanradev93 (changes requested):
We should avoid using `print` in a production-like environment. We should be logging a warning with `repr(e)`.

stefanradev93 (approved):
Nice, thank you!

PR #36, opened 2022-10-14
Fix bug in SBC histogram

vpratz:
The probability for a rank to be in a bin is p = 1 / num_bins, not p = 1 / (num_bins + 1). Probably num_bins was mistaken for L = num_bins - 1 from Talts et. al (2018).

The expected number of ranks per bin is just N / num_bins. As binom.interval gives unequal areas around the median for large intervals, it can not be used to calculate the mean, which is simply N / num_bins.
Demonstration:

```
>>> from scipy.stats import binom
>>> binom.mean(2000, 1/100)
20.0
>>> binom.interval(0.99, 2000, 1/100)
(10.0, 32.0)
```

This has been changed to provide correct CI and mean values.


stefanradev93:
Nice spot!

PR #38, opened 2022-10-27
Added split network: Each subnetwork takes parts of the data, outputs are combined

vpratz:
This is helpful if you have data with different condition and you want separate summary networks for each condition.


stefanradev93:
Nice! `networks.py` is now deprecated. Add new class to `summary_networks.py` instead.

PR #40, opened 2022-11-02
MCMC sampling: Added dosumentation, changed default behavior

vpratz:


stefanradev93:
Nice, thanks for adding the documentation!

PR #43, opened 2022-11-06
Include learning rate in display string

vpratz:


stefanradev93 (approved):
Awesome! Good job! :)

PR #49, opened 2022-11-28
Future

rene-bucchia:
Altered trainer.train_from_presimulation and model.presimulate_and_save to be compatible with defaulting to cosine decay.


stefanradev93:
Refactored as discussed. :)

PR #55, opened 2023-01-20
Standardized code formatting

yannikschaelte:
In software projects, it makes often sense to agree on a common code format and style guide. In Python, [black](https://github.com/psf/black) is an auto-formatter taking care of any such decisions (and bad style such as spaces at end of lines, too long lines, wrong indent, weird line breaks, ...). Here, I added black together with other tools (isort to sort and group imports; nbqa to apply black and isort to notebooks; and some checks). I included them via pre-commit hooks, such that they can be easily run automatically.

The only files I edited manually are ``pyproject.toml`` and (added) ``.pre-commit-hooks.yaml``. The rest was automatically done by running ``pre-commit run --all-files``. By running ``pre-commit install``, one can also instruct the local git clone to always run those style formatters and checks prior to every commit, to make sure things stay consistent. (Additionally, one could add them to github-actions, see [here](https://github.com/ICB-DCM/pyPESTO/blob/main/.github/workflows/ci.yml#L258)).

The only diversion from the black standard is the line-length of 120 I set as many of the lines are currently longer (common in Python projects is 79, as this fits into a normal bash window) -- this is a choice I guess.

**Note**
In my opinion, using an auto-formatter greatly simplifies things. I made this separate branch so you can see how it looks -- feel free to merge or ignore (or reproduce in an own branch) :) As running the auto-formatter changes many lines once, this should probably be done best when there are not too many open branches lying around, to avoid merging issues.


assets/benchmark_network_architectures.py
```diff
@@ -1,135 +1,55 @@
...
-            'num_params': 10,
-            'num_coupling_layers': 2,
-            'coupling_net_settings': COUPLING_SETTINGS_TEST
-        },
-        'likelihood': {
-            'num_params': 10,
-            'num_coupling_layers': 2,
-            'coupling_net_settings': COUPLING_SETTINGS_TEST
-        }
+    "gaussian_linear": {
```
yannikschaelte, line 15: Note: This contraction into 1 line (which is possible due to 120 char limit, probably not with 79) can be avoided by explicitly setting a comma (",") also after the last entry, in which case the previous formatting with a single entry per line will be used.

PR #57, opened 2023-02-03
Minor fixes

arrjon:
Some minor bug fixed and added functionalities

Simulator Class:
- for performing presimulations iterations_per_epoch must always be an integer
- possibility to disable user input while presimulating data (useful for working on a computing cluster)

Trainer Class:
- Remove non-pickle files in default loader (otherwise can lead to error during the training if some other file lies in the folder)
- Allow the possibility to reuse presimulated data during training if specified (as in offline training)


stefanradev93:
Neat!

PR #67, opened 2023-04-11
Bug-Fix while multiprocessing presimulations

arrjon:
While performing presimulations in parallel it can happen that the file `test_batch.pkl` (used to check the batch size) is already overwritten by a parallel process. Adding a timestamp to the file name solved the problem.

Other minor changes are only for readability of the code.

### DIFF
--- a/bayesflow/simulation.py
+++ b/bayesflow/simulation.py
@@ -21,7 +21,7 @@
 import logging
 import os
 import pickle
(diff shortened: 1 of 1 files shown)


stefanradev93:
Nice spot! Integration tests are failing due to typo in the line
`from datatime import datetime` 

which should read: 
`from datetime import datetime`

PR #73, opened 2023-05-23
Update setup.cfg

LuSchumacher:
Remove the m2r2 model from the dependencies because the installation introduced conflicts with the mistune module version


stefanradev93:
Should only change install requirements.

PR #102, opened 2023-09-13
set label names in plot_recovery()

levolz:
qol change to allow to change label names if wanting something other than "ground truth" / "estimated"

### DIFF
--- a/bayesflow/diagnostics.py
+++ b/bayesflow/diagnostics.py
@@ -51,6 +51,8 @@ def plot_recovery(
     color="#8f2727",
     n_col=None,
     n_row=None,
+    xlabel="Ground truth",
+    ylabel="Estimated",
 ):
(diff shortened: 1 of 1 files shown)


stefanradev93:
Pls add new arguments in the docstring of the function too.

stefanradev93:
Pls also pull requests for Development, no direct pulls to master.

levolz:
changed now - re: docstrings, `n_col` & `n_row` are not documented yet

stefanradev93:
Nice. 10x, need to document those too.

levolz:
Wasn't sure about style, though - instead of

`xlabel : 
   XXX
ylabel :
   YYY`

you could (for brevity) do

`xlabel, ylabel :
    XXX, YYY`

stefanradev93:
As per convention, each arg gets its own line with type, default and a new line for description (see existing, looks nice when rendered).

PR #108, opened 2023-11-07
Optimization: add setting for tqdm mininterval

vpratz:
When we do many iterations per second, the load of printing updates with tqdm can become a significant speed factor. By using `mininterval` and setting `refresh=False` in `tqdm.set_postfix_str`, we can control the update rate, thereby reducing the load. For now, I set it as `TQDM_MININTERVAL=0.1` in the settings, but this can be changed according to how smooth we want the updates to be.


stefanradev93:
Great idea, thanks, Valentin!

PR #137, opened 2024-02-23
Make diagnostic plots work with one-parameter models

Kucharssim:
fixes https://github.com/stefanradev93/BayesFlow/issues/135

Just ensuring that we can loop over `ax` even when it's returned by `plt.subplots(nrows=1, ncols=1)`. There is some code duplication between different plots, but I just made a simple fix where needed; it doesn't seem that bad to warrant a larger refactor just yet 🙂


stefanradev93:
Thanks, Simon! No worries, we are currently refactoring the backend massively.

@LarsKue Tests seem to still be failing. Any idea why?

PR #139, opened 2024-02-29
Fix offline training for model comparison ignoring shared context

elseml:


elseml:
Also added a fix for last epoch validation loss not saving (by computing validation loss before checkpoint saving).

Issue #121, opened 2024-02-07
Amortized point estimators


marvinschmitt:
The paper "Likelihood-free parameter estimation with neural Bayes estimators" (Sainsbury-Dale, Zammit-Mangion, & Huser, 2023) enables neural amortized *point* estimation, which is generally faster than fully Bayesian neural inference with conditional generative models (current standard in BayesFlow).

Implementing amortized point estimators in BayesFlow would help with fast prototyping and sanity checks (and surely other tasks, too!). 

**Some BayesFlow pointers/proposals:**
- Implement the point estimation features in a class  `AmortizedPointEstimator` in `bayesflow.amortizers`
- Implement appropriate loss functions in `bayesflow.losses` (depending on the target quantity)
- Implement a suitable **inference network** in `bayesflow.inference_networks` which takes the summary network output (e.g., DeepSet outputs) and regresses to the point estimate
- Summary networks are already implemented in `bayesflow.summary_networks` and probably work out-of-the-box

**Some links:**

- Paper: https://doi.org/10.1080/00031305.2023.2249522
- `NeuralEstimators` package (R): https://github.com/msainsburydale/NeuralEstimators
- `NeuralEstimators.jl` package (Julia): https://github.com/msainsburydale/NeuralEstimators.jl

zengjice1991:
Great proposals! I read the paper. But I am not sure if the method can generate posterior samples, which may be helpful to detect multi-modal distributions.

stefanradev93:
Yes, the method will not deal with multimodality out of the box. I think its main use case will be as a fast and frugal heuristic to obtain estimates very quickly.

paul-buerkner:
Idea: can we train the summary nets based on point inference nets and then
reuse the trained summary nets in combination with full posterior inference
nets?

Stefan Radev ***@***.***> schrieb am Di., 13. Feb. 2024,
12:43:

andrewzm:
Multi-modality on the marginals should be detectable through the posterior
quantiles. The main "compromise" is that there is no easy way to get
information on posterior joint dependence generally. But as Stefan said,
this is more of an "exploratory" step or an alternative for when the full
NF framework is proving difficult to fit.

I like Paul's idea. Possible issue: Summaries that are good for estimating
individual parameters might not be relevant/suitable for extracting
information on posterior correlations (additional summaries might be needed
when doing full posterior inference).

stefanradev93:
[...]

PR #145, opened 2024-03-09
add notebook: amortized point estimation

vpratz:
A quick demonstration on how to use the new `AmortizedPointEstimator` class discussed in #121. Contains the illustrative example of the original paper.


stefanradev93:
Very nice! The ConfigurableMLP is now part of the library in `helper_networks.py`, so the notebook can directly import and use it.

vpratz:
Thanks for the review! I extended `ConfigurableMLP` to allow for arbitrary output dimensionality, so now we can use it for the current use case (input_dim!=output_dim) as well.

stefanradev93:
Cool and good idea with the different dimensionalities!

PR #147, opened 2024-03-11
Bayes Estimators: Loss functions with flexible signature

han-ol:
Implementation proposal to the signature change of the `loss_fun` used for Amortized Bayes Estimators as discussed in https://github.com/stefanradev93/BayesFlow/issues/121.

It stays as simple as before to use `norm_ord`, but the optional `loss_fun` argument expects a function object with two arguments, rather than the difference of parameters and network output.


stefanradev93:
Great idea!

han-ol:
Thanks!

PR #154, opened 2024-03-19
fix #153

### LINKED ISSUE #153: pip install v1.1.5 fails on Mac (M1)
## Problem Description 

After v.1.1.5, a fresh `pip install bayesflow` on Apple Silicon (M1) fails due to conflicting dependencies:

```
ERROR: Cannot install bayesflow==1.1.1, bayesflow==1.1.2, bayesflow==1.1.3, bayesflow==1.1.4 and bayesflow==1.1.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    bayesflow 1.1.5 depends on tensorflow-macos<2.16 and >=2.10; sys_platform == "darwin" and platform_machine == "arm64"
    bayesflow 1.1.4 depends on tensorflow-macos>=2.10; sys_platform == "darwin" and platform_machine == "arm64"
    bayesflow 1.1.3 depends on tensorflow-macos>=2.10; sys_platform == "darwin" and platform_machine == "arm64"
    bayesflow 1.1.2 depends on tensorflow-macos>=2.10; sys_platform == "darwin" and platform_machine == "arm64"
    bayesflow 1.1.1 depends on tensorflow-macos>=2.10; sys_platform == "darwin" and platform_machine == "arm64"
```

## Underlying Cause

After some initial research, this looks like a consequence of a breaking change in tensorflow 2.16.1: Apple Silicon users shall now use `pip install tensorflow` (new) instead of `pip install tensorflow-macos` (old). 

**Source:** https://github.com/tensorflow/tensorflow/releases/tag/v2.16.1

BayesFlow's current `setup.cfg` still differentiates between OS types to install either `tensorflow` or `tensorflow-macos`.

## Solution

BayesFlow's `setup.cfg` needs to be updated. If we decide to stop support for the legacy Apple x86 hardware (Intel chips), we can just remove the conditional statement in `setup.cfg` and always install `tensorflow`. Opinions appreciated.

marvinschmitt:
Fix dependency conflict during installation on Mac Apple Silicon, https://github.com/stefanradev93/BayesFlow/issues/153

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,8 +42,8 @@ install_requires =
     seaborn >= 0.11
     tqdm >= 4.65
     matplotlib >= 3.5
-    tensorflow-macos >= 2.10, < 2.16; sys_platform == 'darwin' and platform_machine == 'arm64'
-    tensorflow >= 2.10.1, < 2.16; sys_platform != 'darwin' or platform_machine != 'arm64'
+    tensorflow-macos >= 2.10, < 2.16; sys_platform == 'darwin' and platform_machine == 'x86_64'
+    tensorflow >= 2.10.1, < 2.16; sys_platform != 'darwin' or platform_machine == 'arm64'
     tensorflow_probability >= 0.17, < 0.24
 
 [options.extras_require]


LarsKue:
Looks good to me. I am still not a fan of the `!=` platform specifiers, since this assumes any platform can install tensorflow, but for now I think this is a good solution.

Issue #144, opened 2024-03-09
Links in the table of contents of the example notebooks do not work


rusty-electron:
* The recent commit 162a616c0267839998588e547ce34aa3bb966c91 adds TOC to all the example notebooks in https://github.com/stefanradev93/BayesFlow/tree/master/examples
* The TOC links work well within the jupyter notebooks (even in VScode notebook mode). But it appears that when the notebook is rendered by sphinx (for the website), it generates additional anchor links for its own TOC list and this breaks the TOC links embedded in the jupyter notebook.

paul-buerkner:
Note to devs: we will need to check this before merging dev into main.

PR #161, opened 2024-04-16
Improve quickstart NB

AlexAndorra:
- [x] Added some explanation
- [x] Fixed typos and hyperlinks (note that the links from the table of content are still broken, as it's handled by #144 )
- [x] Last section: I'll show how to use the trained approximator on real data. 

I'll change the PR out of draft mode when I check the last item, but the PR is already ready for review 🍾 
Open to comments and suggestions!


stefanradev93:
This is very nice and all the links seem to work fine! Only the image refs seem to be currently broken.

AlexAndorra:
Just added the inference part, so this PR is now ready for review 🍾 
Thanks for the first one @stefanradev93 , happy you like it! What do you mean by "the image refs"?

PR #163, opened 2024-05-07
tests for new backend (relevant to decorators?)

rusty-electron:
This PR is a work in progress. More commits will be appended to this.

## list-of-changes
* added test to check the randomness of the tensors obtained when `.sample` method is called for functions decorated by `DistributionDecorator` (for cases of both `is_batched=False` and `is_batched=True`)


LarsKue:
Closed after correspondence.

PR #170, opened 2024-06-03
Exposed Deepset

Chase-Grajeda:
- Added summary_networks folder in experimental for future additions. Added `__init__` file for exposed summary networks.
- Added `deepset` implementation to summary network class hierarchy.


stefanradev93:
Nice! As discussed:

- Add self-contained invariant / equivariant networks into own deepset module (folder with __init__.py)
- Ensure everything works

stefanradev93:
Almost there! The new files should go into their own new folder (deepset) which will sit at the same level as set_transformer, coupling_flow, etc. and the new summary_networks folder should be removed.

PR #180, opened 2024-07-05
Inherit benchmark simulators

Chase-Grajeda:
Implemented example for using an inherited simulator
Added OOP design to two_moons.py


stefanradev93:
Looking good!

PR #193, opened 2024-09-23
DiagonalStudentT distribution

paul-buerkner:
This PR implements log_prob for the DiagonalStudentT distribution for use as latent distributions in generative networks. 

I have taken the math from: https://mc-stan.org/docs/functions-reference/unbounded_continuous_distributions.html#student-t-distribution

Here is some example code:

```Python
import numpy as np
import bayesflow as bf

student_t = bf.distributions.DiagonalStudentT(df = 200, loc = 1.5, scale = 2.0)
student_t.build(input_shape = (5,))
student_t.log_prob(np.zeros(5))

normal = bf.distributions.DiagonalNormal(mean = 1.5, std = 2.0)
normal.build(input_shape = (5,))
normal.log_prob(np.zeros(5))
```

Open points:

- What is a good default value for df? Was 50 but I don't know the reason for this choice
- How do we make the keras.seed_generator compatible with the scipy student_t sample method (if we need to)?
- We currently allow loc and scale to be tensors but this errors if scale is actually a tensor
- The same error happens in `DiagonalNormal` with std tensor by the way because the code expects a scalar.

Related to the two latter points: Shall we make the code compatible with scale/std tensors or only allow scalars?


LarsKue:
@paul-buerkner thank you for the PR!

The lint appears to fail. Please ensure this passes with a pre-commit hook, see [here](https://github.com/stefanradev93/BayesFlow/blob/dev/CONTRIBUTING.md#2-set-up-your-development-environment). You can also run

```bash
ruff format
ruff check --fix
```

to manually verify and apply automatic fixes.

> keras.seed_generator compatibility

JAX backend might need the seed generator. I think the best way to verify this is to run integration code in JAX backend using the simulator. If a compiled version of the distribution does not yield random samples or errors, then we need to use the seed generator.

As for compatibility, I think we can extract the next seed with `keras.ops.convert_to_numpy(seed_generator.next()[0]).item()`. The 2nd element seems to always be zero (we should verify this before applying this solution).

> loc and scale as tensors

I think supporting tensors is a good idea, since it is quite a typical case for distribution means to not lie on the diagonal, even for diagonal

paul-buerkner:
Thanks! code format fixed. 

As for the seed generation, can you point me to an existing integration test that I can amend to try it out?

LarsKue:
I don't think we currently have a good integration test that involves the distributions. They are unit-tested a little bit, perhaps we should add a jax-only test that explicitly compiles them.

LarsKue (approved):

paul-buerkner:
Okay. Let's wait for @stefanradev93 to fix the tensor issue before merging.

stefanradev93:
Implementation now only matches scipy for scale=1., but not for different diagonal factors:

`import matplotlib.pyplot as plt
from scipy import stats
import keras
import bayesflow as bf

x = keras.random.normal((32, 3))

distro = bf.distributions.DiagonalStudentT(df=30, loc=np.array([0., 0., 0.]), scale=np.array([3., 1, 1]))
distro.build((3,))
print(distro.log_prob(x))

scipy_t = stats.multivariate_t(loc=[0, 0, 0], shape=[3., 1, 1], df=30)
print(scipy_t.logpdf(x.numpy()))

plt.scatter(distro.log_prob(x).numpy(), scipy_t.logpdf(x.numpy()))`

bayesflow/distributions/diagonal_student_t.py
```diff
@@ -1,40 +1,121 @@
...
+            trainable=self.dynamic_params,
+            dtype="float32",
+            initializer=keras.initializers.Constant(self.scale_diag),
+        )
 
     def _log_unnormalized_prob(self, tensor: Tensor) -> Tensor:
-        # TODO
-        pass
+        log_unnormalized_pdf = -0.5 * (self.df + self.dim)
+        log_unnormalized_pdf *= ops.log(
```
LarsKue, line 114: This computation fails in torch backend due to invalid broadcasting. Splitting it into multiple lines would help narrow the error (and look a bit cleaner).

LarsKue (changes requested):

LarsKue:
Cleaned up and fixed. Please review again @stefanradev93 for correctness.

stefanradev93:
> Cleaned up and fixed. Please review again @stefanradev93 for correctness.

- The latest commit does not enable the old functionality (e.g., possibly learnable parameters) and appears to be a new implementation from scratch that does not reflect the reasoning behind the previous one.
- Log normalizing constants should be added, not subtracted (as agreed to avoid confusion)
- Std != scale for student t and should be named as proposed in the previous commit to avoid confusion
- Tensor arguments should be discouraged -> users should use numpy arrays.
- etc

paul-buerkner:
some of these issues we had also identified today in our meeting I believe.
I suggest the following: lars reverts his rewrite and fixes stefans
implementation to work as expected also with pytorch (I believe pytorch
complained right?)

Stefan Radev ***@***.***> schrieb am Di., 8. Okt. 2024, 20:22:

LarsKue:
I'm happy to fix and change my implementation with respect to your requested features, but I dislike the idea of reverting.

Just like you, I made these changes for several reasons:
1. There was a lot of shape broadcasting in very few lines, to the point I deemed rewriting faster than trying to understand it. This is now much better split up, so should also be more maintainable in case of future changes.
2. The `sample` method was incorrectly returning numpy arrays when type hinted as Tensor.
3. Learnable parameters are good, but the initialization is incorrect. `keras.initializers.Constant` is only supposed to support scalar initializers, see also https://github.com/keras-team/keras/issues/19767 and may thus change with future keras releases.
4. The seed generator was unused.

If you still want to revert, I would ask that one of you implement the torch fix, as I was unable to recognize where the incorrect broadcasting happens.

paul-buerkner:
I am also happy with you fixing and changing your implemention. I
understand too little of the pros and cons. stefan what do you think? in
any case I am happy to test the implemenation before merging.

Lars ***@***.***> schrieb am Di., 8. Okt. 2024, 21:54:

paul-buerkner:
I made some minor fixes to the math of both normal and student_t and checked it against my reference implementations. Things look correct now. I also removed the default for df since this is really something that should be decided explicitly upon initializing the object. 

@stefanradev93 can you take another look and if you are happy to merge this PR?

tests/test_distributions/conftest.py
```diff
@@ -20,12 +22,17 @@ def random_samples(batch_size, num_features):
...
-
     return DiagonalNormal()
 
 
-@pytest.fixture(params=["diagonal_normal"])
+@pytest.fixture()
+def diagonal_student_t():
+    from bayesflow.distributions import DiagonalStudentT
+
+    return DiagonalStudentT()
```
LarsKue, line 32: This needs an argument now. Tests should be passing otherwise.

LarsKue (approved):

tests/test_distributions/conftest.py
stefanradev93: I am happy with the changes. Should we force a choice of df or use a default close to a Gaussian?

paul-buerkner: I prefer forcing a choice for now. Especially a choice close to Gaussian would not be ideal I think to avoid the following scenario: A user struggles with their model and is told to try out student-t latent space. They try it out as a drop in replacement for normal (with defaults) and find no difference. Then they might think it doesn't make a difference. 

If we end up using a default later on, I would prefer it to be some small df (10 or 7 or even 3) with which we have made good experience in a bunch of experiments. For now, I would leave it without default to make the user aware of the importance of this choice. Later on we still need to document good options properly to give users a basis for decision.

paul-buerkner: I will fix the failing test and then merge.

paul-buerkner:
I will merge this PR now. Should we decide that we want to have a default for df already now, it is just a one-line change. Thank you for working on this feature!

PR #194, opened 2024-09-23
Add new metrics

paul-buerkner:
This PR is just to start the discussion on new metrics (beyond MMD). I have implemented the RMSE and would like you, @LarsKue, to check if the code structure generally looks reasonable. 

Additional questions:

- For MMD, the number of samples (2nd dimension in the tensor?) does not need to be the same for both tensors. For RMSE the number of samples needs to be the same. How do we safely enforce this?
- What is the best way to normalize the RMSE? There are different ways to do it (e.g., https://www.marinedatascience.co/blog/2019/01/07/normalizing-the-rmse/). What is typtically considered in our context?


LarsKue (approved):
LGTM!

PR #196, opened 2024-09-27
[For discussion only] Add a basic-normal example

paul-buerkner:
Based on the two moons case study, I made another case study estimating mean and SD from a normal distribution. In the process, I noted down a lot of things that I think are a bit unclear or whose user interface could be improved. I don't necessarily intend to have this case study in the end (or merge this PR), but primarily opened it for discussion. Tagging @LarsKue.


paul-buerkner:
The resulting linear regression example is now on dev. So I am closing this PR.

Issue #200, opened 2024-10-08
Flow Matching: Allow passing class as subnet (2.0 version)


vpratz:
This is probably already planned: `FlowMatching` currently only accepts strings for the `subnet` argument. Allowing classes as well would help with modularity, as it enables easily plugging in custom networks for power users, without sacrificing ease of use for others.
Handling the case in `utils.find_network` and initializing the class there might be sufficient to enable this.

marvinschmitt:
I like the idea and second the proposal. 

Would you mind drafting a PR that implements this? Based on your description it should be quite straightforward.

marvinschmitt:
closed by #201

PR #201, opened 2024-10-09
Allow calling find_network with type (#200)

vpratz:
* in addition to strings, accept type (i.e., uninstantiated classes) as well
* allows easy use of user-defined networks
* adapt type hints to reflect this change


marvinschmitt:
closes #200

PR #203, opened 2024-10-10
Implement DiagonalStudentT.sample method natively in keras

paul-buerkner:
When trying to run .sample of the student_t distribution the torch backend complained about some incompatible types, presumably because scipy and torch types did not like each other. So I went ahead and implemented the sample method natively using keras functions using the fact that chi-square samples can be obtained for standard gamma samples which are implemented in keras.

Here is a small test providing evidence for the implementation's correctness:

```python
import keras
import numpy as np

## REMOVE ON PRODUCTION
import sys
sys.path.append('../')

import bayesflow as bf

t = bf.distributions.DiagonalStudentT(df = 10, loc = 1.0, scale = 3.0)
t.build(input_shape = (3,))

y = t.sample((100000, ))
y.shape
y.mean(dim = 0)
y.var(dim = 0)

# analytical variance:
# 3.0**2.0 * 10.0 / (10.0 - 2.0) = 11.25
```

@LarsKue can you review this PR? Specifically, can you check whether my shape broadcasting is correct? I assume there is a better pattern than repeat and reshape for what I want to achieve.

### DIFF
--- a/bayesflow/distributions/diagonal_student_t.py
+++ b/bayesflow/distributions/diagonal_student_t.py
@@ -6,6 +6,8 @@ import numpy as np
 
 
 from bayesflow.types import Shape, Tensor
+from bayesflow.utils import expand_tile
+
 from .distribution import Distribution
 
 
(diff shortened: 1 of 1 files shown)


LarsKue (approved):
I made some minor changes, mostly fixing the seed of the call to `keras.random.gamma`. Looks good to me otherwise!

paul-buerkner:
Checks are passing. Merging now.

PR #205, opened 2024-10-10
Complete implementation of Consistency Models

vpratz:
Completes the implementation of Consistency Models.
I tested the implementation as a drop in replacement in the two moons notebook.


marvinschmitt:
Thanks for the PR @vpratz, that's fantastic! Since you were creating a notebook to test the implementation anyway, would you mind adding that to the `examples/` folder as well (maybe with some minimal accompanying text)?

vpratz:
Thanks for your feedback, I have added the notebook. I included a background section that turned out to be not very minimal, feel free to modify/shorten as you see fit.

marvinschmitt (approved):
Thanks again for the PR. I made some changes to the tutorial notebook:

- fix a few typos
- adjust $x \to \theta$ in the initial description of consistency models to be consistent (haha) with our subsequent conditional version, where $\theta$ is the target variable $x$ is the conditioning variable.
- use hyperparameters from the Consistency Model Posterior Estimation paper: 
  - empirical variance of the prior samples as `sigma2` 
  - `max_time=10`

We can merge from my side. 

@stefanradev93 @LarsKue before we merge:
- Any feedback or change requests from your side?
- What's the strategy for adding tests?

PR #219, opened 2024-10-18
Modular Data Adapters and Simulator Dispatch

LarsKue:
This PR has 2 main features:

1. Data Adapters are now much more modular:
```py3
data_adapter = bf.DataAdapter() \
    .drop("dim") \
    .to_array() \
    .as_set("y") \
    .broadcast("n") \
    .convert_dtype(from_dtype="float64", to_dtype="float32") \
    .standardize(exclude="n") \
    .concatenate(["mu", "sigma"], into="inference_variables") \
    .rename("n", "inference_conditions") \
    .rename("y", "summary_variables")
```
2. Simulators can now be created via dispatch utility, which automatically tries to select an optimal implementation for the simulator:
```py3
simulator = bf.simulators.make_simulator([prior, likelihood], meta_fn=meta_fn)
```

I am open to improving the overall simulator structure, still; either with this pull request or after merging. If you feel like there are any critical bugs or missing features, please also do not hesitate to request changes.

Tests may still require updates, so leaving this as a draft PR for now.


paul-buerkner:
This looks very very cool Lars! 

One quick immediate question. Shall we think about the `meta_fn` argument name for a bit? First of all, do we think that `meta` is a good name? I think I remember that I kind of suggested this but I would like us to think about it more before just going with it. Second, do we need the `_fn` suffix? I think it makes the argument name look a bit ugly.

LarsKue:
@paul-buerkner Thank you for bringing this up. I think `meta` is fitting in the sense that we supply meta data about the batch, but I am open to other names. The `_fn` suffix signifies to the user that this is supposed to be a function object (as opposed to a Simulator or bare values), so I would leave this in.

marvinschmitt:
I personally like the function signifier suffix. 

Any thoughts on `_fn` vs. `_fun` vs. `_func`? Needless to say, once we choose one, that should be consistently used throughout the library.

LarsKue:
@marvinschmitt I have consistently used `fn` everywhere so far (which is not too many places to switch now though). But I think this is also because I come from a programming background where this is more commonly used than the other alternatives :smile:

paul-buerkner:
I also prefer _fn given the common options. so let's stick to it!

Marvin Schmitt ***@***.***> schrieb am Fr., 18. Okt. 2024,
15:04:

paul-buerkner (approved):
From the perspective of the user interface I think this looks really good and ready to merge from my side.

PR #220, opened 2024-10-22
allow int as type for batch_shape

paul-buerkner:
As discussed. @LarsKue feel free to edit as you wish :-)

### DIFF
--- a/bayesflow/simulators/composite_simulator.py
+++ b/bayesflow/simulators/composite_simulator.py
@@ -1,8 +1,8 @@
 from collections.abc import Sequence
 import numpy as np
 
-from bayesflow.types import Shape
-from bayesflow.utils import validate_batch_shape
+from bayesflow.types import ShapeLike
+from bayesflow.utils import validate_shape
 
 from .simulator import Simulator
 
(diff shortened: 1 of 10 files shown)


LarsKue:
I agree about putting validators into utils. However, the validator function should also actually validate the shape, not just convert ints to tuples.

stefanradev93:
See my recent commit.

stefanradev93:
We also need a function that validates if NaNs, Infs, etc are present in the approximator / simulator outputs. The latter is especially important, as it can help users immediately spot problems.

paul-buerkner:
would it make sense to rename argument from batch_shape to batch_size even if we (for now) keep allowing multi dim batches? we also have batch_size later on on approximator.sample and I fail to see why we sometimes use batch_shape and sometimes batch_size

LarsKue:
The approximator used to support a batch_shape but I had to simplify this a little bit to get it to work cleanly. Ideally I would like to support multi-dimensional batch shapes everywhere, but this can be overly complicated sometimes

paul-buerkner:
Okay, but there would be no issue renaming batch_shape to batch_size, right? We can still support multi dim batch_sizes where we can/want.

paul-buerkner:
Closing this PR as I think it will be implemented instead in #239

PR #224, opened 2024-10-25
Consistency Model: Fix JIT compilation

### LINKED ISSUE #223: Consistency Models: current training step not correctly tracked with JIT compilation enabled
In the current implementation, we use `self.current_step += 1`, which is not supported in TensorFlow and Jax.

Instead, we have to track the step as a Variable, so that it can be updated via `self.current_step.assign`.

vpratz:
The current consistency model implementation did not take JIT compilation into account.
This PR contains changes to correctly track the current step and apply the discretization step schedule.

To make this work with JAX, which only allows arrays with known size, we calculate the schedule in advance.

Fixes #223

### DIFF
--- a/bayesflow/networks/consistency_models/consistency_model.py
+++ b/bayesflow/networks/consistency_models/consistency_model.py
@@ -1,5 +1,3 @@
-import math
-
 import keras
 from keras import ops
 from keras.saving import (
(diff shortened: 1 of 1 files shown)


stefanradev93:
Thanks, nice spot, Valentin!

PR #234, opened 2024-11-02
Add continuous-time consistency models

vpratz:
Implements continuous-time consistency models as described in [1].

I observe differences in the quality of the samples depending on the backend (all else equal) that I cannot explain yet. Tensorflow performs consistenly better than Jax and PyTorch. If you find any indications for the reasons of this difference, please let me know.

[1] Lu, C., & Song, Y. (2024).
    Simplifying, Stabilizing and Scaling Continuous-Time Consistency Models
    arXiv preprint arXiv:2410.11081


vpratz:
First observation: Loading a model trained in Tensorflow in Jax works with equal quality, so the difference is somewhere in the training process, not in the sampling.

vpratz:
The difference was caused by a coincidence, which led me to discover an error in the loss calculation. It is now fixed, and all backends lead to similar results.
@stefanradev93 I think the code is now ready to review. We should optimize the schedule for the sampling, but we can do this at a later point in time.

PR #239, opened 2024-11-05
Allow `batch_size` instead of just `batch_shape`

### LINKED ISSUE #221: batch_shape vs. batch_size in simulator and related objects
I was thinking about the `batch_shape` argument which we use across bayesflow, especially in the simulator. 

Currently, we use `batch_shape = (int,)` to allow for potentially multi-dimensional batch shapes. My feeling is that we will never need more than one batch dimension such that we could replace `batch_shape = (int,)` with `batch_size = int` patterns throughout. That is, with the `batch_size` approach, we would always only have and allow for a single batch dimension. 

I was wondering what others thing about this? Tagging @LarsKue and @stefanradev93. Also, @daniel-habermann do you think we would need (or at least benefit from) multiple batch dims for multilevel models?

LarsKue:
This PR does two things. For any user-facing function that takes a `batch_shape` argument, we

1. Create a new argument `batch_size` that is aliased to `batch_shape`
2. Convert any non-sequences passed to one of these aliases to a one-item tuple

This ensures that integers can be used in place of tuples.

Closes #221

### DIFF
--- a/bayesflow/distributions/diagonal_normal.py
+++ b/bayesflow/distributions/diagonal_normal.py
@@ -5,6 +5,8 @@ import math
 import numpy as np
 
 from bayesflow.types import Shape, Tensor
+from bayesflow.utils.decorators import allow_batch_size
+
 from .distribution import Distribution
 
 
@@ -73,5 +75,6 @@ class DiagonalNormal(Distribution):
 
         return result
 
+    @allow_batch_size
     def sample(self, batch_shape: Shape) -> Tensor:
         return self.mean + self.std * keras.random.normal(shape=batch_shape + (self.dim,), seed=self.seed_generator)
--- a/bayesflow/distributions/diagonal_student_t.py
+++ b/bayesflow/distributions/diagonal_student_t.py
@@ -7,6 +7,7 @@ import numpy as np
 
 from bayesflow.types import Shape, Tensor
 from bayesflow.utils import expand_tile
+from bayesflow.utils.decorators import allow_batch_size
 
 from .distribution import Distribution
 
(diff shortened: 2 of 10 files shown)


paul-buerkner:
Thanks! How does this show up in the docs? Can users see the batch_size argument there?

LarsKue:
@paul-buerkner Not unless we explicitly mention it. I could look into if it is possible to modify the decorator to account for docs, though.

paul-buerkner:
thanke! since batch_size should be the argument that 95% of users will use,
I would strongly prefer batch_size being the main argument and batch_shape
being the alias. would you mind changing that?

Lars ***@***.***> schrieb am Di., 5. Nov. 2024, 22:30:

LarsKue:
@paul-buerkner This is not really possible internally since the batch shape is more general than a batch size, so it needs to be the actual argument that is converted to.

paul-buerkner:
I see. More important then that batch_size appears in the docs. Thank you for checking out this option!

paul-buerkner:
Another question that I have in this regard. If I call `simulator.sample(10)`, say, would that work? Or do I have to write `simulator.sample(batch_size = 10)`. I would very much prefer the first prefer to work as well. That is, that `batch_shape = int` would also be allowed *in addition* to the batch_size argument.

LarsKue:
The alias is both in positional and keyword args, so this works 🙂

stefanradev93:
> Another question that I have in this regard. If I call `simulator.sample(10)`, say, would that work? Or do I have to write `simulator.sample(batch_size = 10)`. I would very much prefer the former to work as well. That is (presumably), that `batch_shape = int` would also be allowed _in addition_ to the batch_size argument.
> 
> EDIT: based on my reading, this already works, right? That would make me very happy :-)

It works already, as the `batch_size` is an alias for `batch_shape`.

paul-buerkner:
Lovely! Merging this now! We can deal with the doc later if we want to.

PR #246, opened 2024-11-08
Remove batch size from adapter, infer during inference, and propagate…

stefanradev93:
Draft for 
- removing the dependence on `batch_size` in adapters;
- removing the need for the user to specify `batch_size` and hence inferring `batch_size` dynamically **after** adapting the data;
- propagating the `stage` argument through the adapter and controlling the behavior of EMA etc.

The trick was to re-implement `broadcast` to infer the broadcast dim from another existing array, which will always be the case. No other transforms need to know anything about batch sizes. Here is how it works for scalars:

```
# Mock simulator with realistic shapes - broadcasting number of observations
def sim(batch_size):
    return dict(
        N=10,
        x=np.random.randn(batch_size, 10, 2),
        theta=np.random.randn(batch_size, 2)
    )

# Default behavior - will only broadcast to shape[:-1]
adapter = bf.adapters.Adapter().broadcast("N", to="theta", batch_dims_only=True)
print(adapter(sim(3))["N"].shape)
# (3, 1)


# Broadcast completely to shape
adapter = bf.adapters.Adapter().broadcast("N", to="theta", batch_dims_only=False)
print(adapter(sim(3))["N"].shape)
# (3, 2)
```

Here is how it works for non-scalars (arrays, lists):

```
# Can also easily broadcast non-scalar to the left (along batch_dim)
def sim(batch_size):
    return dict(
        N=np.random.randn(2),
        theta=np.random.randn(batch_size, 2)
    )
adapter = bf.adapters.Adapter().broadcast("N", to="theta", batch_dims_only=False)
print(adapter(sim(3))["N"].shape)
# (3, 2)


# This will be undefined! Shape of N must be (batch_size, 1) to broadcast correctly
def sim(batch_size):
    return dict(
        N=np.random.randn(batch_size),
        theta=np.random.randn(batch_size, 2)
    )
adapter = bf.adapters.Adapter().broadcast("N", to="theta", batch_dims_only=False)
print(adapter(sim(3))["N"].shape)
# ValueError: operands could not be broadcast together
```


stefanradev93:
@paul-buerkner @LarsKue Naming of arguments can be improved. Currently, this broadcast has no inverse, but it is possible to implement by tracking the old shapes during forward pass. Too lazy to implement before merge.

paul-buerkner:
Looks good I think.

About the `batch_dims_only` argument. Since we currently implement it as "everything but the last dim", would it perhaps make more sense to rename it to `last_dim` with `last_dim = False` equal to `batch_dims_only = True`. I still struggle seeing everything else than the first dimension as a batch_dim. Happy to be convinced of course.

One question I have about `approximator.sample`. I take it we now have to wrap data as `[variable]`, as in (from the linear regression notebook):

```python
conditions = {"y": [lik["y"]], "x": [lik["x"]], "N": [N]}
```

vs. previously the following worked (with `batch_size = 1` as an added argument):

```python
conditions = {"y": lik["y"], "x": lik["x"], "N": N}
```

I think it is fine to require `[variable]` for single dataset inference. However, the error one sees otherwsie is quite uninformative. I see: 

```python
File ~/Dropbox/Software/bayesflow/BayesFlow/examples/../bayesflow/approximators/continuous_approximator.py:167, in ContinuousApproximator._sample(self, num_samples, inference_conditions, summary_variables)
    [165](https://file+.vscode-resource.vscode-cdn.net/Users/paul-buerkner/Dropbox/Software/bayesflow/BayesFlow/examples/~/Dropbox/Software/bayesflow/BayesFlow/bayesflow/approximators/continuous_approximator.py:165)         inference_conditions = summary_outputs
    [166](https://file+.vscode-resource.vscode-cdn.net/Users/paul-buerkner/Dropbox/Software/bayesflow/BayesFlow/examples/~/Dropbox/Software/bayesflow/BayesFlow/bayesflow/approximators/continuous_approximator.py:166)     else:
--> [167](https://file+.vscode-resource.vscode-cdn.net/Users/paul-buerkner/Dropbox/Software/bayesflow/BayesFlow/examples/~/Dropbox/Software/bayesflow/BayesFlow/bayesflow/approximators/continuous_approximator.py:167)         inference_conditions = keras.ops.concatenate([inference_conditions, summary_outputs], axis=1)
    [169](https://file+.vscode-resource.vscode-cdn.net/Users/paul-buerkner/Dropbox/Software/bayesflow/BayesFlow/examples/~/Dropbox/Software/bayesflow/BayesFlow/bayesflow/approximators/continuous_approximator.py:169) if inference_conditions is not None:
    [170](https://file+.vscode-resource.vscode-cdn.net/Users/paul-buerkner/Dropbox/Software/bayesflow/BayesFlow/examples/~/Dropbox/Software/bayesflow/BayesFlow/bayesflow/approximators/continuous_approximator.py:170)     # conditions must always have shape (batch_size, dims)
    [171](https://file+.vscode-resource.vscode-cdn.net/Users/paul-buerkner/Dropbox/Software/bayesflow/BayesFlow/examples/~/Dropbox/Software/bayesflow/BayesFlow/bayesflow/approximators/continuous_approximator.py:171)     batch_size = keras.ops.shape(inference_conditions)[0]

File /opt/miniconda3/envs/bayesflow/lib/python3.11/site-packages/keras/src/ops/numpy.py:1352, in concatenate(xs, axis)
   [1350](https://file+.vscode-resource.vscode-cdn.net/opt/miniconda3/envs/bayesflow/lib/python3.11/site-packages/keras/src/ops/numpy.py:1350) if any_symbolic_tensors(xs):
   [1351](https://file+.vscode-resource.vscode-cdn.net/opt/miniconda3/envs/bayesflow/lib/python3.11/site-packages/keras/src/ops/numpy.py:1351)     return Concatenate(axis=axis).symbolic_call(xs)
...
--> [389](https://file+.vscode-resource.vscode-cdn.net/opt/miniconda3/envs/bayesflow/lib/python3.11/site-packages/jax/_src/util.py:389)   raise ValueError(f"axis {axis} is out of bounds for array of dimension {num_dims}")
    [390](https://file+.vscode-resource.vscode-cdn.net/opt/miniconda3/envs/bayesflow/lib/python3.11/site-packages/jax/_src/util.py:390) if axis < 0:
    [391](https://file+.vscode-resource.vscode-cdn.net/opt/miniconda3/envs/bayesflow/lib/python3.11/site-packages/jax/_src/util.py:391)   axis = axis + num_dims

ValueError: axis 1 is out of bounds for array of dimension 1
```

Do think this error is informative enough to point users in the right direction?

LarsKue:
Tests are currently not passing. Could you update this @stefanradev93?

I am also not a huge fan of this change in general. When designing, I explicitly chose to let the batch size be manually supplied from the outside since it is fickle to infer it. It should also always be known when the adapter is called, with the exception being that users apparently do not want to supply a sample size when calling the sample method of the continuous approximator.

Even for that case, I think it would be much more robust to expect the batch size to be passed, unless the user explicitly makes it optional by providing the name of another object that it can be inferred from. I.e. this change should be optional.

Do note that I already have an idea for a better solution: explicitly defined variables which include meta-data (such as which axis is the batch axis).

paul-buerkner:
From a dev perspective, it may be more robust to expect the user to pass the batch_size. But from a user perspective, this makes for a worse user-interface. That is why Stefan and I are both in favor of this more automatic (albeit perhaps not 110% robust) solution. 

Your alternative sounds also good potentially, but I would have to see the details and user-interface of this. What is more, I consider these two not mutually exclusive. One might take the batch_size from the simulator if it is specified explicitly there. And if not, infer it as in Stefans proposal.

paul-buerkner:
I think I now have a general enough version of `broadcast` implemented. It may not yet be pretty, but perhaps we can confirm first if it has all the funcationality we want and need.

`broadcast` now has two new arguments, `expand` and `exclude`:

- `expand` informs where new dimensions should be added to `x` to match the number of dimensions in `to`. Can be `"left"`, `"right"`, or an integer or tuple containing the indices of the new dimensions. The latter is need if we want to include a dimension in the middle, which will be required for more advanced cases. By default we expand left.

- `exclude` informs which dimensions (of the dimensions after expansion) should retain their size, rather than being broadcasted to the corresponding dimension size of `to`. By default we exclude the last dimension (usually the "data" dimension) from broadcasting the size.

@stefanradev93 would you mind checking my implementation and verfying that this covers all our use cases?

PR #250, opened 2024-11-19
support auto-generation of documentation using sphinx-autosummary

rusty-electron:
This PR modifies `docsrc/conf.py` and the existing `bayesflow.rst` to allow auto-generation of documentation using [sphinx-autosummary](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html).

> Note: The layout and structure of the documentation pages can be tweaked as per the design requirements. This version is just a proof of concept :)

## Instructions to test:

* clone PR
* cd into `docsrc`
* run `make clean-all && make github`
* view `index.html` in a browser

## Changes

* documentation improvements
* custom sphinx templates to make the docs look similar to the existing design
* Removal of several `.rst` files, these files are now autogenerated


vpratz:
Thanks a lot for the PR, I think it looks quite good already. I have added the `autosummary_imported_members = True` option, so that the summary reflects our intended structure for imports. @rusty-electron Could you please take a quick look at the changes I made and check whether anything goes against your ideas/intentions?

As we currently do not have a functional doc generation for the `dev` branch, I think merging the PR and then incorporating improvements as we go along would be a good strategy.

I see two aspects we might want to think about:

1. handling versions: only having the docs for the newest version available is an anti-pattern and makes it hard to use/understand code from older versions. Something like [sphinxcontrib-versioning](https://sphinxcontrib-versioning.readthedocs.io/) might help to automate having multiple doc versions based on tags in the repository.
2. hiding internals: we follow a pattern where we define functions and classes in submodules, which we then import in `__init__.py`. autosummary currently surfaces those submodules. If we want to hide them, it would require some manual work, which might not be worth it in the end. Any opinions on this are welcome.

As I'm not sure whether @marvinschmitt has capacity for the review, maybe @stefanradev93 or @LarsKue can comment on whether we want to merge this PR in its current state.

marvinschmitt:
Thanks @rusty-electron and @vpratz for your work on this PR! As you correctly anticipate, I don't have the capacity for a proper review at the moment (probably at least until late January). 


Regarding your questions/topics:

- I agree with your assessment regarding versions 👍
- What are some functions that are clearly internal and shouldn't be included? If those are all underscore-functions (like `_fun`) there might be some way to auto-exclude them, worst case being a custom hook?

Cheers!

vpratz:
Thanks for the quick response. The case I mean are mostly aliases, e.g. a class `bayesflow.adapters.adapter.Adapter`, which we import in `bayesflow/adapters/__init__.py` so that a user can use it as `bayesflow.adapters.Adapter`. The latter is the public part which should stay constant, while we ideally want to be able to change the internal structure of the `bayesflow/adapters/` (in this example this might be not too likely, but for example the internal organisation of the utility functions might be subject to change).
Ideally, we would have a "public API" that exposes only the intended import paths, and an "internal API" that cohtains everything. But the manual specification of what is public and what is internal makes this time-consuming.

Issue #118, opened 2024-01-03
Free-form flows for SBI


zimea:
The paper 'Free-form Flows: Make Any Architecture a Normalizing Flow' by Draxler et al. 2023 introduces a loss that allows to use any architecture as inference network for sbi. A new amortizer could implement the loss function and be tested on benchmark tasks.

Link to paper: https://arxiv.org/abs/2310.16624

PR #251, opened 2024-11-20
Add free-form flows as inference networks

vpratz:
WIP, regards #118.

* implements the fff loss as in the [reference implementation](https://github.com/vislearn/FFF)
* still missing: calculation of the log probability

Free-form flows can be an interesting model class for SBI. In my first trials (without a lot of tuning), I could not achieve the performance of our existing inference networks. Nevertheless, a readily available implementation enables exploration and therefore would be worthwhile to have.


stefanradev93:
Great idea and much needed! Looking forward to the implementation. :)

LarsKue:
Also looking forward to this. Please also check out the already backend-agnostic [`make_vjp_fn`](https://github.com/bayesflow-org/bayesflow/blob/dev/bayesflow/utils/jacobian_trace/_vjp.py). I think it would be good to make this a general util, now that it is used in multiple places.

vpratz:
Thanks for the hint regarding the `make_vjp_fn`, @LarsKue. I have renamed the function to `vjp`, in analog to torch and jax which provide `jvp` and `vjp` with the same functionality, but without the `make_fn` part. If you prefer the previous naming, let me know and I will change it accordingly.
I also changed the function calls for torch from the `autograd.functional` API to the `func` API. Using `autograd.functional` broke FFF training, probably because the gradients were not left intact. With `func`, it works without problems. If you see any problems with this change, let me know so that we can discuss how to proceed.

I also added batched jacobian and jacobian determinant computation. Please check the naming to ensure it fits your desired naming scheme.

The implementation of the density is finished as well, so I believe this PR is ready for review.

LarsKue:
> I have renamed the function to `vjp`

Thanks, this is a good change.


> I also changed the function calls for torch from the `autograd.functional` API to the `func` API.

I also like this change. I used autograd because I was copying from code I wrote over a year ago (before `torch.func` was included in the base `torch` installation).



> I also added batched jacobian and jacobian determinant computation. Please check the naming to ensure it fits your desired naming scheme.

These functions produce errors for me under torch backend due to vmap. Please double check. Jacobian computation is also very closely related to Jacobian trace computation, so we could get the fix from how the latter is written, and merge these into one module.

The `batch_wrap` and `double_output` decorators could also be defined inside `compute_jacobian` if they survive the fix and are not used anywhere else. This avoids polluting the namespace or suggesting these have further uses.

> The implementation of the density is finished as well, so I believe this PR is ready for review.

The inference network itself looks good to me. Should be ready to merge once the tests pass!

stefanradev93:
@LarsKue bump.

bayesflow/utils/jacobian.py
```diff
@@ -0,0 +1,129 @@
...
+        'forward'.
+    :param func_kwargs: The keyword arguments to pass to the function.
+        func_kwargs are not batched.
+    :return: The output of the function `fn(x)` and the log jacobian determinant
+        of the function with respect to its input `x` of shape
+        (batch_size, out_dim, in_dim)."""
+
+    x_out, jac = compute_jacobian(x_in, fn, *func_args, grad_type=grad_type, **func_kwargs)
+    jac = ops.reshape(
+        jac, (ops.shape(x_in)[0], ops.prod(list(ops.shape(x_out)[1:])), ops.prod(list(ops.shape(x_in)[1:])))
```
LarsKue, line 125: Would prefer this in multiple lines

LarsKue (approved):
Looks good, thank you for the PR!

PR #263, opened 2024-11-30
Add Docs to Adapter Transforms

eodole:
Added documentation for the following transforms: 
- keep 
- to_array
- convert dtype
- ad_set: gave an explaination but no useage/example as I am unsure what exactly this transform is doing
- constrain: there are many method strings that all map to the same function so I only included the first/second argument


stefanradev93:
Hey Leona, thanks for the docs! Can you ensure that the linter is passing by including the pre-commit hooks?

paul-buerkner:
Thank you for adding some docs! I made some minor edits and ran the linter. Will merge now.

PR #267, opened 2024-12-04
Dev

eodole:
Documentation added to: 
- concatonate.py: not sure if the options are all correct... can "inference_conditions" be specified as the into string? 
- drop.py: this just seems like the opposite of keep, I'm not sure why we have both
- elementwise_transform.py : not sure what this does other than be a parent class

### DIFF
--- a/bayesflow/adapters/transforms/as_set.py
+++ b/bayesflow/adapters/transforms/as_set.py
@@ -5,12 +5,13 @@ from .elementwise_transform import ElementwiseTransform
 
 class AsSet(ElementwiseTransform):
     """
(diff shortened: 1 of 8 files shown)


paul-buerkner:
Thank you! Can you please provide more information names for your PR?

paul-buerkner:
You can just rename your PR. No need to close this one.

eodole:
I realized that concatonate.py didnt get resolved I have to fix it then push another commit

paul-buerkner:
You can always commit more commits to a PR. But I am not sure I understand the issue correctly.

PR #268, opened 2024-12-04
Added Documentation to Adapter Transforms 

eodole:
Documentation added to:

- concatonate.py: not sure if the options are all correct... can inference_conditions" be specified as the into string?
- drop.py: this just seems like the opposite of keep, I'm not sure why we have both
- elementwise_transform.py : not sure what this does other than be a parent class

### DIFF
--- a/bayesflow/adapters/transforms/as_set.py
+++ b/bayesflow/adapters/transforms/as_set.py
@@ -5,12 +5,13 @@ from .elementwise_transform import ElementwiseTransform
 
 class AsSet(ElementwiseTransform):
     """
-    The `.as_set(["x", "y"])` transform indicates that both `x` and `y` are treated as sets. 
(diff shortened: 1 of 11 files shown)


paul-buerkner (changes requested):
Thank you! please see my minor comments.

paul-buerkner:
Thank you! We have both keep and drop because it may be more efficient to use either of the two depending on the application. Doc for ElementwiseTransform looks good.

eodole:
I changed the documentation in 
- standardization.py: I wasn't sure about the mean/std explaination, I could also remove saying that people should specify their own mean/std because im not really sure what the use case would be
- transform.py 
- one_hot.py

paul-buerkner:
I think this now looks good, thank you! Will merge now. About your standardize question: You are not sure what the use-case would be for the mean and SD inferred from data? Or the use case of a user-specified mean and SD?

PR #271, opened 2024-12-09
Final adapter Documentation 

eodole:
All files in the transforms folder now have some form of documentation. Added in this PR are:
- filter_transform: I did my best to describe it but I don't think it is one that is used directly by the user 
- rename.py


paul-buerkner:
Looks good, thank you!

PR #275, opened 2024-12-09
Fix: warnings in softplus and inverse_softplus #270

### LINKED ISSUE #270: softplus and inverse_softplus warn about overflow when they should not
These functions lead to a warning when `expm1`/`exp` overflows, but this overflow is already disarmed by `np.where`.

```python
 def inverse_softplus(x: np.ndarray, beta: float = 1.0, threshold: float = 20.0) -> np.ndarray:
    """Numerically stabilized inverse softplus function."""
    return np.where(beta * x > threshold, x, np.log(beta * np.expm1(x)) / beta)

...

def softplus(x: np.ndarray, beta: float = 1.0, threshold: float = 20.0) -> np.ndarray:
    """Numerically stabilized softplus function."""
    return np.where(beta * x > threshold, x, np.log1p(np.exp(beta * x)) / beta)
```
The expected behavior is: Don't warn when such a caught overflow occurs.

han-ol:
Resolves #270.

### DIFF
--- a/bayesflow/utils/numpy_utils.py
+++ b/bayesflow/utils/numpy_utils.py
@@ -16,7 +16,9 @@ def inverse_shifted_softplus(
 
 def inverse_softplus(x: np.ndarray, beta: float = 1.0, threshold: float = 20.0) -> np.ndarray:
     """Numerically stabilized inverse softplus function."""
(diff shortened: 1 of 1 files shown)


LarsKue (approved):
Thanks for the fix!

PR #281, opened 2024-12-12
Flexible implementation of point estimation

han-ol:
This (draft) pull request is meant to hold discussion and development of point estimation within BayesFlow. 

The functionality per se was also discussed in the issue https://github.com/bayesflow-org/bayesflow/issues/121.

The implementation should make it easy to
* use standard point estimators,
* extend an existing collection with more complex point estimation,
* and use custom research estimators.

Commit 093785d9f80fde678f9b491bb3a95323166325c3 contains a first example including ONLY quantile estimation.

# Writing down draft specifications and guiding thoughts

## Names for everything
What is "inference", does it include point estimation? Are we calling networks discussed below PointInferenceNetworks, PointRegressors, or something else?
For now I stick with `ContinuousPointApproximator` and `PointInferenceNetwork` to make their roles in relation to the existing codebase obvious.

## Components 
A `ContinuousPointApproximator` parallels the `ContinuousApproximator` and bundles some feed-forward architecture with an optional summary network suitable for learning point estimates optimized to minimize some Bayes risk.
Thus it serves the same roles (including Adapter calls, summary network, etc), but instead of a `sample` method it provides an `estimate` method.

`PointInferenceNetwork` parallels the `InferenceNetwork` by providing a base class to inherit from for ~~generative model classes suiteable for the role of approximating a conditional distribution~~ feed-forward model classes suitable for point estimation.

## Convenient default estimator

The API for functionality that covers the need of most users could be something like
```python
point_inference_net = NameForStandardCompositePointInferenceNetwork()
point_approximator = ContinuousPointApproximator(
	summary_network = ...,
	inference_network = point_inference_net,
)
# define optimizer, compile(), fit() as usual
```
, with optional constructor arguments to tweak it a bit:
```python
NameForStandardCompositePointInferenceNetwork.__init__(
	self,
	mean: bool = True,
	std: bool = True,
	quantile_levels: Sequence[float] = [0.1, 0.5, 0.9]
	subnet: str | type = "mlp",
	**kwargs,  # e.g. subnet_kwargs = dict(widths=[64, 32], dropout=0.1)
)
```

**choices in output of `estimate`**

For inference the method `estimate(data_dict))` produces the point estimates for a given input/condition/observation.

The default PointInferenceNetwork would produce 5 point estimates (mean, std and three quantile levels) of the marginal distribution for each inference variable.

These estimates need more explanation than samples provided by generative networks typically need. We need to communicate to user, diagnostics code or other downstream tasks which estimate lands where. It seems to me that it would be helpful if such `estimate` output is structured as a dictionary with the point estimates names as keys, like
dict(mean: tensor, std: tensor, ...) 
rather than a tensor from concatenating all individual estimates.

## Architecture

The architecture has one shared "body" network as well as separate "head" networks for each scoring rule.

The subnet keyword argument has a default of "mlp", in general the argument is resolved by `find_network` from the bayesflow.utils which can take a string for predefined networks or a user defined custom class.
* future extension direction: I made some pleasant experiences with using regression forests and boosted regression trees as point estimation tech for when a learned summary network is unnecessary. They are attractive for their robust performance and near instant fitting speed if applicable. sklearn and xgboost have mature implementations that are naturally not in pure keras3. Can we still somehow make it possible to use a preexisting random forest /tree boosting implementation?
Currently the non-shared networks are just linear layers. 
* Instead we could relax this and provide optional subnet_heads that can replace default linear heads. Since I can't currently see anyone having a preference against weight sharing I am avoiding the complexity for now.
* For some point estimates (scoring rules) it might be beneficial to specify a specialized activation function. For example exploiting that quantiles ordered by quantile level are always sorted since the CDF monotonously increases.

## Extendable design

The first draft only includes a PointInferenceNetwork subclass for quantile estimation, currently called called QuantileRegressor and found in bayesflow/networks/regressors/quantile_regressor.py.
This is just the first step, and we want to support different loss functions (which are typically called scoring rules in this context) that result in other point estimates.

A more flexible API including custom scoring rules /losses could use a `PointInferenceNetwork` that accepts a single or a sequence of `ScoringRule`s.
This can generate the appropriate number of heads in the `build()`method.
It also can pass the respective outputs to the corresponding scoring rules that compute their actual loss contributions and sum them up.

```python
PointInferenceNetwork.__init__(
	self,
	loss_func: ScoringRule = None,
	loss_funcs: Sequence[ScoringRule] = None,
	subnet: str | type = "mlp",
	**kwargs,  # e.g. subnet_kwargs = dict(widths=[64, 32], dropout=0.1)
)
```

A `ScoringRule` / `ScoringLoss` has a name (we want to distinguish multiple of them later) and a `score` / `compute` method.
* in the method score is meant as a verb
* it seems the literature has not converged on whether a score (noun) is always maximized (like the opposite of loss) or can be both -> terminology question

A `ScoringRule` could also compute a `prediction_shape` in the constructor to be accessed by generic code that generates a corresponding head for the scoring rule.

**choice we have here**
* `prediction_shape` could also be a `prediction_size`, an int rather than a tuple
* it is possible to score for example a whole covariance matrix or other multi-dimensional predictions, so allowing an arbitrary shape might be a good idea. 

The head would have `output_shape=(*batch_shape, *prediction_shape, num_variables)`
* e.g.: 
	* a batch_size of 32 -> batch_shape of `(32,)` and
	* a one dimensional point estimator -> `(1,)`
	* and 4 inference variables
	* -> `(32, 1, 4)` shall be head `output_shape`
	* build() creates a Dense layer with `32*1*4` units and reshape operation which together form a head

**interaction with choice of output of estimate method:** If we choose to allow multidimensional `prediction_shape`s rather than only `prediction_size`s, we can not concatenate output of different heads since each can have an individual `prediction_shape`.
Thus the decision interacts with whether the predict method should return tensors or a dict of tensors (see above).

Below are some notes how `ScoringRule` definitions could look like.

```python
class ScoringRule()
	def __init__(
		self,
		name: str = None,
		**kwargs,
	):
		...
	def score(self, prediction, realized_value):
		raise NotImplementedError

class NormedDifferenceLoss(ScoringRule)
	def __init__(
		self,
		name: str = "normed_difference",
		k: int = 2,  # results in an estimator for the mean
		**kwargs,
	):
		super().__init__(**kwargs)

		self.name = name
		self.k = k
		self.prediction_shape = (1,)
		
	def score(self, prediction, realized_value):
		pointwise_differance = prediction - realized_value[:,None,:]
		score = keras.ops.absolute(pointwise_differance) ** self.k
        score = keras.ops.mean(score)
		return score

class QuantileLoss(ScoringRule)
	def __init__(
		self,
		name: str = "quantile",
		quantile_levels: Sequence[float] = [0.1, 0.5, 0.9],
		**kwargs,
	):
		self.name = name
		self.quantile_levels = quantile_levels
		self.prediction_shape = (len(self.quantile_levels),)
		
	def score(self, prediction, realized_value):
        pointwise_differance = prediction - realized_value[:,None,:]

		score = pointwise_differance * (
	            keras.ops.cast(pointwise_differance > 0, float)  - self.quantile_levels[None, :, None]
        )
        score = keras.ops.mean(score)
		return score

```

If we want to support specific activation functions for different scoring rules, we might add a non-parametric activation function to the ScoringRule's definition.
We could also go all the way and have the **`ScoringRule` also contain the head itself**, taking the last joint embedding and mapping it to estimates. This then would naturally include an optional nonlinearity in the end and simultaneously give users the option to tweak the architecture of the non-shared weights.

Other notes: 
* For a PointInferenceNetwork that returns mean and std we *could* provide sample() and log_prob() under a Gaussian assumption. Maybe someone is interested in implementing even things like a LaplaceApproximator.
* later: is there anything to consider for point estimation to fit neatly with the `bf.Workflow`?


han-ol:
Find a notebook to look at training and inference for such a point estimator here: https://github.com/han-ol/bayesflow/blob/point-estimation/examples/draft-point-est.ipynb

han-ol:
After writing this up, in my opinion the best structure is probably to have:
* objects that specify both  a loss function _and_ a head (like proposed in the end of the first comment above),
* as well as a object constructed from a a sequence of those and bundles them together as one `PointInferenceNetwork`.

For class names I would propose:
The heads could be called by their estimator like `MeanEstimator`, `QuantileEstimator`, `CovarianceEstimator`, ... with a base class called `Estimator`.

If we want to support convenient creation of heads by just specifying a loss function, this can be done by subclassing a `BasicEstimator` with `prediction_shape.ndim=1` and a linear (or small mlp) architecture. Or even more succinctly by providing a corresonding macro.

Additionally, I would prefer an implementation where **estimate()** returns a dictionary with named estimates corresponding to the individual heads.

stefanradev93:
Hans, thanks for the great ideas and the very mature first draft! Here are some initial thoughts from my side. More to come later:

* What do you think about the possibility to inherit from `ContinuousApproximator` instead of `Approximator`? As far as I can think, the only conceptual differences are in `compute_metrics` and `estimate` vs `sample`.
* I wouldn't go for `sample` methods on the point estimators in order to keep their interface consistent. Even though it may make sense for certain kinds of parametric point estimators (e.g., heteroskedastic regression), one rarely needs anything beyond the distribution parameters in these cases.
* I would definitely go for named entities in the output of `estimate`. Perhaps its time to start thinking in terms of "data classes".
* I like the idea of an abstract `ScoringRule` class. I would call the arguments `targets` and `references` in keeping with the rest of the lib (e.g., diagnostics).
* The `WorkFlow` object would need to know whether it's dealing with a point estimator or a fully Baysesian approximator. It can check whether an `estimate` or a `sample` method is present, or check for a type (the latter may invalidate my point about inheriting from `ContinuousApproximator`. 
* The "non-parametric activation function" for scoring rules is a good idea, because it lets users specify heads with linear output layers (@paul-buerkner - we may call these "link functions"?).
* I would go for a single argument in the `PointInferenceNetwork` that can be either a single scoring rule or a sequence thereof.

paul-buerkner:
This looks really cool already! Thank you for this PR!

There is a lot of content here in this thread already and I may benefit from a call where you show me the current state. This would help me give reasonable feedback. I will contact you offline about it.

han-ol:
Ok cool, Paul! 

Stefan, thanks for your takes already! Some notes to some of them:
 
>* What do you think about the possibility to inherit from `ContinuousApproximator` instead of `Approximator`? As far as I can think, the only conceptual differences are in `compute_metrics` and `estimate` vs `sample`.

I am not sure about this. Tagging @LarsKue for this question, I think you mentioned a preference of parallel implementation rather than inheritance?

>* I would definitely go for named entities in the output of `estimate`. Perhaps its time to start thinking in terms of "data classes".

Agreed, that some form of naming is necessary. How do the "data classes" you imagine differ from dictionaries of the type:
```
# assuming a batch_size of 2 and 3 inference variables:
dict(
    mean=[[ 1, 2, 3], [2, 1, 3]],
    quantile_0.1=[[ -1, 0, 1], [1, -1, -2]],
    quantile_0.9=[[ 3, 4, 5], [3, 2, 5]],
    ...
)
```
For one thing, carrying the numeric value of the quantile level in the dict key is a bit awkward, but could be acceptable. 

>* I like the idea of an abstract `ScoringRule` class. I would call the arguments `targets` and `references` in keeping with the rest of the lib (e.g., diagnostics).

:+1: 
We need to decide on which class holds the heads' neural networks. It seems reasonable to bundle them with the ScoringRule directly, because there is a one-to-one relationship between head and ScoringRule.
If you prefer to keep the head out of an abstract `ScoringRule`, I would propose to have a `Head` object which owns both an abstract `ScoringRule` and a `keras.Layer`.

>* The `WorkFlow` object would need to know whether it's dealing with a point estimator or a fully Baysesian approximator. It can check whether an `estimate` or a `sample` method is present, or check for a type (the latter may invalidate my point about inheriting from `ContinuousApproximator`.

Ok! Just in their defence, I'd say point estimators are also fully Bayesian ;) Functionals of the proper Bayesian posterior distribution.

>* I would go for a single argument in the `PointInferenceNetwork` that can be either a single scoring rule or a sequence thereof.

Good point! Those `loss_func` and `loss_funcs` is a bit convoluted for no good reason.

stefanradev93:
Regarding the output, I think we can go with dictionaries for now. I believe some custom data classes will come in handy rather soon. 

Just a thought: Can the heads simply be determined automatically assuming the scoring rules know their dims?

han-ol:
Yes, they can mostly be inferred, and I'd suggest linear layers followed by a reshape as an overwritable default.

However, some scoring rules benefit from (e.g. quantiles, monotonously increasing) or need (e.g. covariance matrix, positive definite) a specific architecture.

stefanradev93:
Hi Hans, are we feature-complete here. Do I see it correctly that we are only missing a tutorial but the backend is all in place?

han-ol:
Hi Stefan, I consider the user inferface and internal structure complete.
The MultivariateNormalScore is not reliable yet, but since it is optional and the other `ScoringRule`s work reliably, we can merge with dev as soon as I merged the latest test fixture changes into the PR.

I finished a tutorial on Lodka-Volterra inference just now.

paul-buerkner:
Amazing! Where are the example notebook(s) located? I would love to see the feature in action :-)

stefanradev93:
You can safely link it as an example on the landing page.

han-ol:
Ok, I will add a link to the Lotka-Volterra to the readme.
Until that is live, here is a link that already works: https://github.com/han-ol/bayesflow/blob/point-estimation/examples/Lotka-Volterra.ipynb

We had briefly talked about showing the point estimators in the linear regression starter as well, so I added that too (currently without detailed discussion of the results, just brief explainer what point estimation is and why it might be useful.
https://github.com/han-ol/bayesflow/blob/point-estimation/examples/Linear_Regression_Starter.ipynb

vpratz:
Quick note: I prematurely activated building the docs on PRs (reverted now), sorry for that. If you can build the docs locally, it should be fine.

bayesflow/scores/scores.py
```diff
@@ -0,0 +1,276 @@
...
+        config["links"] = {
+            key: deserialize_value_or_type(link_dict, "link")["link"] for key, link_dict in config["links"].items()
+        }
+
+        return cls(**config)
+
+    def get_head_shapes_from_target_shape(self, target_shape):
+        raise NotImplementedError
+
+    def set_head_shapes_from_target_shape(self, target_shape):
```
LarsKue, line 63: Looks like this should be implemented with a property `self.target_shape` instead.

bayesflow/links/ordered.py
```diff
@@ -0,0 +1,50 @@
+import keras
+from keras.saving import register_keras_serializable as serializable
+
+from bayesflow.utils import keras_kwargs
+
+
+@serializable(package="links.ordered")
+class Ordered(keras.Layer):
```
LarsKue, line 8: Needs docs. I would also like to open a discussion on the name and visibility of the module `bayesflow.links`. Should users use this class?

bayesflow/links/positive_semi_definite.py
```diff
@@ -0,0 +1,17 @@
...
+
+class PositiveSemiDefinite(keras.Layer):
+    def __init__(self, **kwargs):
+        super().__init__(**keras_kwargs(kwargs))
+
+    def call(self, inputs):
+        # add identity to avoid coliniarity of input columns
+        inputs += keras.ops.identity(inputs.shape[-1])
+        # multiply M * M^T to get symmetric matrix
+        return keras.ops.einsum("...ij,...kj->...ik", inputs, inputs)
```
LarsKue, line 14: Don't use einsum in library code if you can avoid it.

han-ol:
Thanks for the review Lars! I will go through your comments individually with commits and/or responses.

bayesflow/networks/point_inference_network.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def __init__(
+        self,
+        scores: dict[str, ScoringRule],
+        subnet: str | type = "mlp",
+        **kwargs,
+    ):
+        super().__init__(
+            **keras_kwargs(kwargs)
+        )  # TODO: need for bf.utils.keras_kwargs in regular InferenceNetwork class? seems to be a bug
```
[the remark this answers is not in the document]
han-ol: Should we filter with keras_kwargs here? Then why should we not do it in bayesflow.networks.inference_network too?

bayesflow/networks/point_inference_network.py
```diff
@@ -0,0 +1,160 @@
...
+        self.subnet = find_network(subnet, **kwargs.get("subnet_kwargs", {}))
+
+        self.config = {
+            **kwargs,
+        }
+        self.config = serialize_value_or_type(self.config, "subnet", subnet)
+        self.config["scores"] = serialize(self.scores)
+
+    def build(self, xz_shape: Shape, conditions_shape: Shape = None) -> None:
+        if conditions_shape is None:  # unconditional estimation uses a fixed input vector
```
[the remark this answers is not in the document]
LarsKue: Interoperability is a good point, particularly because the intended use case for point inference networks is as a primer before using fully generative networks. I agree that we can leave it in until we decide if we should also disallow unconditional learning for generative networks.

bayesflow/approximators/point_approximator.py
```diff
@@ -0,0 +1,105 @@
...
+        self,
+        conditions: dict[str, np.ndarray],
+        split: bool = False,
+        **kwargs,
+    ) -> dict[str, dict[str, np.ndarray]]:
+        if not self.built:
+            raise AssertionError("PointApproximator needs to be built before predicting with it.")
+
+        conditions = self.adapter(conditions, strict=False, stage="inference", **kwargs)
+        conditions = keras.tree.map_structure(keras.ops.convert_to_tensor, conditions)
```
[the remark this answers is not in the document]
han-ol: Sure, I have no reservations against absorbing tensor mapping into the adapter. I would be in favor of doing this refactor independently to this PR after it is merged.

bayesflow/approximators/point_approximator.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+@serializable(package="bayesflow.approximators")
+class PointApproximator(ContinuousApproximator):
+    """
+    Defines a workflow for performing fast posterior or likelihood inference.
+    The distribution is approximated by point estimators with a feed-forward network and an optional summary network.
+    """
+
+    def estimate(
```
han-ol, line 19: My two cents about nesting complexity.

1. Inherent complexity of output

The output structure of PointApproximator.estimate should identify the variable names and the point estimate kinds. To be as close as possible to the ContinuousApproximator.sample output I am quite happy with a variable name major nesting.

Replacing tensors with their shapes this looks like the following
```
{'alpha': {'mean': (500, 1), 'quantiles': (500, 5, 1)},
 'beta': {'mean': (500, 1), 'quantiles': (500, 5, 1)},
 'gamma': {'mean': (500, 1), 'quantiles': (500, 5, 1)},
 'delta': {'mean': (500, 1), 'quantiles': (500, 5, 1)}}
```

To see it in context check the notebook
https://github.com/han-ol/bayesflow/blob/point-estimation/examples/Lotka_Volterra_point_estimation_and_expert_stats.ipynb

2. code readability / implementation complexity

Having the functionality in the first place requires a few computations that are nested. Maybe some nice utility functions can aid here, but before coming up with those I wanted to err on the side of not hiding how it works. If it not readable I should improve the comments. What do you think?

bayesflow/approximators/continuous_approximator.py
```diff
@@ -131,6 +131,50 @@ def get_config(self):
...
+    def estimate(
+        self,
+        conditions: dict[str, np.ndarray],
+        split: bool = False,
+        estimators: dict[str, callable] = None,
+        num_samples: int = 1000,
+        **kwargs,
+    ) -> dict[str, dict[str, np.ndarray]]:
+        estimators = estimators or {}
+        estimators = (
```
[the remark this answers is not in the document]
han-ol: The ContinuousApproximator.estimator method is (just) a convenience method to bring ease interoperability with the PointApproximator.estimate method.

Here, the nesting is irreducible if we want output to be the same as with the PointApproximator.estimate method.

I commented on nesting complexity over there (PointApproximator.estimate) since this is where the original estimation takes place and I believe there is the real question of whether to refactor with respect to nesting.

han-ol (commented):
I got to answer some of your comments, will answer the rest later.

bayesflow/scores/scores.py
han-ol: I don't think it is necessary or beneficial to have `self.target_shape` in a score, but I will get rid of the `set_head_shapes_from_target_shape`. This results in a slight refactor of PointInferenceNetwork.build and corresponding tests.

Maybe it is already clear, but I want to mention here, that target shape and a head's output shape are two different things.

bayesflow/networks/point_inference_network.py
```diff
@@ -0,0 +1,160 @@
...
+        output = self(x, conditions)
+
+        metrics = {}
+        # calculate negative score as mean over all scores
+        neg_score = 0
+        for score_key, score in self.scores.items():
+            score_value = score.score(output[score_key], x)
+            neg_score += score_value
+            metrics |= {score_key: score_value}
+        neg_score /= len(self.scores)
```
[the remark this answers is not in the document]
han-ol: For readability or something else?
keras.ops.mean(metrics.values()) is not supported. Also convert_to_tensor won't convert a dict values object.

How would you do it?

bayesflow/links/positive_semi_definite.py
han-ol: Ok! I would postpone refactoring this, since the MultivariateNormalScore with which it is used is currrently unstable, so changes are coming to this anyway.

bayesflow/approximators/continuous_approximator.py
```diff
@@ -131,6 +131,50 @@ def get_config(self):
...
+        samples = self.sample(num_samples=num_samples, conditions=conditions, split=split, **kwargs)
+
+        estimates = {
+            variable_name: {
+                estimator_name: func(samples[variable_name], axis=1) for estimator_name, func in estimators.items()
+            }
+            for variable_name in samples.keys()
+        }
+
+        def squeeze_dict(d):
```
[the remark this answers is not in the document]
han-ol: Done.

bayesflow/networks/point_inference_network.py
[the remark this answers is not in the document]
LarsKue: The reason is that you should not reinvent the wheel. This goes with respect to readability, but also the inevitable reintroduction of bugs.

This should do the trick:
`keras.ops.mean(list(metrics.values())`

tests/utils/assertions.py
```diff
@@ -33,3 +31,6 @@ def assert_layers_equal(layer1: keras.Layer, layer2: keras.Layer):
         x2 = keras.ops.convert_to_numpy(v2)
         msg = f"Variable '{v1.name}' for Layer '{layer1.name}' is not equal: {x1} != {x2}"
         assert keras.ops.all(keras.ops.isclose(x1, x2)), msg
+
+    # The names of layers need not stay the same
+    # assert layer1.name == layer2.name
```
LarsKue, line 36: When the network is serialized properly, they should. Otherwise, you end up with confusing output when something goes wrong. Why did you remove this test? It works for all other networks.

bayesflow/approximators/point_approximator.py
LarsKue: I am fine with doubly nested outputs as you show them, assuming they are necessary, and we cannot efficiently do something like `estimate_mean` and `estimate_quantiles` instead. However, internally, you have a triply nested dictionary here, and I don't yet see the reason for it.

tests/utils/assertions.py
han-ol: I just checked it again and it fails for . The names are sequential_103 and sequential_104. I don't know where the issue is in serialization or how to fix it.

But I do know that the networks outputs are correct (identical) since I have a test for that.

han-ol: Sure, I added it in the message above.

Only this test fails: https://github.com/han-ol/bayesflow/blob/point-estimation/tests/test_networks/test_point_inference_network/test_point_inference_network.py#L64

It explicitly calls `assert_layers_equal(net_head, loaded_head)` for the point inference network's head, which fails since the heads do not have the same name.

LarsKue: Thanks for the update. From what I can see, the code that is responsible for name assignment (calling `super().get_config()` and merging the config dicts and forwarding kwargs to `keras.Layer` in `__init__`) is correct. Can you add a test that simply compares the serialization dictionaries for the quantile estimation heads? The names should also be registered in those.

han-ol: Good suggestion, I checked and the names do not appear in `PointInferenceNetwork.get_config()`, because the heads are only created inside the build method based on the configs of the scores.

This turned out to become a learning experience about tweaking serialization details. I figured out multiple paths to resolve this. I'll write a description out here and also commit code, so you can take a look.
3 options:
1. Tolerate that the names of the heads are not kept by serialization.
2. Write custom `build_from_config` and `get_build_config` for only the `PointInferenceNetwork` that overwrite the name attribute of the heads only. This relies on build to create all state and copies over the correct (old) names, just as keras usually copies over the learnt weights after building the state.
3. Write custom `build_from_config` and `get_build_config` methods for **both** the `PointInferenceNetwork` and the heads.
This introduces some code duplication because `PointInferenceNetwork.build_from_config` cannot simply call build, and it requires heads to be of a new custom type with overwritten `build_from_config` and `get_build_config` methods.

I would prefer option 1 or 2, because for those the work is done. Option 3 is most invasive, but could be preferred for being more strict.

bayesflow/networks/point_inference_network.py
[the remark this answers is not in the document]
han-ol: Great, thanks!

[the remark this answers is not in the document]
han-ol: Done

bayesflow/approximators/point_approximator.py
han-ol: They are necessary.

Triple is necessary in the approximator, because an additional nesting is introduced by the adapter, just as in the `ContinuousApproximator`.

bayesflow/scores/scores.py
han-ol: Thanks for clarifying. Even though it is not what you are proposing: If you check out how it is now since commit [c2ca810](https://github.com/bayesflow-org/bayesflow/pull/281/commits/c2ca8100ead42181ca8956d48c2f2e7f5afe645f), do you see issues?

Seems to me we are fine without `target_shape` property.

stefanradev93:
This now looks great! @LarsKue upon a final quick check from you, I believe we can merge. Does the notebook include an example for saving / loading?

han-ol:
While adding a saving/loading example to the notebook I found an issue with serialization that was unfortunately not covered by tests: the quantile levels attribute in QuantileScore and the OrderedQuantiles objects is not properly serialized.
I'll try to fix it tomorrow.

han-ol:
[a55612e](https://github.com/bayesflow-org/bayesflow/pull/281/commits/a55612eadaffe8fe663d4b02dc4b43a55fa2aa72) contains a fix and test coverage for the serialization bug.

I also added a save and load example for the PointApproximator to the example notebook.

tests/utils/assertions.py
han-ol: We went with option 2, so I resolve the conversation.

Issue #228, opened 2024-10-28
Introduce better support for bypassing dispatch utils


LarsKue:
Currently, a lot of networks use string-based dispatch for internal architectures, e.g. [Single Couplings internal subnet](https://github.com/bayesflow-org/bayesflow/blob/7ef3612074b3880db330b5d29d17c76e8c82b69a/bayesflow/networks/coupling_flow/couplings/single_coupling.py#L22). Ideally, we should allow users to always pass serializable keras objects too.

This comes at the cost of having to define a custom `from_config` and `get_config` method for each object that allows this, but it would make BayesFlow much more extensible from the outside.

vpratz:
For the moment, we have decided to keep it limited to subnets, but if the need arises we will extend this to other arguments as well. If this is the case, we will discuss it in a new issue or reopen this one.

PR #284, opened 2024-12-13
Feature: Allow serialization of custom networks

vpratz:
This PR addresses #228. It introduces utility functions and extends existing networks to
enable serialization of complete networks when custom network types are
passed as arguments (e.g., for sub-networks in coupling flows).

The main complications were:

* Objects of type `type` (uninstantiated classes) cannot be serialized using `keras.saving.serialize_keras_object`, as the have no  `get_config` function. `keras.saving.get_registered_name` has to be used.

* We want to support both strings and types as parameters, leading to the need to distinguish those during manual serialization/deserialization.

* Auto-discovery of __init__ parameters is only active when `get_config` is not overridden, necessitating to manually store the configuration for serialization.

For storing the types, we use `keras.saving.get_registered_name`, which can be reconstructed at deserialization using `keras.saving.get_registered_object`.

Handling the different cases is moved the utility functions `(de)serialize_val_or_type`, which uses a naming scheme to determine which deserialization method to use.

The same setup can be extended to other custom types, e.g. distributions.


vpratz:
If you agree with the general design, I think the main question is for which parameters we want to use this. For now, I only implemented it for subnets, but it should be easily transferable to other parameters like distributions. Do we want to apply this broadly before merging this, or do we do it incrementally?

I am not totally happy with my adaptations to the test, but I did not see a better way to solve this. Simply adding the subnet fixture to all inference nets doubles the test time for the networks, approximately from 5min to 10min. As this is only relevant for serialization, I deemed this not acceptable and split it up, leading to a reduced runtime but less readable test code. Any ideas to make those tests prettier/easier maintainable are welcome.

paul-buerkner:
Thank you for working on this feature! I think it is very important to have! 

I am not the best person to ask about the internals so I am refering to @LarsKue and @stefanradev93 for a proper review.

stefanradev93:
Hi Valentin, I like the general design. I would only opt for a more verbose name of the utility: `(de)serialize_value_or_type`. I also agree with your choice to keep the runtime of the tests low. Do you like me to merge the PR or did I miss a certain TODO?

vpratz:
Hey Stefan, thanks for taking a look, I have renamed the functions. I think for this set of changes we are ready to merge. If you have time you can comment on the following:
For which parameters do we want to use this? We have it for subnets now, but we could also do it e.g. for distributions.

stefanradev93:
Thanks! I think the most pertinent usage will be for subnets. We can always enable it for distros if there is demand.

PR #285, opened 2024-12-17
Made Adapters Sliceable 

### LINKED ISSUE #274: Make Adapter sliceable
Currently, the `Adapter` is implemented as a series of transforms.

It would be nice if indexing or slicing the `Adapter` would yield another valid `Adapter`, in the form of:

```py3
adapter = bf.Adapter().rename("x", "y").rename("y", "z").rename("z", "y").rename("y", "x")
rename_x_to_y = adapter[0]
rename_x_to_z = adapter[0:2]

assert isinstance(rename_x_to_y, bf.Adapter)
```

We could also allow assigning an adapter to a slice then, in the form of:

```py3
adapter = ...
adapter[0:2] = bf.Adapter().rename("x", "z")
```

eodole:
I made the adapters sliceable but one concern I have is that when I ask for `generic_adapter[0:2]` this will return an entirely new adapter instance. Depending on the use case this one matter but could cause problems. I also made a test file but it's not tracked by github, please let me know if I should also upload that. 

I also noticed that ElementwiseTransform wasn't being imported so added that as well. 
Resolves #274

### DIFF
--- a/bayesflow/adapters/adapter.py
+++ b/bayesflow/adapters/adapter.py
@@ -78,6 +78,27 @@ class Adapter:
     def __repr__(self):
         return f"Adapter([{' -> '.join(map(repr, self.transforms))}])"
 
+    def __getitem__(self, index):
+        if isinstance(index, slice): 
+            sliced_transforms = self.transforms[index]
+            print("Are the sliced transforms a sequence")
+            print(isinstance(sliced_transforms, Sequence))
+            print("Is there an associate print method?")
+            print(sliced_transforms)
+            
+            new_adapter = Adapter(transforms = sliced_transforms)
+            return new_adapter
+        elif isinstance(index, int): 
+            if index < 0:
+                index = index + len(self.transforms) # negative indexing 
+            if index < 0 or index >= len(self.transforms): 
+                raise IndexError("Adapter index out of range.")
+            sliced_transforms = self.transforms[index]
+            new_adapter = Adapter(transforms = sliced_transforms)
+            return new_adapter
+        else:
+            raise TypeError("Invalid index type. Must be int or slice.")
+    
     def add_transform(self, transform: Transform):
         self.transforms.append(transform)
         return self
@@ -104,6 +125,7 @@ class Adapter:
         self.transforms.append(transform)
         return self
 
+    # Begin of transformed derived from transform classes 
     def as_set(self, keys: str | Sequence[str]):
         if isinstance(keys, str):
             keys = [keys]
--- a/bayesflow/adapters/adapter.py
+++ b/bayesflow/adapters/adapter.py
@@ -112,28 +112,43 @@ class Adapter:
         
     
     def __setitem__(self, index, new_value): 
+
+        if not isinstance(new_value, Adapter): 
+            raise TypeError("new_value must be an Adapter instance")
         
+        
+        new_transform = new_value.transforms 
+        
+        if len(new_transform) == 0: 
+            raise ValueError("new_value is an Adapter instance without any specified transforms, new_value Adapter must contain at least one transform.")
+
+
         if isinstance(index, slice): 
             if index.start > index.stop: 
                 raise IndexError("Index slice must be positive integers such that a < b for adapter[a:b]")
+            
             if index.stop < len(self.transforms):
-                new_transform = new_value.transforms 
-                # print("what is self.transforms[index]?")
-                # print(self.transforms[index])
-                # print("what is the value of the newvalue")
-                # print(new_transform)
-                # print(type(new_transform))
                 self.transforms[index] = new_transform
-                # else raise theory 
+            
             else: 
                 raise IndexError("Index slice out of range")
             
+
         elif isinstance(index, int): 
-            return 
-            # check if in range 
-            # if not inrange but it is just the len of the transforms (append )
-            # 
-            # else raise error 
+            if index < 0: # negative indexing 
+                index = index + len(self.transforms)
+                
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thank you! I will take a look soon!

paul-buerkner:
A few short questions:

- Where do you anticipate issues with slicing creating new adapter instances?
- A test file (if set up in the appropriate test folder etc.) would be great and should be added to this commit I think.
- Index numbers in the printing would be great! How much effort would it be for you to add this?
- I don't fully understand what you mean by "appending a new transform via slice assignment so if index == len(adapter.transforms) then simply use the add_transform method." Can you elaborate? Also perhaps use some example code how what you have in mind would look?

eodole:
Essentially what i meant with my last comment is that you can add a transform to the end of the adapter by slicing. 
I have to find the code responsible currently for printing but I think the modification should be pretty easy

paul-buerkner:
Ah, I see. Could you show a code example? Let me know once you have added everything you wanted to add to this PR.

LarsKue:
Appending via slicing is fine, imo. We should have checks in place and error if slices are assigned non-Adapter objects (thus making the adapter self-recursive, each slice is an Adapter itself).

That said, best practice should be to use add_transform (could just call this append as well). We could add an insert method to mimic list behavior, too.

bayesflow/adapters/adapter.py
```diff
@@ -76,7 +76,68 @@ def __call__(self, data: dict[str, any], *, inverse: bool = False, **kwargs) ->
         return self.forward(data, **kwargs)
 
     def __repr__(self):
-        return f"Adapter([{' -> '.join(map(repr, self.transforms))}])"
+        str_transf = ''
```
LarsKue, line 79: Use double quotes for strings to conform to our style guide.

bayesflow/adapters/adapter.py
```diff
@@ -76,7 +76,68 @@ def __call__(self, data: dict[str, any], *, inverse: bool = False, **kwargs) ->
...
+        for i in range(0, len(self.transforms)):
+            str_transf = str_transf + str(i) + ': ' + repr(self.transforms[i]) 
+            if i != len(self.transforms) - 1:
+                str_transf = str_transf + ' -> ' 
+        return f"Adapter([{str_transf}])"
+
+    def __getitem__(self, index):
+        if isinstance(index, slice):
+            if index.start > index.stop:
+                raise IndexError("Index slice must be positive integers such that a < b for adapter[a:b]")
```
LarsKue, line 89: Should just return an empty adapter.

bayesflow/adapters/adapter.py
```diff
@@ -76,7 +76,68 @@ def __call__(self, data: dict[str, any], *, inverse: bool = False, **kwargs) ->
...
+            if index.stop < len(self.transforms):
```
LarsKue, line 90: I don't think this works for index.stop = None, which happens when you call adapter[start:]

Best would be to just mimic list behavior via returning Adapter(self.transforms[index])

bayesflow/adapters/adapter.py
```diff
@@ -76,7 +76,68 @@ def __call__(self, data: dict[str, any], *, inverse: bool = False, **kwargs) ->
...
+                sliced_transforms = self.transforms[index]
+                new_adapter = Adapter(transforms=sliced_transforms)
+                return new_adapter
+            else:
+                raise IndexError("Index slice out of range")
```
LarsKue, line 95: We can let the internal list handle this

bayesflow/adapters/adapter.py
```diff
@@ -76,7 +76,68 @@ def __call__(self, data: dict[str, any], *, inverse: bool = False, **kwargs) ->
...
+
+        elif isinstance(index, int):
+            if index < 0:
+                index = index + len(self.transforms)  # negative indexing
+            if index < 0 or index >= len(self.transforms):
+                raise IndexError("Adapter index out of range.")
+            sliced_transforms = self.transforms[index]
+            new_adapter = Adapter(transforms=sliced_transforms)
+            return new_adapter
```
LarsKue, line 104: Same idea here, let the internal list handle edge cases

bayesflow/adapters/adapter.py
```diff
@@ -76,7 +76,68 @@ def __call__(self, data: dict[str, any], *, inverse: bool = False, **kwargs) ->
...
+        else:
+            raise TypeError("Invalid index type. Must be int or slice.")
+
+    def __setitem__(self, index, new_value):
```
LarsKue, line 108: Seems a bit verbose, does the following snippet not work?

```py3
self.transforms[index] = new_value.transforms[:]
```

Or for an integer index:

```py3
self.transforms[index:index + 1]  = new_value.transforms[:]
```

bayesflow/adapters/adapter.py
```diff
@@ -104,6 +165,7 @@ def apply(
         self.transforms.append(transform)
         return self
 
+    # Begin of transformed derived from transform classes
```
LarsKue, line 168: I don't understand this comment.

LarsKue (commented):
Looks good overall, see individual comments for potential changes and discussion.

bayesflow/adapters/adapter.py
eodole: oh this was just a comment for myself so I could see when i was scrolling if I had gone too far.

paul-buerkner:
@eodole Did you address all the comments by lars already? If yes, can you resolve them? If not, can you make the necessary edits? Let me know if you have questions.

PR #291, opened 2025-01-14
Add Rational Quadratic Spline Transforms to Normalizing Flows

LarsKue:
See title. As a side-effect, this PR includes an update to the interface for backend-agnostic Jacobian computation (and related functions, such as `vjp` and `jacobian_trace`), previously discussed in #251.

The set transformer is failing some tests in my local env, but I assume that is unrelated to this PR.
The linter also appears to not be happy with `dict_utils.py` which is unchanged by this PR.


bayesflow/utils/tensor_utils.py
```diff
@@ -186,9 +186,8 @@ def searchsorted(sorted_sequence: Tensor, values: Tensor, side: str = "left") ->
         case "tensorflow":
             import tensorflow as tf
 
-            out_type = "int32" if len(sorted_sequence) <= np.iinfo(np.int32).max else "int64"
-
-            indices = tf.searchsorted(sorted_sequence, values, side=side, out_type=out_type)
+            # always use int64 to avoid complicated graph code
```
LarsKue, line 189: This is not backend agnostic, as jax forces this kind of behavior.

PR #293, opened 2025-01-20
Add notebook showcasing two-level components

noahg2:
Hi @stefanradev93 - this contains the notebook I mentioned in [our discussion](https://discuss.bayesflow.org/t/time-series-dimension-errors-in-hierarchicalnetwork/122/3) on the forum.

To be transparent, I would describe myself as being in the process of self-teaching both the library and Bayesian methods in general, so if you see something that looks like a newb mistake please point that out. That being the case, I'd be curious to know if you think the approach to validation here is reasonable. I'm not sure, for instance, if it entirely makes sense to compare the latent distributions computed for the subgroups against each other.


stefanradev93:
Thank you, Noah! I will have a look later today and provide feedback, if needed. :)

stefanradev93:
Looks pretty good to me for a minimal example.

stefanradev93:
Can you also add the notebook link to the readme file (as with the other examples) before merging?

noahg2:
Thanks for the feedback! I made some updates in response - please let me know if you think additional edits are necessary.

PR #295, opened 2025-01-27
Fix NormalSimulator broadcasted output shapes

han-ol:
The `NormalSimulator` used for testing seems to have unintended output shapes.

The `mean` and `std` had repeated values, corresponding to the number of observations.

The PR:
1. undoes the repetition by slicing
2. flattens the observations

The latter could be avoided if a summary network or adapter transform is chosen.


vpratz:
Good spot, thanks for the fix! I refactored it a bit to use automatic broadcasting, please take a look and let me know if you are happy with those changes.

han-ol:
Thanks, the refactor looks good. Happy!

PR #299, opened 2025-02-04
Example Notebook for Users of 1.0 moving to 2.0

eodole:
This is the first draft, I still need to make the validation section and have a few sections with ??? where there is a change that I feel should be explained but I'm not sure why that decision was made. I also have a section with additional features in which additonal features could be highlighted that are not included in this tutorial. 
There are also several points in which is say "refer to the documentation" which I feel could be more specific but Im not sure if I should point to papers, or specific parts of the api or what exactly.


LarsKue (changes requested):
The adapter changes made their way into this PR. @eodole can you make a new PR with just the notebook changes, please?

Issue #292, opened 2025-01-16
Batched Dynamic Step Size Integration for Differential Equation Models


LarsKue:
Methods like Flow Matching and Diffusion Models use differential equations to integrate their output velocity from latent to sample space. For a fixed step size, this is non-problematic, as we perform $n$ function calls for the integration:

$$ x_{t_1} = \int_{t = t_0}^{t = t_1} f(x_{t}; t) \text{d}t$$

Dynamic step size is trickier, since the exact convergence criterion and update method is well-defined only for a single sample, not for a batch. The step size is computed and adjusted on the fly during integration, and may differ across a batch, which means we may no longer have a set number of function calls $n$ to perform.

With this issue, I intend to open a discussion on the optimal method we should implement for BayesFlow. Here are some options we have:

### Suggested Methods

1. `vmap` over $f$ and perform $n_i$ calls to $f$ for each sample $i$ in the batch
2. Update each entry as normal and ignore updates to finished parameters once their respective time reaches $t_1$
3. Consolidate the batch of step sizes into one step size that is used across the whole batch (e.g. the minimum).
4. ...? There are probably more methods that I cannot think of. Contributions welcome.

### Benefits and Drawbacks

1. removes the problem of batches entirely. However, it may be wasteful if the `vmap` is inefficient, or the batch size is very large. Simultaneously, this performs the minimum number of calls for each sample, so could be more efficient than the other methods.
2. is harder to implement, but does not have the performance downsides (or benefits) of `vmap`. It may also be wasteful if a single sample in the batch requires significantly more steps than all the others.
3. is easiest to implement, but presents us with an arbitrary choice of consolidation function, which is critical for good performance. The minimum may be too pessimistic if we can just discard bad samples, for instance. The biggest benefit here is that we can use the required redundancy of repeated calls to `f` to further improve the quality of samples.

If anyone knows of any scientific publications that cover the problem of batched dynamic step size integration, I would be glad to read them.

stefanradev93:
@arrjon What do you think after revamping the integrators?

arrjon:
On the low-dimensional benchmarks of Lueckmann, our adaptive samplers on the batch (so a joint step for all samples) took almost the same number of steps as for single samples. I think there is no need for batched dynamic step size integration (at least not at the moment).

PR #300, opened 2025-02-06
Improved Integrators

### LINKED ISSUE #288: Propagate training flag to FlowMatching Integrator
Currently, dropout is ignored in FlowMatching. 
I confirmed that the corresponding subnet_kwargs is propagated correctly to the subnet constructor. Nevertheless, dropout has no effect on training. 

The Integrator and its subnet needs to receive a training=True flag to turn on dropout at training time.

LarsKue:
This PR:

- Introduces integrator methods for sampling with Flow Matching
  - This is only proof of concept for now, so we are limited to Euler and Runge-Kutta integration.
- Allows dynamic step size selection
  - The initial step size is currently chosen as a constant. This needs to be updated, still.
- Allows choosing the integrator method at sampling time
  - This means we are no longer locked into a single algorithm with the fitted network.
- Temporarily addresses #292 by implementing step size consolidation.
- Fixes #288

The current code for integration is still a little messy due to lack of proper encapsulation. However, we can update this as we go without much impact to the user, if any.

I tested this running Flow Matching on the Two Moons example, and on some toy examples, both in non-compiled and compiled contexts.


LarsKue:
@stefanradev93 In my tests, RK45 runs in slightly fewer function evaluations with dynamic step size on. For a static step size, I would still choose Euler, since Flow Matching tends to have straight trajectories. But we can leave RK45 as the default, I think, or even implement more algorithms.

That said, there is still an issue somewhere, presumably with negative step sizes (when `start_time > stop_time`). I will mark the PR as ready for review again when I have fixed this.

Issue #256, opened 2024-11-21
Minor improvements to consistency model interface


paul-buerkner:
I have just played around a bit with consistency models, and gather some minor observations here. I may add more points later on.

- [ ] `ConsistencyModel` has an argument `total_steps` without default, which is suggested to be set to `epochs * num_batches` in the doc. Can we set this argument this internally automatically by default? This would make the code a bit cleaner. Also it should be clarified what happens if we set it to a different value than `epochs * num_batches`.
- [ ] `ContinuousConsistencyModel` is a dangerous name for an architecture different to `ConsistencyModel`. Since we also have a `ContinuousApproximator`, users may think they need to combine it `ContinuousConsistencyModel` or that `ConsistencyModel` is not for continuous variables. Can we perhaps rename `ContinuousConsistencyModel` or even make it part of the `ConsistencyModel` class to be activated via an argument like `continuous_time = TRUE`?

vpratz:
Thanks for taking a look at this.

- Regarding `total_steps`: This controls the discretization schedule during training, so we have to know where the scheduling ends. The last step of the training is the natural choice, but you could also end earlier. I don't know whether we want to suggest this though, but we certainly can document it. Determining this automatically would have to happen in `fit`, which would add some complexity to the approximators. I'm a bit hesitant to do this, because it would affect all our networks and training methods, and communicating what happens implicitly is a lot harder than to explain what to do to specify it explicitly. As a side-note: we do not need this parameter for the continuous-time CMs. If they should perform sufficiently well (which we still have to test), we could propose them as default and only power-users will encounter this difficulty.

- Regarding the naming: Yes, I share that concern. The hard part here is that it mainly concerns the way it is trained (discrete-time vs continuous-time), and not all the other meanings discrete and continuous can apply to (not even the sampling, which is discrete-time for both of them). We could also use a different axis of distinction (like `ConsistencyModel` and `SimplifiedConsistencyModel`/`ImprovedConsistencyModel`), but I did not yet come up with one that is intuitive, so any ideas are welcome :) Combining them in one class might work, but might only move the problems one level lower. We still have to explain what it means to activate `continuous_time`, and in addition make it explicit in the doc string which arguments are used for which method. Code-wise, having two classes instead of if-clauses in most functions improves readability a lot.

vpratz:
[...]

PR #301, opened 2025-02-07
Move continuous-time consistency model to experimental module.

vpratz:
Related to #256

As we have not extensively tested their behavior and I have observed subpar performance in some settings, we move this model class to experimental for further testing, to see whether it is ready for broader use, or whether it turns out to be too unstable to be practical.


vpratz:
@stefanradev93 How much structure do we want in the experimental module? For now I just used a single file inside the module, but if we want a more nested structure, I will move the files around accordingly.
How do we deal with examples/preliminary tests of experimental functionality? Do we want to include an `experimental` folder in the `examples` directory?

stefanradev93:
I think the structure is fine for now. As you suggest, I would include an `experimental` examples that are not visible on the landing page.

vpratz:
Thanks for taking a look! have made the corresponding change...

Issue #307, opened 2025-02-11
Rethinking names "targets" and "references"


han-ol:
How should we call the output of Approximators?

The current choice, targets, is confusable with the true values which we are currently calling references (e.g. in diagnostics).

han-ol:
Current ideas that could replace `targets`: `predictions`, `approximations`, ... ?

Both are not typically used for generative networks. `approximations` seems more suited since every model is an approximation in some sense.
We can also call it `samples` and `estimates` for generative and point inference networks respectively.

`samples` is not really specific to samples from the approximator, but at least short. If next to `references`, it is quite obvious would would be meant.

If we are looking for a general name that includes both output of generative networks and point estimates, `samples` is not available.

Please mention your perspectives and ideas!

stefanradev93:
Hey Hans, I am starting to agree with it. On a second read, this collides with `PyTorch`'s documentation that uses `targets` to denote ground truths:

https://pytorch.org/docs/stable/generated/torch.nn.functional.cross_entropy.html

paul-buerkner:
I agree with you too although I was involved in the original target and reference naming :D

I currently feel that `estimates` for the network output and `targets` for the ground truth could be sensible. I would prefer `approximations` over `estimates` in principle but `approximations` is too long for my taste.

Issue #309, opened 2025-02-11
Diagnostics in example notebooks are broken


han-ol:
Currently, the diagnostics plots in the notebooks **Linear_Regression** and **Quickstart_Amortized_Posterior_Estimation** are broken since they use a discontinued API.

I was not able to provide a quick fix for this since some functionality for filtering and renaming keys that the notebook in its current state expects seem to be missing.

I made a more specific issue regarding the problematic functionality (https://github.com/bayesflow-org/bayesflow/issues/308), but with the nearing release the notebooks are important in their own right.

stefanradev93:
@paul-buerkner @jerrymhuang  Let's unify the regression notebook and the quickstart, making the regression nb the new quickstart nb?

paul-buerkner:
Will look at this today. So the goal is to extend the linear regression example with the quickstart stuff and then remove the old quickstack notebook?

Issue #308, opened 2025-02-11
Diagnostics that can filter keys and configure pretty names


han-ol:
Should we modify `prepare_plot_data()`  to provide two separete arguments, one to filter keys and one to rename them?
Should we provide renaming functionality as a dictionary that defines the renaming directly?

If I want to have pretty variable names with latex, I cannot passing the posterior draws as a dict and override the names.

I'll provide a concrete example to show some of the complexity we have to deal with here:

Suppose I want to plot diagnostics based on a dictionary of posterior draws, which has the following keys and value shapes

```python
$ keras.tree.map_structure(keras.ops.shape, pdraws)
{'beta': (200, 1000, 2), 'sigma': (200, 1000, 1)}
```

During preparation of the  data all diagnostic plots will call `bf.utils.split_arrays`

beta's last dimension has length 2, so it is split, resulting in 
```python
$ keras.tree.map_structure(keras.ops.shape, bf.utils.split_arrays(pdraws))
{'beta_1': (200, 1000), 'beta_2': (200, 1000), 'sigma': (200, 1000)}
```

So we have to think about at which point do we apply the filtering? Before or after the split?

Also, at which point do we apply the renaming? Is it maybe best to first change the names pre-split, and have a latex-aware split rename function?
* convert "beta" -> "$\beta$", e.g. with a user provided dictionary
* when split is called, insert the `_1` or `_2` **before** the last `$` sign if one is present, rather than attach it to the end?

stefanradev93:
>Should we modify prepare_plot_data() to provide two separate arguments, one to filter keys and one to rename them?

Yes, I think so. This functionality was removed during the revamp, but it can be quite useful.
>Should we provide renaming functionality as a dictionary that defines the renaming directly?

Yes.

For Latex-friendly names, we can simply use a string like `rename="latex"` as an alternative to a dictionary, performing:

>when split is called, insert the _1 or _2 before the last $ sign if one is present, rather than attach it to the end?

Otherwise, I would rename **before** the split. What does @LarsKue think?

LarsKue:
`rename=latex` sounds like a dream come true, but I am not sure how compatible we can make it.

As a more modern and probably easier alternative, we could go with typst:
https://github.com/daskol/mpl-typst
https://github.com/messense/typst-py

Here, `beta` would already render as `$\beta$` would in Latex.

paul-buerkner:
[...]

PR #310, opened 2025-02-12
Make linear regression the quickstart notebook

paul-buerkner:
This PR addresses #309 by making the linear regression notebook the quickstart notebook and removing the current WIP quickstart notebook.

If few things we need to do before this can be merged:

- [x] Improve diagnostics plot interfaces as discussed in #307 and #308. Currently, its awkward to select multiple parameters for some of the same plots. I would like to have a simpler parameter selection interface. This notebook could be the test bed.
- [x] Use flow matching instead of coupling flow. I had some errors during inference on my mac this morning so I switched just to run it for now. (Was an issue with JAX and Mac GPUs)


paul-buerkner:
I have made some progress in making the diagnostic plots prettier and easier to use. Specifically, I added back the `filter_keys`. We should also discuss the name of this but I will make more edits first.

I know it's not super clean that I will put all of this in one PR but I don't have the time to make both the starter notebook pretty and fix all the plots at the same time if they are not on the same branch.

paul-buerkner:
This PR is now ready for review. It closes #307, #308, and #309. 

@stefanradev93 would you mind taking a look and merging if you are happy?

I left some remaining ToDos in the code, which would further improve the plots but are not essential at the moment. I will take care of them at a later point.

LarsKue:
Made some adjustments with feedback from @paul-buerkner. IMO, the notebook looks great. Ready to merge from my side @stefanradev93.

PR #319, opened 2025-02-14
Fix explanation in linear regression notebook

han-ol:
Previously mentioned modules having been imported that had not been. Slight rephrasing for reading flow.


paul-buerkner:
thanks!

PR #321, opened 2025-02-18
introduce VariableArray class

paul-buerkner:
This PR further improves the diagnostics module:

- Introduces a basic VariableArray class to carry around variable_keys and variable_names with an array containing posterior draws. This class is only meant for internal use for now and we can see if extending it in the future could be sensible.

- Further improves posterior pairs plots to enable plotting the prior again.

For example, using the linear regression notebook, we can now do something like:

```python
bf.diagnostics.plots.pairs_posterior(
    estimates=post_draws, 
    targets=val_sims,
    priors=sim_draws,
    dataset_id=0,
    variable_names=par_names,
)
plt.show()
```

to show both posterior, prior, and the true parameter values. The previous implementation didn't quite work in terms of correctly showing the prior.

There is one issue which I couldn't quite resolve and I would like to get your input @stefanradev93 and @jerrymhuang: Currently, the posterior is always plotted behind the prior on the diagonal. I tried to change that via `hue_order` for example but it either breaks the off diagonal plots or mislabels prior and posterior. It kind of looks like a bug in seaborn to me as I cannot imagine this to be intended. Any ideas would be highly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,9 +94,9 @@ conda env create --file environment.yaml --name bayesflow
 Check out some of our walk-through notebooks below. We are actively working on porting all notebooks to the new interface so more will be available soon!
 
 1. [Linear regression starter example](examples/Linear_Regression_Starter.ipynb)
-2. [Two moons starter example](examples/Two_Moons_Starter.ipynb)
-3. [SIR model with custom summary network](examples/SIR_Posterior_Estimation.ipynb)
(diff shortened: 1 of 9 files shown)


paul-buerkner:
I think I now fixed also the histogram issue to the degree that seaborn allows. I think this PR is ready to merge once @stefanradev93 approves.

PR #326, opened 2025-02-19
Fix base distribution in deep set

### LINKED ISSUE #325: Cannot set a base distribution of a summary network (deep set)
if I understand it correctly, it should be possible to optionally specify a base distribution for a summary network, e.g.

```{python}
summary_network = bf.networks.DeepSet(base_distribution="normal")
```

but that gives a `ValueError: Unrecognized keyword arguments passed to EquivariantModule: {'base_distribution': 'normal'}`

LarsKue:
Fixes #325


LarsKue:
Accidentally targeted main branch.

PR #332, opened 2025-02-21
Feature: allow use of Keras functional API by sanitizing build input shapes

### LINKED ISSUE #329: Nework build pattern using `self.call` not compatible with Keras functional API
When using the Keras functional API, the networks are automatically built. This includes passing a shape with the batch_size specified as `None`.

``` python
inputs = keras.layers.Input(shape=(10, 2))
outputs = bf.networks.DeepSet()(inputs)
model = keras.Model(inputs=inputs, outputs=outputs)
```

This errors in the second line with the following message (TensorFlow backend):

```
[bayesflow/networks/deep_set/deep_set.py#line=82), in DeepSet.build(self, input_shape)
     81 def build(self, input_shape):
     82     super().build(input_shape)
---> 83     self.call(keras.ops.zeros(input_shape))

ValueError: Attempt to convert a value (None) with an unsupported type (<class 'NoneType'>) to a Tensor.
```

Do we want to support using the networks in the functional API? If so, we should extend the `self.call(keras.ops.zeros(input_shape))` by an appropriate check for None in the first dimension, and replace it with a dummy batch size.

vpratz:
Fixes #329.

This PR adds a decorator `sanitize_input_shape` to replace the first entry of input_shape with a dummy batch size of 32 in case it is None.
The decorator is added to the relevant network build functions.

If you have any ideas/wishes regarding naming, please let me know and I'll adjust it.

### DIFF
--- a/bayesflow/networks/deep_set/deep_set.py
+++ b/bayesflow/networks/deep_set/deep_set.py
@@ -5,6 +5,7 @@ from keras.saving import register_keras_serializable as serializable
 
 from bayesflow.types import Tensor
 from bayesflow.utils import filter_kwargs
+from bayesflow.utils.decorators import sanitize_input_shape
 
 from .equivariant_module import EquivariantModule
 from .invariant_module import InvariantModule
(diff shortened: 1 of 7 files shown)


LarsKue (approved):
Looks good to me. Ideally, in the long term, we should move away from the (expensive and unreliable) way of using `self.call` in `self.build`, though.

vpratz:
Thanks for the quick review!

Issue #328, opened 2025-02-20
Some inference nets do not allow `inference_variables` during sampling


Kucharssim:
I am on this commit: https://github.com/bayesflow-org/bayesflow/commit/1dbaedd606bdbfa7f9af77830c102ad03dcb54ee

It should be possible to easily obtain posterior samples using a dataset from a simulator, e.g.

```python
dataset = simulator.sample(1000)
posterior_samples = amortizer.sample(num_samples=500, conditions=dataset)
```

Some inference nets however do not play nice if `inference_variables` are in the input dict during sampling. Specifically,

- Flow Matching

```
TypeError: Exception encountered when calling FlowMatching.call().

rk45_step() got an unexpected keyword argument 'inference_variables'
```

This error is independent of which integrator we select (presumably, I just only tried euler as an alternative)

- Consistency Model

```
ValueError: In a nested call() argument, you cannot mix tensors and non-tensors. Received invalid mixed argument: kwargs={'density': False, 'inference_variables': Array([[-0.19996756],
       [ 0.17675304],
       [ 0.27663922],
       [-1.2399167 ],
       [ 1.1492295 ],
       [-1.2453966 ],
       [ 0.35556716],
       [-1.3496909 ],
       [ 0.14714664],
       [ 0.46096817]], dtype=float32)}
```

For both nets, if we remove inference variables from the conditions, they run fine.

For affine and spline coupling networks, it runs fine.

Full code:

```python
def prior():
    mu = np.random.normal(loc=0, scale=1)
    return dict(mu = mu)

def likelihood(mu):
    x = np.random.normal(loc=mu, scale=1)
    return dict(x=x)

simulator = bf.make_simulator([prior, likelihood])


workflow = bf.BasicWorkflow(
    simulator=simulator,
    # uncomment whichever network you want to test
    #inference_network=bf.networks.FlowMatching(),
    #inference_network=bf.networks.CouplingFlow(),
    #inference_network=bf.networks.CouplingFlow(transform="spline"),
    #inference_network=bf.networks.ConsistencyModel(total_steps=100),
    inference_variables="mu",
    inference_conditions="x"
)

h=workflow.fit_online(epochs=1)

dataset=simulator.sample(10)

# uncomment the next line to make flow matching and consistency model run
# dataset=dict(x=dataset["x"])

posterior_samples=workflow.sample(num_samples=100, conditions=dataset)
```

paul-buerkner:
[...]

PR #333, opened 2025-02-21
fix issue #328

paul-buerkner:
This PR fixes issue #328 by removing `inference_variables` from conditions in the continuous approximator before passing conditions to the networks. Taggine @Kucharssim to double check that this fixes the issue and @LarsKue for reviewing this PR.


paul-buerkner:
Ah yeah, will change this. Thank you!

paul-buerkner:
Changed.

LarsKue:
@paul-buerkner Let's also add a comment to the Model Comparison line, so it doesn't confuse readers.

paul-buerkner:
done.

Kucharssim:
Yes, this PR fixes the issue :)

PR #334, opened 2025-02-21
`expected_calibration_error` does not crash with `VariableArray` objects

Kucharssim:
PR https://github.com/bayesflow-org/bayesflow/pull/321 broke the ece because the function made an incorrect assumption that if the inputs are not `type` of `np.ndarray`, they must be `tf.Tensor` with a `.numpy()` method. 

The fix can be tested against the model comparison example notebook.

### DIFF
--- a/bayesflow/utils/comp_utils.py
+++ b/bayesflow/utils/comp_utils.py
@@ -1,4 +1,5 @@
 import numpy as np
+from keras import ops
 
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Looks good. Thank you! Waiting for checks to pass and will then merge.

Issue #290, opened 2024-12-28
Separate private from public API: Underscore convention or `__all__`


vpratz:
This is related to the documentation issue #280, where we only want to render the public API documentation, but goes a bit further.

In many cases, our current setup indicates a desired import structure (e.g. importing `bayesflow.transforms.Transform` instead of `bayesflow.transforms.transform.Transform`. This is intuitive, but it is implicit and we cannot automatically (for generating the API docs) tell that the latter is not intended usage and should be private.

As far as I can tell, there are two usual approaches:
a) prepending private modules and files with an underscore (e.g., `bayesflow.transforms._transform.Transform`)
b) adding public functions, classes and modules to `__all__`, while excluding all private ones

I think for Sphinx to work efficiently, we probably need b), either everywhere where we import functions that we want to make publicly availiable (when `autosummary_imported_members = False`), or everywhere where we use external modules (keras, scikit-learn, ...) with `autosummary_imported_members = True`.

One problem I encountered when trying this out is Python's peculiar behavior with relative imports. When doing

``` python
from .transform import Transform
```

we would expect that one variable is added to the namespace, `Transform`. But it turns out two objects are added: `transform` and `Transform`. The reason for this is explained in [this thread](https://discuss.python.org/t/why-do-relative-from-imports-also-add-the-submodule-itself-to-the-namespace/24440).
For us, this means that without explicitly specifying `__all__` or renaming `transform` to `_transform`, both are detected as public, even though our intention was to only make `Transform` publicly available.

We can do b) semi-automatically, by populating `__all__` by a function like the following:

``` python
import inspect


def _add_imports_to_all(exclude=None):
    """Get global dict from stack."""
    exclude = exclude or []
    calling_module = inspect.stack()[1]
    local_stack = calling_module[0]
    global_vars = local_stack.f_globals
    if "__all__" not in global_vars:
        global_vars["__all__"] = []
    all_var = global_vars["__all__"]
    global_vars["__all__"] = [
        e for e in list(set(all_var).union(set(global_vars.keys()))) if e not in exclude and not e.startswith("_")
    ]
```

This would allow us to exclude unwanted relative imports (or, by modifying it a bit, only to add functions and classes, while manually specifying modules we want to add). If we also have a), we can exclude them automatically, as they start with an underscore.

[...]

PR #337, opened 2025-02-24
Semi-automatically populate `__all__` in `__init__.py` files

vpratz:
This PR implements one possible solution to the problem highlighted in #290. By specifying `__all__` in the relevant files, it makes explicit which modules and functions should be documented, so that only desired functions and modules are included. This also allows setting `autosummary_imported_members = False`, which greatly speeds up building the documentation.

It requires manual action when:

- a new public submodule is added: has to be added to the `include_submodules` list in the parent module
- a new `__init__.py` is added: must add the following boilerplate at the end of the file (number of dots before utils depends on the path):

```
from ..utils._docs import _add_imports_to_all

_add_imports_to_all(include_modules=[])
```

Please give feedback on the underlying mechanism. Is this an option we would be happy with, or should we rather try something else?

### DIFF
--- a/bayesflow/adapters/__init__.py
+++ b/bayesflow/adapters/__init__.py
@@ -1,2 +1,6 @@
 from . import transforms
 from .adapter import Adapter
+
+from ..utils._docs import _add_imports_to_all
+
+_add_imports_to_all(include_modules=["transforms"])
--- a/bayesflow/adapters/transforms/__init__.py
+++ b/bayesflow/adapters/transforms/__init__.py
@@ -16,3 +16,7 @@ from .rename import Rename
 from .standardize import Standardize
 from .to_array import ToArray
 from .transform import Transform
+
+from ...utils._docs import _add_imports_to_all
+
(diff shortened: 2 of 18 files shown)


vpratz:
@LarsKue Do you have any concerns against merging this, at least as a temporary solution? As there is now an effort to complete the docs (#340), I think the quick build times enabled by this change would be nice, as they enable fast iteration and checking of the docs. Also, while the focus is on this, we are able to detect if we missed anything that should be in `__all__`, should we want to keep this implementation as a permanent solution.

paul-buerkner:
I cannot speak for Lars but I would be in favor of merging this :-)

Valentin Pratz ***@***.***> schrieb am Fr., 28. Feb. 2025,
18:31:

LarsKue:
As a temporary solution, I think it is fine for now. We should not fall into the trap of having to manually maintain `__all__` though.

paul-buerkner:
Great! Valentin, feel free to merge :-)

PR #338, opened 2025-02-25
Allow inclusive bounds in `Adapter.constrain()`

LarsKue:
Simply adds a small value to bounds that are specified as inclusive.

### DIFF
--- a/bayesflow/adapters/adapter.py
+++ b/bayesflow/adapters/adapter.py
@@ -233,8 +233,8 @@ class Adapter(MutableSequence[Transform]):
         lower: int | float | np.ndarray = None,
         upper: int | float | np.ndarray = None,
         method: str = "default",
-        inclusive: str = "default",
-        epsilon: float = 1e-16,
+        inclusive: str = "both",
+        epsilon: float = 1e-15,
     ):
         if isinstance(keys, str):
             keys = [keys]
--- a/bayesflow/adapters/adapter.py
+++ b/bayesflow/adapters/adapter.py
@@ -233,12 +233,17 @@ class Adapter(MutableSequence[Transform]):
         lower: int | float | np.ndarray = None,
         upper: int | float | np.ndarray = None,
         method: str = "default",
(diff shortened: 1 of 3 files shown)


paul-buerkner:
thank you! looks good. Question: Why would you make the lower bound inclusive but the upper exclusive by default?

LarsKue:
@paul-buerkner In my opinion, the most common case for double-bounded variables is uniform distribution. This is typically in the half-open interval `[lower, upper)`.

paul-buerkner (changes requested):
Looks good except for my suggestion for symmetric default values that I added as direct comment to the issue.

paul-buerkner:
@LarsKue do i have permission to edit this PR in the direction I mentioned aboved?

LarsKue:
@paul-buerkner yes. I do feel like it could be a bit opaque to new users though, since most randomly generated numbers are in open or half-open intervals, regardless of the statistical interpretation of the bounds.

paul-buerkner:
Is there even a difference for continuous distributions? I mean, the probability of sampling exactly from the boundary is zero. So I am confused by your statement that intervals would ever be half open in software. For discrete variables perhaps but those are not our main concern in the adapter transformations.

paul-buerkner:
Just to add one thought. Our point here is not to determine whether a null-set is included in an interval or not, but to ensure numerical stability. And for the latter, we need the epsilon, i.e., both inclusive bounds. Would you agree?

paul-buerkner (approved):
From my side, the PR is ready now. I change the default inclusive to "both" and added some tests. During testing, I also noticed that epsilon 1e-16 is too small to ensure finite outputs if the input is exactly at the upper boundary. So I changed it to 1e-15 which seems to work well. 

Can you, @LarsKue take another quick look and approve as well? I will merge then as soon as the checks are all passing.

paul-buerkner:
I further extended the tests to now hopefully cover all relevant cases. Will merge now.

PR #346, opened 2025-03-05
Basic workflow: change default for save best only to False

vpratz:
By default, use the latest state instead of the best, as the losses can be unreliable for selecting the best model.


paul-buerkner:
Looks good. You added some other code not related to the change in the title. Was this intentional?

vpratz:
Thanks for taking a look. Yes, the changes were intentional and related to checkpoint saving as well (warning if existing checkpoints will be overridden), I forgot to mention them in the text.

vpratz:
But looking at it again was good, I found a typo (`checkpoint_filepath` instead of `checkpoint_full_filepath`)...

PR #348, opened 2025-03-10
adapter: support log and sqrt transforms

paul-buerkner:
This PR adds support for log and sqrt transforms as requested by @stefanradev93.

@LarsKue I noticed that we are using two different patterns to deal with `keys` in the adapter. One that uses keys directly in the internal adapter transforms (e.g., expand_dims) and another that loops over the keys inside the main `Adapter` (e.g., `as_set`). Is this international? If so, when shall we apply which pattern?

### DIFF
--- a/bayesflow/adapters/adapter.py
+++ b/bayesflow/adapters/adapter.py
@@ -483,7 +483,7 @@ class Adapter(MutableSequence[Transform]):
         if isinstance(keys, str):
             keys = [keys]
 
-        transform = ExpandDims(keys, axis=axis)
+        transform = MapTransform({key: ExpandDims(axis=axis) for key in keys})
         self.transforms.append(transform)
         return self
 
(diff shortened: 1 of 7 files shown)


bayesflow/adapters/adapter.py
```diff
@@ -500,6 +502,23 @@ def keep(self, keys: str | Sequence[str]):
         self.transforms.append(transform)
         return self
 
+    def log(self, keys: str | Sequence[str], *, p1: bool = False):
+        """Append an :py:class:`~transforms.Log` transform to the adapter.
+
+        Parameters
+        ----------
+        keys : str or Sequence of str
+            The names of the variables to expand.
```
vpratz, line 511: typo: expand -> transform (below as well)

paul-buerkner:
Thanks! That makes sense!

paul-buerkner:
Done. Can you check again @LarsKue ?

LarsKue (approved):
Thank you for the changes. I think this looks good :+1:

paul-buerkner:
Great! Merging now.

PR #350, opened 2025-03-11
Numpy transforms

LarsKue:
This PR
- removes the old `LambdaTransform`
- introduces a new `NumpyTransform`

This allows trivial serialization for the most common applications of the `LambdaTransform`.

PR #352, opened 2025-03-12
Upgrade to Keras 3.9

LarsKue:
We need this upgrade to get rid of a security vulnerability in Keras < 3.9:

https://nvd.nist.gov/vuln/detail/CVE-2025-1550
https://github.com/advisories/GHSA-48g7-3x6r-xfhp


LarsKue:
Looks like 3.9.0 is not available yet on conda-forge. We should probably update our CI install pipeline to rely on pip rather than conda.

LarsKue:
There are some errors still that are subject to be fixed by #350.

PR #357, opened 2025-03-17
SSM wrapper

Chase-Grajeda:
Implements a wrapper to run a Mamba SummaryNetwork with BayesFlow

- Current implementation is locked to Torch backends only given fundamental constraints.
- Nvidia GPUs are highly recommended; AMD cards have extra setup required for compatibility. 
- CPU is not supported. 

https://github.com/state-spaces/mamba


stefanradev93:
Great, Chase! Can you merge dev into your branch and make sure the tests are running. Also, no need to change the SIR notebook (revert to default), since we will take one of your upcoming notebooks as an example.

stefanradev93:
Implemented as we did in the project.

PR #364, opened 2025-03-21
Fix `OfflineDataset` for non-array Simulator output

### LINKED ISSUE #363: Offline dataset not possible with a meta function
Hi!

When using `bayesflow.OfflineDataset` with a simulator that includes a `meta_fn`, an error occurs because the dataset initialization assumes all values in the dictionary have a `.shape` attribute.

Code:
```python
import os
os.environ["KERAS_BACKEND"] = "jax"
os.environ["JAX_PLATFORMS"] = "cpu"
import bayesflow as bf
import numpy as np

ad = bf.Adapter()
ad.to_array()

def prior(N):
    return {"w": np.random.normal(1, 0.2, N)}

def meta(batch_size):
    return {"N": np.random.randint(5, 15)}

simulator = bf.make_simulator([prior], meta_fn=meta)


data = simulator.sample(8)
dataset = bf.OfflineDataset(data, batch_size=2, adapter=ad)
dataset[0]
```
Error Message:
```python
Traceback (most recent call last):
  File "/data/homes/reiser/projects/bayesflow/offline_dataset.py", line 20, in <module>
    dataset = bf.OfflineDataset(data, batch_size=2, adapter=ad)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/homes/reiser/projects/bayesflow/bayesflow/datasets/offline_dataset.py", line 19, in __init__
    self.num_samples = next(iter(data.values())).shape[0]
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'shape'
```

LarsKue:
Fixes #363

### DIFF
--- a/bayesflow/datasets/offline_dataset.py
+++ b/bayesflow/datasets/offline_dataset.py
@@ -2,6 +2,7 @@ import keras
 import numpy as np
 
 from bayesflow.adapters import Adapter
+from bayesflow.utils import logging
 
 
 class OfflineDataset(keras.utils.PyDataset):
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks for the fix! Looks good I think.

bayesflow/datasets/offline_dataset.py
```diff
@@ -11,12 +12,20 @@ class OfflineDataset(keras.utils.PyDataset):
...
+    def __init__(
+        self, data: dict[str, np.ndarray], batch_size: int, adapter: Adapter | None, num_samples: int = None, **kwargs
+    ):
         super().__init__(**kwargs)
         self.batch_size = batch_size
         self.data = data
         self.adapter = adapter
-        self.num_samples = next(iter(data.values())).shape[0]
+
+        if num_samples is None:
```
paul-buerkner, line 23: I do like that the user has both options (a) to automatically compute num_samples (default) or (b) to manually specify it, if for some reason, num_samples cannot be reliably inferred by the data.

PR #369, opened 2025-03-24
Example Notebook explaining changes from 1.0 to 2.0 

eodole:
This notebook aims to explain relevant differences for current users of 1.0 as they switch to 2.0 
There are a few sections that I feel need clarification: 
1. In the adapter and data configuration section there's a sentence with ??? where I didn't know what to write
2. Are there any new features that I am missing?


paul-buerkner:
Merging now since only the notebook was added so checks will be unaffected.

PR #378, opened 2025-03-27
Update docs

LarsKue:


vpratz (approved):
Looks good to me. I have fixed some typos in the alt text I had written, as well as the links to the example notebook. For historical reasons, they require an underscore before examples, but I think this is something we might want to change.

Issue #304, opened 2025-02-11
Correct transformation of second-order tensors in adapters


han-ol:
Currently, only first-order tensors, a.k.a. vectors can be transformed in both forward and inverse direction through an adapter object.

Can we transform also different orders, particularly second-order?

The question arises because a `PointInferenceNetwork` can estimate the covariance matrix of the `inference_variables`.
Some adapters represent a change of basis and origin, for example, `standardize` is a linear coordinate transformation from "original coordinates" to "standardized coordinates".
Thus, the inference_variables live in the standardized coordinates, and the covariance matrix needs to be adapted in inverse direction to relate to unstandardized coordinates.

When a covariance matrix is estimated in the standardized coordinates, the inverse of the coordinates transformation is different from naively transforming matrix columns as if they were vectors in standardized coordinates.
Rather, a covariance matrix $\Sigma^{-1}$ transforms as an order 2 tensor: the basis change matrix $A$ is multiplied from both sides $A^T \Sigma^{-1} A$. 
For `standardize`, $A=diag(\sigma_i)$ with $\sigma_i$ the standard deviations for each dimension, so $A^T \Sigma^{-1} A = \Sigma^{-1} \cdot diag(\sigma_i^2)$.

Solutions to this issue are likely related to a separate issue on keeping track of jacobians of adapter transforms, as mentioned in https://github.com/bayesflow-org/bayesflow/issues/245.

paul-buerkner:
For non-linear transformations, I see little hope that we can make it work generally for point estimates. There are some combinations of point estimates and transformations that work such as quantiles and strictly monotonic transforms. But I think that, for now, we should warn somewhere when a point estimate is passed through a non-linear transformation and recommend to just not use these transformation with pointInferenceNetworks.

Standarize is a bit of a different situation, both because it is linear and because it is so important for the networks to work in a stable manner in lots of situtations. So mean and quantiles can just be back transformed through standardize and all is well. 

Now, for a covariance matrix the situation is different as you highlight above. Same actually for only the variance since we need to square the scaling. 

I suggest something like the following solution to this general problem (covariance matrix is even harder than this):

Every point estimation loss implemented natively by us (e.g. squared loss) knows which point estimate it produces (e.g., mean). So we can enrich it by a structure telling us what to do with inverese transforms. This is turn could be a dict of 3 options.

- Names of transforms that are valid for this point estimate, such that the point estimate can be safely transformed by the default approach (e.g., standardize for mean)
[...]

PR #380, opened 2025-03-27
Stabilizing multivariate normal approximation

han-ol:
This PR seeks to adress the stability of estimating mean and covariance using
```python
inference_network=bf.networks.PointInferenceNetwork(
        scores=dict(
            mvn=bf.scores.MultivariateNormalScore()
        )
)
# -> results in warning: MultivariateNormalScore is unstable.
```
It is fixed by using link function called PositiveDefinite, that is based on the Cholesky decomposition.


paul-buerkner:
Thanks! Is it intentional that this is to be merged into the `build-deepset-without-call` branch instead of into main (or dev)?

han-ol:
Definitely not intentional. This was actually developed against the current dev branch, so I changed it accordingly.

The new `links.PositiveDefinite` relies on `fill_triangular_matrix`. Which fails for tensorflow backend with `pred must not be a Python bool`, without it being clear where such a pred variable is defined

han-ol:
Pair programming with Valentin was successful (thx!) and we smoothed out the flaky tests. All backends support estimation of multivariate normal parameters now.

I removed some rough edges, added tests, so hopefully the codecov/path tests pass too now.
 
The PR is ready for review as soon as tests passed.

bayesflow/scores/scoring_rule.py
```diff
@@ -29,6 +29,8 @@ def __init__(
         self.subnets_kwargs = subnets_kwargs or {}
         self.links = links or {}
 
+        self.not_transforming_like_vector = []
```
vpratz, line 32: Please add a comment documenting what this should contain, so people who want to set this in a subclass know what to use

bayesflow/links/positive_definite.py
```diff
@@ -0,0 +1,47 @@
+import keras
+
+# import numpy as np
```
LarsKue, line 3: Please remove this

bayesflow/networks/point_inference_network.py
```diff
@@ -132,7 +132,7 @@ def call(
         if xz is None and not self.built:
             raise ValueError("Cannot build inference network without inference variables.")
         if conditions is None:  # unconditional estimation uses a fixed input vector
-            conditions = keras.ops.convert_to_tensor([[1.0]], dtype=keras.ops.dtype(xz))
```
LarsKue, line 135: Under what circumstances does this break? It seems like a good choice to use the dtype

bayesflow/approximators/point_approximator.py
```diff
@@ -119,6 +119,7 @@ def sample(
     def _prepare_conditions(self, conditions: dict[str, np.ndarray], **kwargs) -> dict[str, Tensor]:
         """Adapts and converts the conditions to tensors."""
         conditions = self.adapter(conditions, strict=False, stage="inference", **kwargs)
+        conditions.pop("inference_variables", None)
```
LarsKue, line 122: We could add this function to the ContinuousApproximator, if it is identical between it and the Point Approximator

bayesflow/approximators/point_approximator.py
```diff
@@ -130,6 +131,12 @@ def _apply_inverse_adapter_to_estimates(
         for score_key, score_val in estimates.items():
             processed[score_key] = {}
             for head_key, estimate in score_val.items():
+                if head_key in self.inference_network.scores[score_key].not_transforming_like_vector:
```
LarsKue, line 134: This seems like a very specific attribute. Can you explain some more about it?

bayesflow/scores/multivariate_normal_score.py
```diff
@@ -97,6 +98,8 @@ def sample(self, batch_shape: Shape, mean: Tensor, covariance: Tensor) -> Tensor
         Tensor
             A tensor of shape (batch_size, num_samples, D) containing the generated samples.
         """
+        if len(batch_shape) == 1:
+            batch_shape = (1,) + batch_shape
```
LarsKue, line 102: This fails if the user passes a list type batch shape

bayesflow/scores/scoring_rule.py
LarsKue: +1, I would also like to know if we can handle this in a better way than a special attribute.

bayesflow/scores/scoring_rule.py
```diff
@@ -95,14 +97,14 @@ def get_link(self, key: str) -> keras.Layer:
...
+    def get_head(self, key: str, output_shape: Shape) -> keras.Sequential:
         """For a specified head key and shape, request corresponding head network.
 
         Parameters
         ----------
         key : str
             Name of head for which to request a link.
-        shape: Shape
+        output_shape: Shape
             The necessary shape for the point estimators.
```
LarsKue, line 108: Could improve this doctring now that the name is changed

tests/test_links/test_links.py
```diff
@@ -59,16 +52,21 @@ def test_quantile_ordering(quantiles, unordered):
...
+    input_shape = positive_definite.compute_input_shape((batch_size, num_variables, num_variables))
 
-    activation = PositiveSemiDefinite()
+    # Too strongly negative values lead to numerical instabilities -> reduce scale
+    random_preactivation = keras.random.normal(input_shape) * 0.1
+    output = positive_definite(random_preactivation)
 
-    output = activation(random_matrix_batch)
+    # Check if output is invertible
+    np.linalg.inv(output)
```
LarsKue, line 63: Don't use numpy for Tensors. The workflow won't show it, but this fails on Torch CUDA environments.

Either use numpy everywhere, or use tensors and then convert to numpy before calling it, or use `keras.ops.inv`, whichever is most appropriate (depends on the input requirements of `positive_definite` here)

LarsKue (approved):
Code overall looks good. See individual comments. I don't know enough about the requirements of this task to give a definite approval, but I trust that @han-ol did their research on this 🙂

bayesflow/scores/scoring_rule.py
vpratz: Do we want this to be variable within a class, i.e., does this have to be an instance variable? If not, a constant [class variable](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables) might be an option, like
``` python
class ScoringRule:
    #: This variable lists ... (the #: syntax should allow sphinx to parse this)
    NOT_TRANSFORMING_LIKE_VECTOR = tuple()  # use immutable type tuple instead of list
```

bayesflow/networks/point_inference_network.py
han-ol: Previously it broke because `xz` can be None. The method `estimate()` passes `xz=None` to `call()`.

(A non-None `xz` is passed to `call()` during training and is necessary for building the heads correctly.)

Solution: We can fix the dtype if `xz` is available and use float3 as a fallback if `xz` is None:

```python
conditions = keras.ops.convert_to_tensor(
    [[1.0]], dtype=keras.ops.dtype(xz) if xz is not None else "float32"
)
```

bayesflow/scores/multivariate_normal_score.py
han-ol: Good catch! Should we just use a tuple conversion?
```python
batch_shape = (1,) + tuple(batch_shape)
```

LarsKue: I think the tuple conversion is ok. Internally, we only use tuple shapes, so I would have to think some more about changing the type hint itself.

But of course we want users to be able to pass lists where possible (and not a significant burden).

bayesflow/scores/scoring_rule.py
han-ol: Good idea to use a class variable for this! Implemented it in [d87b0b9](https://github.com/bayesflow-org/bayesflow/pull/380/commits/d87b0b911b6dbab8790dd390ce237738c48c6b1f). 

I added documenting comments as well.

han-ol: Regarding a better way of handling this in general:

There is a discussion in #304 about the long term plans for special estimators and how they play with adapter transformations.
In my opinion, what we have here suffices for now and is a reasonable first step.

bayesflow/approximators/point_approximator.py
han-ol: I am marking this conversation as resolved in favor of the one where we have a thread about the attribute.

bayesflow/approximators/point_approximator.py
han-ol: Yes! This and similar refactoring of the ContinuousApproximator is a good idea (but I would keep them out of this PR).
There is also the option of moving the conversion to tensor into the adapter. Possibly with an optional bool flag convert_to_tensor that is by default False.

PR #381, opened 2025-03-28
Populate API Reference landing page

LarsKue:
Some docstrings don't yet render correctly:

- datasets is missing the 2nd half of the text
- experimental is missing the "Proceed with caution."
- simulators is missing the 2nd half of the text

We could also try to improve the links to `keras` classes with [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html).

### DIFF
--- a/bayesflow/adapters/__init__.py
+++ b/bayesflow/adapters/__init__.py
@@ -1,5 +1,6 @@
 """
-A collection of :py:class:`~bayesflow.adapters.Adapter` transforms, which tell BayesFlow how to interpret your :py:class:`~bayesflow.simulators.Simulator` output and plug it into neural networks for training and inference.
+A collection of :py:class:`~bayesflow.adapters.Adapter` transforms, which tell BayesFlow how to interpret your
+:py:class:`~bayesflow.simulators.Simulator` output and plug it into neural networks for training and inference.
 """
 
 from . import transforms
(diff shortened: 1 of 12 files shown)


vpratz (approved):
Looks good to me, thanks for writing them :)

LarsKue:
@vpratz Do you know why some of the text does not render correctly? I have tried searching for a solution to this but have not come up with anything.

vpratz:
Sorry, I missed the original post, I'll take a look.
Regarding intersphinx for Keras: Unfortunately, as Keras is not using sphinx, intersphinx is not an option here as far as I could tell. If you come across a way to use intersphinx (or something similar) here, please let me know, it would certainly be a nice addition.

vpratz:
From my first impression, it does not seem configurable. I think our main options are:
- Rephrase/shorten, so that we only have one sentence that goes into the table. The full description will show up on the modules page.
- Using UTF-8 fake dots, which are not matched by the regular expression, I think [one dot leader](https://www.compart.com/en/unicode/U+2024) would be the character of choice here. This is confusing for anyone trying to modify those docs in the future though.
- Replacing the table with something manually written in the docs. This gives us more freedom, but is prone to duplicates/incongruent information of code vs. docs.

What do you think is best here? Or do you have other ideas?

LarsKue:
@vpratz Thank you for the research. I'll improve the first sentence of the docs.

vpratz:
Looks good to me, thanks for the changes :)

PR #382, opened 2025-04-01
Docs: minor fix, :py:function: must be :py:func:

vpratz:
Only a small fix, I missed this in the review of #381.


LarsKue (approved):
Thank you for the fix!

PR #385, opened 2025-04-01
Docs: more explicit make command names

vpratz:
There was some confusion arising from the fact that `make docs` did not build the local docs. To avoid this confusion in the future, the naming is made more explicit with the following changes:

make local -> make local-docs
make docs -> make production-docs
make docs-sequential -> make production-docs-sequential

The documentation and the GitHub actions are adapted as well.

This PR also includes a minor fix in the `make.bat`, where the jump label for `local` was wrong.

### DIFF
--- a/.github/workflows/multiversion-docs.yaml
+++ b/.github/workflows/multiversion-docs.yaml
@@ -43,7 +43,7 @@ jobs:
         run: |
           cd ./repo/docsrc
           make clean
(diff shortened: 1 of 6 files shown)


han-ol (approved):
Thanks! Renaming this is a good idea. I don't think you missed an occurrence. Should be good to go.

PR #389, opened 2025-04-04
Add c2st

stefanradev93:
Adds classifier two-sample test (C2ST) metric to the `bayesflow.diagnostics.metrics` module. Example usage:

`
samples = ...
reference_samples = ...

# Using default arguments
c2st = classifier_two_sample_test(samples, reference_samples)

# Customizing arguments - will return full classification report + classifier and use a smaller validation data ratio
c2st_dict = classifier_two_sample_test(samples, reference_samples, return_metric_only=False, validation_split=0.2)
`
What we need to discuss is the conceptual separation of metrics, namely those intended to capture the performance of an ensemble of posteriors (e.g., SBC, RMSE) and depend on the prior or those intended to capture the quality of individual posteriors (e.g., MMD, C2ST). Should these two classes of metrics reside in different submodules? @paul-buerkner @vpratz @LarsKue 

Note also, that I removed the internal cross-validation in favor of a simple train-test split. Since the C2ST will generally be computed on different source-target posterior samples, any epistemic uncertainty will average out. However, I could allow K-fold as well with some custom helpers, mainly because I wanted to avoid the explicit dependence on `sklearn`.

### DIFF
--- a/bayesflow/diagnostics/metrics/classifier_two_sample_test.py
+++ b/bayesflow/diagnostics/metrics/classifier_two_sample_test.py
@@ -1,0 +1,121 @@
+from typing import Sequence, Mapping, Any
+
+import numpy as np
+
+import keras
+
+from bayesflow.utils.exceptions import ShapeError
+from bayesflow.networks import MLP
+
+
+def classifier_two_sample_test(
+    estimates: np.ndarray,
+    targets: np.ndarray,
+    metric: str = "accuracy",
+    patience: int = 10,
+    max_epochs: int = 1000,
+    batch_size: int = 64,
+    return_metric_only: bool = True,
+    validation_split: float = 0.5,
+    standardize: bool = True,
+    mlp_widths: Sequence = (256, 256),
+    **kwargs,
+) -> float | Mapping[str, Any]:
+    """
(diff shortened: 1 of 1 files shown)


vpratz:
Thanks for the PR :)

> Should these two classes of metrics reside in different submodules?

I like the idea, but lack good names that we could use for the respective submodules. Do you have an idea on how to distinguish them with two words that could serve as submodule names?

> Note also, that I removed the internal cross-validation in favor of a simple train-test split.

As far as I can tell, using cross-validation gives us a more powerful test for the same number of samples, so I would opt against removing this in the default setting. How are other implementations handling this? Is there some default setup we might want to implement?
We use `sklearn` in `expected_calibration_error` and `mc_confusion_matrix` as well. Is the plan to refactor those as well, or are we going to keep the dependency anyway?

paul-buerkner:
I think the PR looks good. Would you have time to add some tests too perhaps?

I would keep them in the same module for now even if their inputs are different. We clearly document the difference in input and I am not sure how we would name them if we went for two modules.

stefanradev93:
@vpratz I removed all dependencies on `sklearn` across the library. @paul-buerkner  Added tests for sensitivity of c2st.

LarsKue (approved):

PR #399, opened 2025-04-10
Add class for custom transforms to adapter, similar to `LambdaTransform`.

### LINKED ISSUE #362: Simplify use of arbitrary custom transforms in the `Adapter`
The removal of the `LambdaTransform` leaves a hole that needs to be filled, as arbitrary custom transforms are something that we require in our projects. Implementing custom transforms via subclassing is possible, but requires a lot more boilerplate, and is harder to get right. If one wants to combine it with the `FilterTransform` as was possible with `.apply`, the complexity increases even more.

I personally think what `LambdaTransform` and `.apply` offered was a very convenient wrapper, that let's users focus on the main part: Writing the forward and inverse transform.  in the constructor of `LambdaTransform`, we could check if the provided functions are already registered and therefore serializable. If they are not, we can warn/error with instructions of how to achieve that (which is: put a decorator before the function). This could be accompanied by a more detailed tutorial-style notebook, that can also be linked in the warning. In my opinion, this should be sufficient to get the power users that want to use such transforms on the right track, and be able to alleviate the concerns regarding serialization.

Please let me know what you think.
@LarsKue @stefanradev93 @paul-buerkner @elseml

vpratz:
Closes #362.

This PR reintroduces the features that were present in `LambdaTransform`, but only allowing registered functions. While being stricter, this allows for closer scaffolding and raising errors early on, so that users cannot provide functions that will not be (de)serializable later on.

As there are a few failure modes, the focus is on providing detailed error messages to enable users to solve problems without external help.

Open questions:

- When writing the code, I chose the naming to make it very clear that you cannot pass any functions, but only serializable ones. In my impression, I overdid it a bit, and the point should be sufficiently clear, even when we rename `serializable_forward/inverse_fn` to `forward/inverse` again. What do you think?
- I opted for a method in the `Adapter`, as it helps in easily providing the user with the `FilterTransform` functionality. Is this ok with you, @LarsKue 
- As always, naming suggestions are welcome.


tests/test_adapters/test_adapters.py
```diff
@@ -110,3 +111,71 @@ def test_simple_transforms(random_data):
...
+    def registered_but_changed(x):
+        return x
+
+    def registered_but_changed(x):  # noqa: F811
+        return 2 * x
+
+    # method instead of function provided
+    with pytest.raises(ValueError):
+        SerializableCustomTransform(serializable_forward_fn=A.fn, serializable_inverse_fn=registered_fn)
+        SerializableCustomTransform(serializable_forward_fn=registered_fn, serializable_inverse_fn=A.fn)
```
LarsKue, line 144: These need to be in separate `with` blocks. The block exits on the first `ValueError` that is raised.

bayesflow/adapters/adapter.py
```diff
@@ -283,6 +284,88 @@ def apply(
         self.transforms.append(transform)
         return self
 
+    def apply_serializable(
+        self,
+        include: str | Sequence[str] = None,
+        *,
+        serializable_forward_fn: Callable[[np.ndarray, ...], np.ndarray],
+        serializable_inverse_fn: Callable[[np.ndarray, ...], np.ndarray],
```
LarsKue, line 292: I agree that these are a bit verbose. The intent is clear from the name of the function and the docs, I think. Let's go with your suggestion to rename them back to `forward` and `inverse`.

bayesflow/adapters/transforms/serializable_custom_transform.py
```diff
@@ -0,0 +1,184 @@
...
+        GENERAL_EXAMPLE_CODE = f"""The example code below shows the structure of a correctly decorated function:
+
+```
+import keras
+
+@keras.saving.register_keras_serializable('custom')
+def my_{label}(...):
+    [your code goes here...]
+```
+"""
```
LarsKue, line 71: Triple-quoted strings lead to some super weird indenting here. Can we move to using single-quoted strings with `\n`? Not high priority, though.

LarsKue (changes requested):
Very awesome changes, thank you for covering this so thoroughly!

I do see the need for some minor changes, particularly in the tests. See individual comments.

tests/test_adapters/test_adapters.py
vpratz: Good spot, thanks!

vpratz:
Thanks a lot for the review, I have addressed your comments. The tests are currently failing because I pulled in the changes from `dev`. If you have no objections, I will merge when the tests on `dev` are running again.

LarsKue:
@vpratz Yes, dev is now fixed. Can you please merge dev into this again and if tests pass, merge?

PR #401, opened 2025-04-10
Proof-read and clarify v1 to v2 transition notebook.

vpratz:
In my opinion, the notebook would benefit from more details, so feel free to add things @LarsKue @stefanradev93.


vpratz:
Thanks a lot for the proof-reading. I have added a short comment on the `configurator`, to make the connection clearer, but feel free to extend.

PR #404, opened 2025-04-11
Enable integration with fixed, non-uniform schedule.

### LINKED ISSUE #402: Integrators: Allow passing an integration schedule with fixed time-points
Some methods (e.g., diffusion models) [1] provide a fixed schedule for integration that takes into account the (approximately known) curvature of the underlying ODE.

Currently, only fixed single step size and (automatic) adaptive step size schedules are possible, but no fixed schedules with varying steps sizes.

@LarsKue How can we introduce them best?
One option woud be the following: If the `steps` parameter to `integrate` is `iterable` and numeric, we interpret it as times for a fixed schedule with `[start_time, t_1, t_2, ..., stop_time]`. In this setting, we would make the `start_time` and `stop_time` parameters optional, and introduce checks to ensure they are passed in the other cases.

What do you think? If you are ok with this design, I would offer to create a PR for this.

[1] https://arxiv.org/abs/2206.00364

vpratz:
This PR enables specialized, manually specified schedules, as for example required in diffusion models for inference.

Regarding the changes in `integrate`: I changed the `match-case` to an `if-else` structure, as I was not able to express it properly otherwise, as custom types like `Tensor` seem not to be easily compatible with match statements. If you want to keep the match statement and know how to do it, feel free to change it.

I'm also not sure whether we want to allow `Sequence` in general (lists, tuples, ...), or only numpy arrays and tensors.
If you have any thoughts, please let me know.

Closes #402.


LarsKue:
Looks great, thank you for the PR. @vpratz can you please re-merge dev into this and see if tests pass? Ready to merge from my side otherwise.

PR #407, opened 2025-04-12
Fixed weighted mean everywhere

stefanradev93:


vpratz (approved):
Looks good to me, thanks for the quick fix.

Issue #403, opened 2025-04-11
Add diffusion model as inference network


vpratz:
As we have multiple people working on this, I want to use this issue to coordinate a bit.

@arrjon @niels-leif-bracher What is the state of your projects/implementations, and what are your plans regarding the integration of diffusion models as inference networks into BayesFlow?

I have an implementation of the EDM/k-diffusion from [1], similar in style to the `FlowMatching` class.
I assume that our implementations are probably quite similar, and I think it would be nice to get one into the `experimental` module, so we can test it out and compare it to flow matching.

@stefanradev93 I think you are the one with the best overview, so if you have thoughts, please also share them here :)

[1] Elucidating the Design Space of Diffusion-Based Generative Models: https://arxiv.org/abs/2206.00364

stefanradev93:
@arrjon @niels-leif-bracher have implementations of various score-based models according to the classification x/epsilon/u prediction (https://diffusionflow.github.io/), also similar to `FlowMatching`. 

I believe we can sync the implementations quite well. You can PR your code when ready and add @arrjon / @niels-leif-bracher as reviewers to start the discussion going.

arrjon:
So the best solution might be to start with your code @vpratz and I add my stuff from there as Stefan suggested?

vpratz:
Closed by #408

PR #408, opened 2025-04-13
Add diffusion model implementation

vpratz:
This PR adds a diffusion model implementation for use as an inference network, as discussed in #403. It implements the design introduced as "EDM" in [1]. The overall structure is taken from the `FlowMatching` class.

@arrjon @niels-leif-bracher I would appreciate if you take a look and make suggestions regarding how we can incorporate the other diffusion model variants as well. For now, I chose to only expose the `sigma_data` parameter to the end user, and keep everything else private. This should enable us to also change the internals later on and incrementally add new functionality.

Please let me know how we want to proceed and how much capacity you have to move this forward, so that we can decide whether we want to include the additional options before we merge, or if we merge early and then incrementally add to it later. I have situated the class in the `experimental` module for now, so that we have some freedom to also change things in the future as we see fit.

[1] https://arxiv.org/abs/2206.00364

### DIFF
--- a/bayesflow/experimental/diffusion_model.py
+++ b/bayesflow/experimental/diffusion_model.py
@@ -1,0 +1,346 @@
+from collections.abc import Sequence
+import keras
+from keras import ops
+from keras.saving import register_keras_serializable as serializable
+
+from bayesflow.types import Tensor, Shape
+import bayesflow as bf
+from bayesflow.networks import InferenceNetwork
+
+from bayesflow.utils import (
+    expand_right_as,
+    find_network,
+    jacobian_trace,
+    keras_kwargs,
+    serialize_value_or_type,
+    deserialize_value_or_type,
+    weighted_mean,
+    integrate,
+)
+
+
+@serializable(package="bayesflow.networks")
+class DiffusionModel(InferenceNetwork):
+    """Diffusion Model as described as Elucidated Diffusion Model in [1].
+
+    [1] Elucidating the Design Space of Diffusion-Based Generative Models: arXiv:2206.00364
+    """
+
+    MLP_DEFAULT_CONFIG = {
+        "widths": (256, 256, 256, 256, 256),
+        "activation": "mish",
+        "kernel_initializer": "he_normal",
+        "residual": True,
+        "dropout": 0.0,
+        "spectral_normalization": False,
+    }
+
+    INTEGRATE_DEFAULT_CONFIG = {
+        "method": "euler",
+        "steps": 100,
+    }
+
+    def __init__(
+        self,
+        subnet: str | type = "mlp",
+        integrate_kwargs: dict[str, any] = None,
+        subnet_kwargs: dict[str, any] = None,
+        sigma_data=1.0,
+        **kwargs,
+    ):
+        """
+        Initializes a diffusion model with configurable subnet architecture.
+
+        This model learns a transformation from a Gaussian latent distribution to a target distribution using a
+        specified subnet type, which can be an MLP or a custom network.
+
+        The integration steps can be customized with additional parameters available in the respective
+        configuration dictionary.
+
+        Parameters
+        ----------
+        subnet : str or type, optional
+            The architecture used for the transformation network. Can be "mlp" or a custom
+            callable network. Default is "mlp".
+        integrate_kwargs : dict[str, any], optional
+            Additional keyword arguments for the integration process. Default is None.
+        subnet_kwargs : dict[str, any], optional
+            Keyword arguments passed to the subnet constructor or used to update the default MLP settings.
+        sigma_data : float, optional
+            Averaged standard deviation of the target distribution. Default is 1.0.
+        **kwargs
+            Additional keyword arguments passed to the subnet and other components.
+        """
+
+        super().__init__(base_distribution=None, **keras_kwargs(kwargs))
+
+        # internal tunable parameters not intended to be modified by the average user
+        self.max_sigma = kwargs.get("max_sigma", 80.0)
+        self.min_sigma = kwargs.get("min_sigma", 1e-4)
+        self.rho = kwargs.get("rho", 7)
+        # hyper-parameters for sampling the noise level
+        self.p_mean = kwargs.get("p_mean", -1.2)
+        self.p_std = kwargs.get("p_std", 1.2)
+
+        # latent distribution (not configurable)
+        self.base_distribution = bf.distributions.DiagonalNormal(mean=0.0, std=self.max_sigma)
+        self.integrate_kwargs = self.INTEGRATE_DEFAULT_CONFIG | (integrate_kwargs or {})
+
+        self.sigma_data = sigma_data
+
+        self.seed_generator = keras.random.SeedGenerator()
+
+        subnet_kwargs = subnet_kwargs or {}
+        if subnet == "mlp":
+            subnet_kwargs = self.MLP_DEFAULT_CONFIG | subnet_kwargs
+
+        self.subnet = find_network(subnet, **subnet_kwargs)
+        self.output_projector = keras.layers.Dense(units=None, bias_initializer="zeros")
+
+        # serialization: store all parameters necessary to call __init__
+        self.config = {
+            "integrate_kwargs": self.integrate_kwargs,
+            "subnet_kwargs": subnet_kwargs,
+            "sigma_data": sigma_data,
+            **kwargs,
+        }
+        self.config = serialize_value_or_type(self.config, "subnet", subnet)
+
+    def build(self, xz_shape: Shape, conditions_shape: Shape = None) -> None:
+        super().build(xz_shape, conditions_shape=conditions_shape)
+
+        self.output_projector.units = xz_shape[-1]
+        input_shape = list(xz_shape)
+
+        # construct time vector
+        input_shape[-1] += 1
+        if conditions_shape is not None:
+            input_shape[-1] += conditions_shape[-1]
+
+        input_shape = tuple(input_shape)
+
+        self.subnet.build(input_shape)
+        out_shape = self.subnet.compute_output_shape(input_shape)
+        self.output_projector.build(out_shape)
+
+    def get_config(self):
+        base_config = super().get_config()
+        return base_config | self.config
+
+    @classmethod
+    def from_config(cls, config):
+        config = deserialize_value_or_type(config, "subnet")
+        return cls(**config)
+
+    def _c_skip_fn(self, sigma):
+        return self.sigma_data**2 / (sigma**2 + self.sigma_data**2)
+
+    def _c_out_fn(self, sigma):
+        return sigma * self.sigma_data / ops.sqrt(self.sigma_data**2 + sigma**2)
+
+    def _c_in_fn(self, sigma):
+        return 1.0 / ops.sqrt(sigma**2 + self.sigma_data**2)
+
+    def _c_noise_fn(self, sigma):
+        return 0.25 * ops.log(sigma)
+
+    def _denoiser_fn(
+        self,
+        xz: Tensor,
+        sigma: Tensor,
+        conditions: Tensor = None,
+        training: bool = False,
+    ):
+        # calculate output of the network
+        c_in = self._c_in_fn(sigma)
+        c_noise = self._c_noise_fn(sigma)
+        xz_pre = c_in * xz
+        if conditions is None:
+            xtc = keras.ops.concatenate([xz_pre, c_noise], axis=-1)
+        else:
+            xtc = keras.ops.concatenate([xz_pre, c_noise, conditions], axis=-1)
+        out = self.output_projector(self.subnet(xtc, training=training), training=training)
+        return self._c_skip_fn(sigma) * xz + self._c_out_fn(sigma) * out
+
+    def velocity(
+        self,
+        xz: Tensor,
+        sigma: float | Tensor,
+        conditions: Tensor = None,
+        training: bool = False,
+    ) -> Tensor:
+        # transform sigma vector into correct shape
+        sigma = keras.ops.convert_to_tensor(sigma, dtype=keras.ops.dtype(xz))
+        sigma = expand_right_as(sigma, xz)
+        sigma = keras.ops.broadcast_to(sigma, keras.ops.shape(xz)[:-1] + (1,))
+
+        d = self._denoiser_fn(xz, sigma, conditions, training=training)
+        return (xz - d) / sigma
+
+    def _velocity_trace(
+        self,
+        xz: Tensor,
+        sigma: Tensor,
+        conditions: Tensor = None,
+        max_steps: int = None,
+        training: bool = False,
+    ) -> (Tensor, Tensor):
+        def f(x):
+            return self.velocity(x, sigma=sigma, conditions=conditions, training=training)
+
+        v, trace = jacobian_trace(f, xz, max_steps=max_steps, seed=self.seed_generator, return_output=True)
+
+        return v, keras.ops.expand_dims(trace, axis=-1)
+
+    def _forward(
+        self,
+        x: Tensor,
+        conditions: Tensor = None,
+        density: bool = False,
+        training: bool = False,
+        **kwargs,
+    ) -> Tensor | tuple[Tensor, Tensor]:
+        integrate_kwargs = self.integrate_kwargs | kwargs
+        if isinstance(integrate_kwargs["steps"], int):
+            # set schedule for specified number of steps
+            integrate_kwargs["steps"] = self._integration_schedule(integrate_kwargs["steps"], dtype=ops.dtype(x))
+        if density:
+
+            def deltas(time, xz):
+                v, trace = self._velocity_trace(xz, sigma=time, conditions=conditions, training=training)
+                return {"xz": v, "trace": trace}
+
+            state = {
+                "xz": x,
+                "trace": keras.ops.zeros(keras.ops.shape(x)[:-1] + (1,), dtype=keras.ops.dtype(x)),
+            }
+            state = integrate(
+                deltas,
+                state,
+                **integrate_kwargs,
+            )
+
+            z = state["xz"]
+            log_density = self.base_distribution.log_prob(z) + keras.ops.squeeze(state["trace"], axis=-1)
+
+            return z, log_density
+
+        def deltas(time, xz):
+            return {"xz": self.velocity(xz, sigma=time, conditions=conditions, training=training)}
+
+        state = {"xz": x}
+        state = integrate(
+            deltas,
+            state,
+            **integrate_kwargs,
+        )
+
+        z = state["xz"]
+
+        return z
+
+    def _inverse(
+        self,
+        z: Tensor,
+        conditions: Tensor = None,
+        density: bool = False,
+        training: bool = False,
+        **kwargs,
+    ) -> Tensor | tuple[Tensor, Tensor]:
+        integrate_kwargs = self.integrate_kwargs | kwargs
+        if isinstance(integrate_kwargs["steps"], int):
+            # set schedule for specified number of steps
+            integrate_kwargs["steps"] = self._integration_schedule(
+                integrate_kwargs["steps"], inverse=True, dtype=ops.dtype(z)
+            )
+        if density:
+
+            def deltas(time, xz):
+                v, trace = self._velocity_trace(xz, sigma=time, conditions=conditions, training=training)
+                return {"xz": v, "trace": trace}
+
+            state = {
+                "xz": z,
+                "trace": keras.ops.zeros(keras.ops.shape(z)[:-1] + (1,), dtype=keras.ops.dtype(z)),
+            }
+            state = integrate(deltas, state, **integrate_kwargs)
+
+            x = state["xz"]
+            log_density = self.base_distribution.log_prob(z) - keras.ops.squeeze(state["trace"], axis=-1)
+
+            return x, log_density
+
+        def deltas(time, xz):
+            return {"xz": self.velocity(xz, sigma=time, conditions=conditions, training=training)}
+
+        state = {"xz": z}
+        state = integrate(
+            deltas,
+            state,
+            **integrate_kwargs,
+        )
+
+        x = state["xz"]
+
+        return x
+
+    def compute_metrics(
+        self,
+        x: Tensor | Sequence[Tensor, ...],
+        conditions: Tensor = None,
+        sample_weight: Tensor = None,
+        stage: str = "training",
+    ) -> dict[str, Tensor]:
+        training = stage == "training"
+        if not self.built:
+            xz_shape = keras.ops.shape(x)
+            conditions_shape = None if conditions is None else keras.ops.shape(conditions)
+            self.build(xz_shape, conditions_shape)
+
+        # sample log-noise level
+        log_sigma = self.p_mean + self.p_std * keras.random.normal(
+            ops.shape(x)[:1], dtype=ops.dtype(x), seed=self.seed_generator
+        )
+        # noise level with shape (batch_size, 1)
+        sigma = ops.exp(log_sigma)[:, None]
+
+        # generate noise vector
+        z = sigma * keras.random.normal(ops.shape(x), dtype=ops.dtype(x), seed=self.seed_generator)
+
+        # calculate preconditioning
+        c_skip = self._c_skip_fn(sigma)
+        c_out = self._c_out_fn(sigma)
+        c_in = self._c_in_fn(sigma)
+        c_noise = self._c_noise_fn(sigma)
+        xz_pre = c_in * (x + z)
+
+        # calculate output of the network
+        if conditions is None:
+            xtc = keras.ops.concatenate([xz_pre, c_noise], axis=-1)
+        else:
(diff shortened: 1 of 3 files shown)


arrjon:
Thanks @vpratz for the implementation! `sigma_data` is already specific for the EDM version. I think, we have to make it even more general, so some additional arguments which can be passed, depending on the type of schedule one wants to use. 

I plan to add on top of this additional schedules and samplers until the end of the week.

arrjon:
Part of the pre-conditioning can be expressed as a special kind of weighting function: see appendix D.1 in [here](https://proceedings.neurips.cc/paper_files/paper/2023/file/ce79fbf9baef726645bc2337abb0ade2-Paper-Conference.pdf).

So yes, the aim would be to have one nice framework!

arrjon:
I added some more noise schedules and started to make the implementation more general. This is just a first draft, so you @vpratz get an idea, how we could do it. We should discuss this then and how to move forward.

arrjon:
I added a class `NoiseSchedule` and different schedules, so it should be easy now to extend to more schedules if necessary. Since EDM has a specific sampling scheme for inference, this is now also defined in the noise schedule. Therefore, we do not have to specify specific sampling step sizes anymore.

Next step would be add stochastic samplers as well.

arrjon:
I just added a simpler euler maruyama stochastic solver

vpratz:
Thanks a lot for the fixes, they increase the performance of the new implementation a lot. The old standalone EDM implementation seems to be a little bit better still, but the difference might be down to hyperparameter tuning. I have added the `examples/experimental/Two_Moons_Diffusion_Comparison.ipynb` notebook, which allows testing both implementations on the same benchmark, and plotting the results against each other. This is for two moons, but feel free to expand it with other benchmarks as well.

As far as I can tell, the open steps before we finalize this PR are:
- optimising performance: ensuring that we achieve the same performance as with the standalone EDM implementation, so that we do not forego performance by not including it
- related: find and set good defaults
- remove the `LinearNoiseSchedule` (as discussed privately, maybe moving it to a tutorial): the other schedules perform better, so we do not need to include and maintain it here
- add tests to cover all relevant cases/combinations
- proof-reading docstrings and maybe supplying a tutorial
- remove the `examples/experimental/Two_Moons_Diffusion_Comparison.ipynb` notebook before merging

Did I miss anything, @arrjon , or do you have any other comments on the current state?

arrjon:
The performance issue is fixed now, it was mainly due to a missing scaling factor of the log_snr, which goes into the network.

I am also implementing the stochastic sampler: it is working for all backend but jax at the moment. After this, only the things @vpratz mentioned are missing.

arrjon:
The stochastic sampler is now also working for jax. So all features done for the moment!

vpratz:
Great! Thanks a lot for putting in the work and for the quick fixes! I'll try to add the relevant tests and work on some of the other missing things in the next few days.

vpratz:
Ok, I think the most important things are covered, and the PR is ready for review. I suggest we keep the class in the experimental module for a few weeks/months for additional testing, and if we do not encounter issues we can stabilize it.
@LarsKue could you conduct a review for this? I think you can mostly focus on the style, not on the specific equations.

vpratz:
See the commit message of [8402a3f](https://github.com/bayesflow-org/bayesflow/pull/408/commits/8402a3f71b76c565ee97bfe580dd83f7364c2762) for details regarding the tests.

arrjon (approved):

LarsKue:
Sorry for not reviewing earlier. I will add my review ASAP.

bayesflow/utils/integrate.py
```diff
@@ -293,3 +294,115 @@ def integrate(
...
+    current_state = state.copy()
+
+    # keras.ops.fori_loop does not support keras seed generator in jax
+    for i in range(steps):
+        # Execute the step with the specific seed for this step
+        current_state, time = step_fn(
+            state=current_state,
+            time=time,
+            step_size=step_size,
+        )
```
LarsKue, line 406: As discussed, we should add an issue that mentions potential performance improvements by refactoring this to use `keras.ops.fori_loop`. @arrjon can you add this issue and mention some of the methods you have tried already to make this work?

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+            Default is "edm".
+        prediction_type: str, optional
+            The type of prediction used in the diffusion model. Can be "velocity", "noise" or "F" (EDM).
+             Default is "F".
+        **kwargs
+            Additional keyword arguments passed to the subnet and other components.
+        """
+        super().__init__(base_distribution="normal", **kwargs)
+
+        if isinstance(noise_schedule, str):
```
LarsKue, line 356: Use a dispatch function instead (see utils/dispatch for some examples). Does not necessarily need to live in utils, if you would have issues with circular imports.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+            if noise_schedule == "cosine":
+                noise_schedule = CosineNoiseSchedule()
+            elif noise_schedule == "edm":
+                noise_schedule = EDMNoiseSchedule()
+            else:
+                raise ValueError(f"Unknown noise schedule: {noise_schedule}")
+        elif not isinstance(noise_schedule, NoiseSchedule):
+            raise ValueError(f"Unknown noise schedule: {noise_schedule}")
```
LarsKue, line 364: This should be a `TypeError`.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+    jacobian_trace,
+    layer_kwargs,
+    weighted_mean,
+    integrate,
+    integrate_stochastic,
+)
+
+
+@serializable
+class NoiseSchedule(ABC):
```
LarsKue, line 25: please separate this into multiple files

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+    def from_config(cls, config, custom_objects=None):
+        return cls(**deserialize(config, custom_objects=custom_objects))
+
+    def validate(self):
+        """Validate the noise schedule."""
+        if self.log_snr_min >= self.log_snr_max:
+            raise ValueError("min_log_snr must be less than max_log_snr.")
+        for training in [True, False]:
+            if not ops.isfinite(self.get_log_snr(0.0, training=training)):
+                raise ValueError("log_snr(0) must be finite.")
```
LarsKue, line 137: It might be valuable to show in the message what value the `training` argument takes for when this errors.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+    We can also define a weighting function for each noise level for the loss function. Often the noise schedule is
+    the same for the forward and reverse process, but this is not necessary and can be changed via the training flag.
+
+    [1] Variational Diffusion Models 2.0: Understanding Diffusion Model Objectives as the ELBO with Simple Data
+        Augmentation: Kingma et al. (2023)
+    """
+
+    def __init__(self, name: str, variance_type: str, weighting: str = None):
+        self.name = name
+        self._variance_type = variance_type  # 'exploding' or 'preserving'
```
LarsKue, line 44: use a literal type hint instead: `Literal["exploding", "preserving"]`

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        self.log_snr_min = -15  # should be set in the subclasses
+        self.log_snr_max = 15  # should be set in the subclasses
```
LarsKue, line 46: If this should be set in the subclass, set it to `None` here.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        if self._variance_type == "preserving":
+            # variance preserving schedule
+            alpha_t = ops.sqrt(ops.sigmoid(log_snr_t))
+            sigma_t = ops.sqrt(ops.sigmoid(-log_snr_t))
+        elif self._variance_type == "exploding":
+            # variance exploding schedule
+            alpha_t = ops.ones_like(log_snr_t)
+            sigma_t = ops.sqrt(ops.exp(-log_snr_t))
+        else:
+            raise ValueError(f"Unknown variance type: {self._variance_type}")
```
LarsKue, line 104: Should be a TypeError

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        elif self._weighting == "sigmoid":
+            # sigmoid weighting based on Kingma et al. (2023)
+            return ops.sigmoid(-log_snr_t + 2)
+        elif self._weighting == "likelihood_weighting":
+            # likelihood weighting based on Song et al. (2021)
+            g_squared = self.get_drift_diffusion(log_snr_t=log_snr_t)
+            sigma_t = self.get_alpha_sigma(log_snr_t=log_snr_t, training=True)[1]
+            return g_squared / ops.square(sigma_t)
+        else:
+            raise ValueError(f"Unknown weighting type: {self._weighting}")
```
LarsKue, line 122: TypeError

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        elif self._variance_type == "exploding":
+            # variance exploding schedule
+            alpha_t = ops.ones_like(log_snr_t)
+            sigma_t = ops.sqrt(ops.exp(-log_snr_t))
+        else:
+            raise ValueError(f"Unknown variance type: {self._variance_type}")
+        return alpha_t, sigma_t
+
+    def get_weights_for_snr(self, log_snr_t: Tensor) -> Tensor:
+        """Get weights for the signal-to-noise ratio (snr) for a given log signal-to-noise ratio (lambda). Default is 1.
```
LarsKue, line 108: Confusing docs: Is the default for `log_snr_t=1` or `exp(log_snr_t)=1`, or is 1 the default for the return value?

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+@serializable
+class CosineNoiseSchedule(NoiseSchedule):
+    """Cosine noise schedule for diffusion models. This schedule is based on the cosine schedule from [1].
+    For images, use s_shift_cosine = log(base_resolution / d), where d is the used resolution of the image.
+
+    [1] Diffusion Models Beat GANs on Image Synthesis: Dhariwal and Nichol (2022)
+    """
+
+    def __init__(
+        self, min_log_snr: float = -15, max_log_snr: float = 15, s_shift_cosine: float = 0.0, weighting: str = "sigmoid"
```
LarsKue, line 159: This needs argument docs

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        self._t_max = self.get_t_from_log_snr(log_snr_t=self.log_snr_min, training=True)
+
+    def _truncated_t(self, t: Tensor) -> Tensor:
+        return self._t_min + (self._t_max - self._t_min) * t
+
+    def get_log_snr(self, t: Union[float, Tensor], training: bool) -> Tensor:
+        """Get the log signal-to-noise ratio (lambda) for a given diffusion time."""
+        t_trunc = self._truncated_t(t)
+        # SNR = -2 * log(tan(pi*t/2))
+        return -2 * ops.log(ops.tan(math.pi * t_trunc / 2)) + 2 * self._s_shift_cosine
```
LarsKue, line 176: The comment and the code differ. Where is the offset in the comment?

Also, nitpick: use `0.5 * x` instead of `x / 2` for best performance.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+
+@serializable
+class EDMNoiseSchedule(NoiseSchedule):
+    """EDM noise schedule for diffusion models. This schedule is based on the EDM paper [1].
+    This should be used with the F-prediction type in the diffusion model.
+
+    [1] Elucidating the Design Space of Diffusion-Based Generative Models: Karras et al. (2022)
+    """
+
+    def __init__(self, sigma_data: float = 1.0, sigma_min: float = 1e-4, sigma_max: float = 80.0):
```
LarsKue, line 213: Needs argument docs

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        ----------
+        subnet : str or type, optional
+            The architecture used for the transformation network. Can be "mlp" or a custom
+            callable network. Default is "mlp".
+        integrate_kwargs : dict[str, any], optional
+            Additional keyword arguments for the integration process. Default is None.
+        subnet_kwargs : dict[str, any], optional
+            Keyword arguments passed to the subnet constructor or used to update the default MLP settings.
+        noise_schedule : str or NoiseSchedule, optional
+            The noise schedule used for the diffusion process. Can be "cosine" or "edm".
```
LarsKue, line 346: 'If str, can be "cosine" or "edm".'

Again, in this case you should use a `Literal` type hint, unless there may be many more possibilities for future implementations.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        self.noise_schedule.validate()
+
+        if prediction_type not in ["noise", "velocity", "F"]:  # F is EDM
+            raise ValueError(f"Unknown prediction type: {prediction_type}")
+        self._prediction_type = prediction_type
+        self._loss_type = kwargs.get("loss_type", "noise")
+        if self._loss_type not in ["noise", "velocity", "F"]:
+            raise ValueError(f"Unknown loss type: {self._loss_type}")
+        if self._loss_type != "noise":
+            warnings.warn(
```
LarsKue, line 376: For plain warnings, use `bayesflow.utils.logging.warning` instead

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+
+        self.integrate_kwargs = self.INTEGRATE_DEFAULT_CONFIG | (integrate_kwargs or {})
+        self.seed_generator = keras.random.SeedGenerator()
+
+        if subnet_kwargs:
+            warnings.warn(
+                "Using `subnet_kwargs` is deprecated."
+                "Instead, instantiate the network yourself and pass the arguments directly.",
+                DeprecationWarning,
+            )
```
LarsKue, line 396: Since this is a new network, let's not support `subnet_kwargs` to begin with.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+
+        subnet_kwargs = subnet_kwargs or {}
+        if subnet == "mlp":
+            subnet_kwargs = self.MLP_DEFAULT_CONFIG | subnet_kwargs
+
+        self.subnet = find_network(subnet, **subnet_kwargs)
+        self.output_projector = keras.layers.Dense(units=None, bias_initializer="zeros")
+
+        self._kwargs = kwargs
```
LarsKue, line 405: This should not be saved.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+
+        self.subnet.build(input_shape)
+        out_shape = self.subnet.compute_output_shape(input_shape)
+        self.output_projector.build(out_shape)
+
+    def get_config(self):
+        base_config = super().get_config()
+        base_config = layer_kwargs(base_config)
+
+        config = self._kwargs | {
```
LarsKue, line 431: Should remove the `self._kwargs` since all valid `kwargs` are already contained in the `super().get_config()`.

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+
+    def _transform_log_snr(self, log_snr: Tensor) -> Tensor:
+        """Transform the log_snr to the range [-1, 1] for the diffusion process."""
+        # Transform the log_snr to the range [-1, 1]
+        return (
+            2
+            * (log_snr - self.noise_schedule.log_snr_min)
+            / (self.noise_schedule.log_snr_max - self.noise_schedule.log_snr_min)
+            - 1
+        )
```
LarsKue, line 538: this would look much more readable with some (temporary) variable renaming

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        def f(x):
+            return self.velocity(x, time=time, stochastic_solver=False, conditions=conditions, training=training)
+
+        v, trace = jacobian_trace(f, xz, max_steps=max_steps, seed=self.seed_generator, return_output=True)
+
+        return v, ops.expand_dims(trace, axis=-1)
+
+    def _transform_log_snr(self, log_snr: Tensor) -> Tensor:
+        """Transform the log_snr to the range [-1, 1] for the diffusion process."""
+        # Transform the log_snr to the range [-1, 1]
```
LarsKue, line 532: This comment is superfluous as it matches the docstring above

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        **kwargs,
+    ) -> Tensor | tuple[Tensor, Tensor]:
+        integrate_kwargs = (
+            {
+                "start_time": 0.0,
+                "stop_time": 1.0,
+            }
+            | self.integrate_kwargs
+            | kwargs
+        )
```
LarsKue, line 555: Use a multi-line statement instead of a bracketed statement:

```py3
integrate_kwargs = {"start_time": 0.0, "stop_time": 1.0}
integrate_kwargs = integrate_kwargs | self.integrate_kwargs
integrate_kwargs = integrate_kwargs | kwargs
```

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        **kwargs,
+    ) -> Tensor | tuple[Tensor, Tensor]:
+        integrate_kwargs = (
+            {
+                "start_time": 1.0,
+                "stop_time": 0.0,
+            }
+            | self.integrate_kwargs
+            | kwargs
+        )
```
LarsKue, line 609: Use a multi-line statement (see above)

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        noise_schedule_training_stage = stage == "training" or stage == "validation"
+        if not self.built:
+            xz_shape = ops.shape(x)
+            conditions_shape = None if conditions is None else ops.shape(conditions)
+            self.build(xz_shape, conditions_shape)
+
+        # sample training diffusion time as low discrepancy sequence to decrease variance
+        # t_i = \mod (u_0 + i/k, 1)
+        u0 = keras.random.uniform(shape=(1,), dtype=ops.dtype(x), seed=self.seed_generator)
+        i = ops.arange(0, ops.shape(x)[0], dtype=ops.dtype(x))  # tensor of indices
```
LarsKue, line 683: avoid in-line comments if you can

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        t = (u0 + i / ops.cast(ops.shape(x)[0], dtype=ops.dtype(x))) % 1
+        # i = keras.random.randint((ops.shape(x)[0],), minval=0, maxval=self._timesteps)
+        # t = ops.cast(i, ops.dtype(x)) / ops.cast(self._timesteps, ops.dtype(x))
```
LarsKue, line 686: this can be removed

bayesflow/experimental/diffusion_model.py
```diff
@@ -0,0 +1,737 @@
...
+        eps_t = keras.random.normal(ops.shape(x), dtype=ops.dtype(x), seed=self.seed_generator)
+
+        # diffuse x
+        diffused_x = alpha_t * x + sigma_t * eps_t
+
+        # calculate output of the network
+        if conditions is None:
+            xtc = ops.concatenate([diffused_x, self._transform_log_snr(log_snr_t)], axis=-1)
+        else:
+            xtc = ops.concatenate([diffused_x, self._transform_log_snr(log_snr_t), conditions], axis=-1)
```
LarsKue, line 705: use `bayesflow.utils.tensor_utils.concatenate_valid` instead

bayesflow/utils/integrate.py
```diff
@@ -4,10 +4,11 @@
 import keras
 
 import numpy as np
-from typing import Literal
+from typing import Literal, Union, List
```
LarsKue, line 7: prefer `collections.abc` for types like `Sequence` instead of `List`. If you really need a list (and not a tuple, for example), you can use either the python built-in `list` or `collections.abc.MutableSequence` instead.

bayesflow/utils/integrate.py
LarsKue: Looking at this again, it seems like we could fix the previous issue by simply passing the seed generator as an explicit variable of the loop state, i.e. drop it from the partial above, which effectively lambda-captures it, leading to side-effects, and instead pass it as part of the `loop_state`.

bayesflow/experimental/diffusion_model.py
arrjon: what exactly do you mean here?

arrjon: got it

arrjon:
@LarsKue I included your feedback. Thank you for the extensive review! Could you have a look at the serialization, I think this is not not working anymore with the new release and I am not sure, what has to be done here.

The last thing to do for me, is to resolve the last open comment.

bayesflow/utils/integrate.py
arrjon: I already tried this. This gives me
`Argument '<keras.src.random.seed_generator.SeedGenerator object at 0x1683be890>' of type '<class 'keras.src.random.seed_generator.SeedGenerator'>' is not a valid JAX type`

vpratz:
I will update the file structure a bit, to put everything into a directory. I will move the dispatch function there as well, to avoid circular imports while also fixing the serialization issue.

vpratz:
The docs test is failing now, without apparent change from our side, so maybe it is a dependency thing. I cannot reproduce it locally yet, though. I'll see how to fix this...

bayesflow/utils/integrate.py
arrjon: this is now solved by precomputing the noise

arrjon:
The stochastic sampler is now as efficient as the ODE solver by precomputing the noise. It works for all backends now!

vpratz:
The tests are passing now, I think we are ready for a (hopefully) final review, @LarsKue . Thanks a lot @arrjon for implementing the changes!

bayesflow/utils/integrate.py
LarsKue: Nice solution! I did some more experiments with trying to pass the seed to the loop, but unfortunately, keras and jax are not very compatible in that aspect.

Do we want to allow random values returned from the `drift_fn` and `diffusion_fn`? If so, using `fori_loop` seems problematic in any case.

LarsKue (changes requested):
Very cool changes! Some minor things remain, see individual comments.

bayesflow/utils/integrate.py
arrjon: At the moment `drift_fn` and `diffusion_fn` are always deterministic, so the current solution is fine.

bayesflow/experimental/diffusion_model/dispatch.py
```diff
@@ -0,0 +1,59 @@
+from functools import singledispatch
+from .noise_schedules import NoiseSchedule
+
+
+@singledispatch
+def find_noise_schedule(arg, *args, **kwargs):
```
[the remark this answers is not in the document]
vpratz: I will add the tests on Monday

bayesflow/experimental/diffusion_model/noise_schedules.py
```diff
@@ -0,0 +1,339 @@
...
+        dsnr_dt = dsnr_dx * (self._t_max - self._t_min)
+        factor = ops.exp(-log_snr_t) / (1 + ops.exp(-log_snr_t))
+        return -factor * dsnr_dt
+
+    def get_config(self):
+        return dict(min_log_snr=self.log_snr_min, max_log_snr=self.log_snr_max, shift=self._shift)
+
+    @classmethod
+    def from_config(cls, config, custom_objects=None):
+        return cls(**deserialize(config, custom_objects=custom_objects))
```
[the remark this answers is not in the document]
vpratz: I will add serialization tests, but would leave the rest of the tests to end-to-end testing.

bayesflow/utils/integrate.py
vpratz: Can we mark this as resolved?

vpratz:
I have run the tests for all variants locally without failure. As the tests are very slow, I decided to skip all but two variants (EDM/F, Cosine/velocity), at least while it is in the experimental module. As it is still in experimental we have to be less shy to change things if we should encounter problems.
Thanks to @arrjon and @niels-leif-bracher for the initial discussions, @arrjon for providing most of the implementation, and @LarsKue for the reviews!

Issue #278, opened 2024-12-11
New adapter transforms: subsetting arrays


paul-buerkner:
Sometimes, the simulators may return larger arrays than we want to use in our networks, so it would be great to have some subsetting adapter transforms.

We need:

- Subsetting within an axis (taking only some elements) while keeping the number of axes the same. This is essentially the `np.take` functionality so we might want to call this transform `take`. In contrast to `np.take`, I would make `axis` a mandatory arguments or default it to the last axis. Example: `adapter.take("x", 1:3, axis = -1)`
- Subsetting using a random set of indices (of user-specified size) within an axis. We might call this `subsample`. Internally it would call `take` after sampling the indices. Example: `adapter.subsample("x", size = 3, axis = -1)`
- Removing an axis of length one. Following numpy, I would call this transform `squeeze`: Example: `adapter.squeeze("x", axis = 1)`

CorentinChab:
Hello @paul-buerkner , can I be assigned to this issue ?

marvinschmitt:
> Hello @paul-buerkner , can I be assigned to this issue ?

I have assigned you, thanks for volunteering to work on this issue!

CorentinChab:
Thanks ! @stefanradev93 told me there was a Slack channel. Can I join so we can agree on the test specification I am writing, before developping actual implementation ? It's my first time doing something like this so I might have some questions 😓 Hope not be a burden haha

marvinschmitt:
> Thanks ! @stefanradev93 told me there was a Slack channel. Can I join so we can agree on the test specification I am writing, before developping actual implementation ? It's my first time doing something like this so I might have some questions 😓 Hope not be a burden haha
> 

No worries at all, all contributions are greatly appreciated and you're more than welcome to ask all the questions you have at any point :) 

Could you please share your email address to mail.marvinschmitt@gmail.com, then I'll send you a Slack invite.

eodole:
Hello @CorentinChab @marvinschmitt  I had talked to @paul-buerkner about working on this issue but didnt actually get assigned to it. I already had some work done on it. @CorentinChab would you want to collaborate on this?

vpratz:
Closed by #411, thanks to @eodole and @LarsKue!

PR #411, opened 2025-04-14
Subset arrays

eodole:
Addresses Issue #278 

Implemented Take and subsample methods, however as discussed with Lars the requested squeeze functionality is essentially just the inverse of expand dims


tests/test_links/test_links.py
```diff
@@ -30,9 +30,9 @@ def check_ordering(output, axis):
     assert np.all(np.diff(output, axis=axis) > 0), f"is not ordered along specified axis: {axis}."
     for i in range(output.ndim):
         if i != axis % output.ndim:
-            assert not np.all(np.diff(output, axis=i) > 0), (
-                f"is ordered along axis which is not meant to be ordered: {i}."
-            )
+            assert not np.all(
+                np.diff(output, axis=i) > 0
+            ), f"is ordered along axis which is not meant to be ordered: {i}."
```
LarsKue, line 35: I'm not sure why this is being reordered now, are you running ruff version `0.11.2`?

bayesflow/adapters/transforms/take.py
```diff
@@ -0,0 +1,24 @@
...
+    A transform to reduce the dimensionality of arrays output by the summary network
+    Axis is a mandatory argument and will default  to the last axis.
+    Example: adapter.take("x", 1:3, axis = -1)
+
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, data, indices, axis=-1):
```
LarsKue, line 19: IMO, the `indices` need to be passed in the constructor for this transform. See above.

bayesflow/adapters/transforms/subsample_array.py
```diff
@@ -0,0 +1,30 @@
...
+    A transform that takes a random subsample of the data within an axis.
+
+    Example: adapter.subsample("x", sample_size = 3, axis = -1)
+
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, data: np.ndarray, sample_size: int, axis=-1):
```
LarsKue, line 19: IMO, the `sample_size` should be part of the constructor for this transform. It seems like a bit of a hassle to have to pass this argument in the forward call of the Adapter, unless you have a specific use case in mind?

bayesflow/adapters/transforms/subsample_array.py
```diff
@@ -0,0 +1,30 @@
...
+        max_sample_size = data.shape[axis]
+
+        sample_indices = np.random.permutation(max_sample_size)[
+            0 : sample_size - 1
+        ]  # random sample without replacement
```
LarsKue, line 24: We could optionally let the user specify that they want to sample with replacement.

LarsKue (changes requested):
Thank you for the PR! The core of these changes already looks good, but there are some things I would like to see changed. Please also change the base branch of the PR to `dev`. See [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request) on how to do that.

bayesflow/adapters/transforms/subsample_array.py
LarsKue: I agree. It's not a priority to implement this. Let's keep it as-is.

bayesflow/adapters/transforms/subsample_array.py
stefanradev93: I would also allow it to be a float in [0, 1] specifying a **proportion** of the sample to subsample.

eodole (commented):
I'm honestly not sure that this should be a map transform as written, one consequence of the way its written now is that all keys specified by this transform will have random subsamples of the same size from the same axis.  I think that all datasets that a user wants to subsample should be specified individually, so that axis and sample size are specified individually. If this is the case, would it not be better to force the map transform to reject a sequence of keys but rather only take one key?

tests/test_links/test_links.py
eodole: apparently I was running ruff 0.8.1 but I will update it

eodole:
You also asked me to rename subsample_array to random_subsample my question is should all associated files also be renamed?

LarsKue:

.gitignore
```diff
@@ -39,3 +39,6 @@ docs/
 
 # MacOS
 .DS_Store
+
+# Rproj
+.Rproj.user
```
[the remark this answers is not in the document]
LarsKue: according to @stefanradev93, this should be `.Rproj`

bayesflow/adapters/transforms/subsample_array.py
LarsKue: I second this, but it raises the concern of whether we should floor or ceil the resulting value. I am thinking ceiling should be better.

eodole: what case do you imagine for use of ceiling over floor? For most applications I assume users would want a slightly smaller rather than larger sample size.

bayesflow/adapters/adapter.py
```diff
@@ -541,6 +543,47 @@ def one_hot(self, keys: str | Sequence[str], num_classes: int):
...
+        
+        if isinstance(key, Sequence[str]) and len(keys) >1:
+            TypeError("`key` should be either a string or a list of length one. Only one dataset may be modified at a time.")
+
+        if isinstance(key, str):
+            keys = [key]
+
+        transform = MapTransform(
+            transform_map={
+                key:RandomSubsample(sample_size=sample_size, axis=axis)
```
LarsKue, line 579: Looks like the linter did not run. Did you install pre-commit as shown in `CONTRIBUTING.md` with `pre-commit install`? For the workflows to run, you will also need to merge dev into this first and resolve any conflicts.

bayesflow/adapters/adapter.py
```diff
@@ -618,7 +693,7 @@ def to_array(
         include : str or Sequence of str, optional
             Names of variables to include in the transform.
         exclude : str or Sequence of str, optional
-            Names of variables to exclude from the transform.
+            Names of variabxles to exclude from the transform.
```
LarsKue, line 696: typo

bayesflow/adapters/adapter.py
```diff
@@ -631,3 +706,9 @@ def to_array(
         )
         self.transforms.append(transform)
         return self
+    
+    
+    
+    
+    
+
```
LarsKue, line 714: These empty lines would be removed by the formatter, which should automatically run on pre-commit.

bayesflow/adapters/transforms/random_subsample.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
+from keras.saving import register_keras_serializable as serializable
```
LarsKue, line 2: Please use the new serialization pipeline in `bayesflow.utils.serialization`:

```py3
from bayesflow.utils.serialization import deserialize, serializable, serialize
```

bayesflow/adapters/transforms/random_subsample.py
```diff
@@ -0,0 +1,47 @@
...
+        max_sample_size = data.shape[axis]
+        
+        if isinstance(self.sample_size, int):
+            sample_size = self.sample_size
+        else: 
+            sample_size = np.round(self.sample_size * max_sample_size)
+
+        sample_indices = np.random.permutation(max_sample_size)[
+            0 : sample_size - 1
+        ]  # random sample without replacement
```
LarsKue, line 41: please put comments above the line they are referring to, rather than in-line

LarsKue (changes requested):
Thank you for addressing my previous comments. Before we can merge this into dev, it looks like you will need to resolve merge-conflicts by merging dev into this first. If you need help with that, let me know.

eodole:
I have no idea why the linter failed, i have it in the precommit hook and i merged dev into this branch. Im not really sure why it doesnt pass

bayesflow/adapters/adapter.py
eodole: i made a new environment based on the contribution.md
and now the linter runs automatically. on my end the most recent commit it passed all checks so im not sure why it's failing on the pull request

LarsKue:
@eodole I did some clean-up to make the linter pass and I adjusted the transforms, docs, and tests to what I think your intention was. Please double-check.

The tests are still failing, however, due to some incorrect shape broadcasting. Can you take care of fixing this?

LarsKue (changes requested):
See comment above.

LarsKue:
I just realized that it's simply an issue with the tests, since they assume invertibility and determinism at some points. I also implemented the serialization pipeline for the transforms, so all tests should pass now. Thanks for the PR!

bayesflow/adapters/adapter.py
LarsKue: Try running `pre-commit run --all-files` manually once.

LarsKue (approved):

PR #416, opened 2025-04-15
Add Pandas `to_dict` adapter transform

LarsKue:
Usage:

```py3
adapter = bf.Adapter().to_dict()
df = pd.DataFrame(...)
adapter(df)
>>> dict(...)
```

Automatically converts any `object` and `category` columns to one-hot encodings.

This is nice for datasets that are based on `pandas.DataFrame` without requiring too much manual conversion in those datasets. This transform should always be used first, as it assumes the sole input is a `pandas.DataFrame`.

Docs are still missing, if anyone wants to contribute this feel free to!


LarsKue:
@paul-buerkner if test pass, could you please add your review and then merge?

Issue #417, opened 2025-04-16
Default `stage` for transforms (especially `Standardize`)


vpratz:
While using BayesFlow, I have managed multiple times to ruin my trained adapter when trying to use it directly, as the naive `adapter(data)` call defaults to `stage="training"` for the `Standardize` transform. I would propose changing this, as I think it is a footgun that can mess up the pipeline/results after training without the user noticing necessarily. It's also not obvious that one has to be on the lookout for this, as the momentum is enabled by default.

@LarsKue @stefanradev93 What do you think?

vpratz:
Closed by #418

PR #418, opened 2025-04-16
Set default stage in adapter to "inference"

vpratz:
Concerns #417.
Question:
As Lars already highlighted when creating the current implementation, we cannot pass the stage to the `__getitem__` methods of the datasets. How do we want to handle this?
For now, I have made the stage that will be used configurable in the `__init__` function.
What should its default be? For now, I have set it to "training". The alternative would be to set it to "inference" and dynamically change it to "training" in `Approximator.fit`.
What do you think, @LarsKue and @stefanradev93? Setting this to a draft until we have discussed this...

Minor point: Do we want to convert the stages to an [enum](https://docs.python.org/3/library/enum.html). This would be a bit more typing (but autocomplete should help), and might prevent future bugs due to typos in the strings. Not sure whether it is worth it though, what do you think?

### DIFF
--- a/bayesflow/adapters/adapter.py
+++ b/bayesflow/adapters/adapter.py
@@ -79,13 +79,15 @@ class Adapter(MutableSequence[Transform]):
     def get_config(self) -> dict:
         return {"transforms": serialize(self.transforms)}
 
-    def forward(self, data: dict[str, any], **kwargs) -> dict[str, np.ndarray]:
+    def forward(self, data: dict[str, any], *, stage: str = "inference", **kwargs) -> dict[str, np.ndarray]:
         """Apply the transforms in the forward direction.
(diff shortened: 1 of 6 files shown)


LarsKue:
Datasets will pretty much only ever be used directly in training. I am fine with this change, although I do wish we had a better alternative available (such as simply not requiring a stage argument for the Adapter, since it should not be trainable in the first place).

LarsKue (commented):
I don't think we should change the `stage` argument  to an enum, because either users will then have to make another import (cumbersome), or we make it compatible with strings anyway, in which case this becomes the primary use case.

stefanradev93:
I suggest we stick to the simplest solution as proposed by Valentin.

Issue #245, opened 2024-11-07
Data adapter: Keep track of the jacobian for variable transforms


Kucharssim:
Given that the data adapter is now used for variable transforms (for e.g., implementing parameter constraints or scaling), it would be nice to keep track of the jacobian of the transform, so that if the user wants to use the `log_prob` method, it is possible to automatically obtain the log prob on the "constrained/data space" instead of the "unconstrained/network space". 

For custom transforms, the user would have to supply corresponding `log_det_jacobian` function, but of course that should be only optional to those cases where the user wants to use `log_prob` in the first place.

I know we talked about this with @LarsKue and @paul-buerkner already at different time points, just making it an issue so that we can keep track of it here.

stefanradev93:
Hi Simon, I like the idea! Would you like to try a draft implementation a few pull requests down the line?

Kucharssim:
Sure, I can give it a go :)

Kucharssim:
Sorry for sitting on this.

I have made progress on this at some point and then let it get stale, I have now rebased it on the current `dev` branch. You can check the functionality here: https://github.com/Kucharssim/BayesFlow/tree/jacobian

I have not done the jacobian for the inverse (though it shouldn't be that difficult for the most part), and there are still some transforms for which I haven't implemented the jacobian (e.g., `NumpyTransform`). But the basic use case works on that branch.

Since I am not a pythonista, I am sure I didn't implement this in the cleanest way possible. If you have some tips how to simplify the code, that would be very much appreciated.

I can in the meantime look into implementing the missing features, and then I can make a PR for this. Does that sound ok?

LarsKue:
Hi @Kucharssim, I checked out your fork and from my side it looks fine. In Normalizing Flows, the convention is to compute the `log_det` along with the forward call for efficiency, but I can see how that is not as relevant for Adapter Transforms since they are cheap, and also more difficult to implement with the current state of the Adapter.

That said, if you want to copy this pattern, feel free to! Please make a PR whenever you deem the feature ready for review.

Kucharssim:
Yeah, I saw that pattern, but it seems to me that the majority of adapter calls happen during training, for which we typically do not need the log_det_jac at all, only the forward transform (whereas with normalizing flows, we need the log_det_jac for training). So computing the jacobians along with the forward transforms seems to me a bit wasteful. But I would be happy to refactor this if you have a strong preference for it.

LarsKue:
No strong preference on my side.

PR #419, opened 2025-04-17
Adapter keeps track of the transform jacobians

Kucharssim:
fixes https://github.com/bayesflow-org/bayesflow/issues/245

I have not implemented it for `NumpyTransform` because I was not sure about implementation w.r.t. serialization. I tried to make sure that if we cannot keep track of the jacobians, we raise an error rather than returning an incorrect output.

### DIFF
--- a/bayesflow/adapters/adapter.py
+++ b/bayesflow/adapters/adapter.py
@@ -80,7 +80,7 @@ class Adapter(MutableSequence[Transform]):
         return serialize(config)
 
     def forward(
-        self, data: dict[str, any], *, stage: str = "inference", jacobian: bool = False, **kwargs
+        self, data: dict[str, any], *, stage: str = "inference", log_det_jac: bool = False, **kwargs
     ) -> dict[str, np.ndarray] | tuple[dict[str, np.ndarray], dict[str, np.ndarray]]:
         """Apply the transforms in the forward direction.
 
@@ -90,7 +90,7 @@ class Adapter(MutableSequence[Transform]):
             The data to be transformed.
         stage : str, one of ["training", "validation", "inference"]
             The stage the function is called in.
-        jacobian: bool, optional
+        log_det_jac: bool, optional
             Whether to return the log determinant jacobians of the transforms.
         **kwargs : dict
             Additional keyword arguments passed to each transform.
@@ -98,10 +98,10 @@ class Adapter(MutableSequence[Transform]):
         Returns
         -------
         dict | tuple[dict, dict]
-            The transformed data or tuple of transformed data and jacobians.
+            The transformed data or tuple of transformed data and log determinant jacobians.
         """
         data = data.copy()
-        if not jacobian:
+        if not log_det_jac:
             for transform in self.transforms:
                 data = transform(data, stage=stage, **kwargs)
             return data
@@ -114,7 +114,7 @@ class Adapter(MutableSequence[Transform]):
         return data, log_det_jac
 
     def inverse(
-        self, data: dict[str, np.ndarray], *, stage: str = "inference", jacobian: bool = False, **kwargs
+        self, data: dict[str, np.ndarray], *, stage: str = "inference", log_det_jac: bool = False, **kwargs
     ) -> dict[str, np.ndarray] | tuple[dict[str, np.ndarray], dict[str, np.ndarray]]:
         """Apply the transforms in the inverse direction.
 
@@ -124,7 +124,7 @@ class Adapter(MutableSequence[Transform]):
             The data to be transformed.
         stage : str, one of ["training", "validation", "inference"]
             The stage the function is called in.
-        jacobian: bool, optional
+        log_det_jac: bool, optional
             Whether to return the log determinant jacobians of the transforms.
         **kwargs : dict
             Additional keyword arguments passed to each transform.
@@ -132,10 +132,10 @@ class Adapter(MutableSequence[Transform]):
         Returns
         -------
         dict | tuple[dict, dict]
-            The transformed data or tuple of transformed data and jacobians.
+            The transformed data or tuple of transformed data and log determinant jacobians.
         """
         data = data.copy()
-        if not jacobian:
+        if not log_det_jac:
             for transform in reversed(self.transforms):
                 data = transform(data, stage=stage, inverse=True, **kwargs)
             return data
@@ -166,7 +166,7 @@ class Adapter(MutableSequence[Transform]):
         Returns
         -------
         dict | tuple[dict, dict]
-            The transformed data or tuple of transformed data and jacobians.
+            The transformed data or tuple of transformed data and log determinant jacobians.
         """
         if inverse:
             return self.inverse(data, stage=stage, **kwargs)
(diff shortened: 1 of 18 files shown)


LarsKue:
Thanks for the PR. Could you add tests for this behavior @Kucharssim?

Kucharssim:
Yes, on it!

LarsKue:

bayesflow/adapters/transforms/rename.py
```diff
@@ -58,3 +58,6 @@ def inverse(self, data: dict[str, any], *, strict: bool = False, **kwargs) -> di
 
     def extra_repr(self) -> str:
         return f"{self.from_key!r} -> {self.to_key!r}"
+
+    def log_det_jac(self, data: dict[str, any], log_det_jac: dict[str, any], inverse: bool = False, **kwargs):
+        return self.inverse(data=log_det_jac) if inverse else self.forward(data=log_det_jac, strict=False)
```
LarsKue, line 63: I think this should be the default method on `Transform`, since you override the default on a bunch of methods that have no sensible definition for a jacobian.

bayesflow/adapters/transforms/standardize.py
```diff
@@ -120,3 +120,12 @@ def inverse(self, data: np.ndarray, **kwargs) -> np.ndarray:
         std = np.broadcast_to(self.std, data.shape)
 
         return data * std + mean
+
+    def log_det_jac(self, data, inverse: bool = False, **kwargs) -> np.ndarray:
+        if self.std is None:
+            return None
```
LarsKue, line 126: Should this not raise an error?

bayesflow/approximators/continuous_approximator.py
```diff
@@ -387,11 +387,15 @@ def log_prob(self, data: Mapping[str, np.ndarray], **kwargs) -> np.ndarray | dic
         np.ndarray
             Log-probabilities of the distribution `p(inference_variables | inference_conditions, h(summary_conditions))`
         """
-        data = self.adapter(data, strict=False, stage="inference", **kwargs)
+        data, jacobian = self.adapter(data, strict=False, stage="inference", jacobian=True, **kwargs)
```
LarsKue, line 390: let's call this variable `log_det`

bayesflow/approximators/continuous_approximator.py
```diff
@@ -387,11 +387,15 @@ def log_prob(self, data: Mapping[str, np.ndarray], **kwargs) -> np.ndarray | dic
...
         data = keras.tree.map_structure(keras.ops.convert_to_tensor, data)
         log_prob = self._log_prob(**data, **kwargs)
         log_prob = keras.tree.map_structure(keras.ops.convert_to_numpy, log_prob)
 
+        jacobian = jacobian.get("inference_variables")
+        if jacobian is not None:
+            log_prob = log_prob + jacobian
```
LarsKue, line 397: I would add a comment that this is the change of variables formula

bayesflow/adapters/transforms/concatenate.py
```diff
@@ -115,3 +115,37 @@ def extra_repr(self) -> str:
...
+    ) -> dict[str, np.ndarray]:
+        # copy to avoid side effects
+        log_det_jac = log_det_jac.copy()
+
+        if inverse:
+            if log_det_jac.get(self.into) is not None:
+                raise ValueError(
+                    "Cannot obtain an inverse jacobian of concatenation. "
+                    "Transform your variables before you concatenate."
+                )
```
LarsKue, line 136: I don't understand the nature of this error, and it is not tested. Could you add a test or an explaining comment? The former would be preferred, if you have capacity to (GPT generated unit tests for these small untested lines are fine)

bayesflow/adapters/adapter.py
```diff
@@ -88,22 +90,32 @@ def forward(self, data: dict[str, any], *, stage: str = "inference", **kwargs) -
...
+            Whether to return the log determinant jacobians of the transforms.
         **kwargs : dict
             Additional keyword arguments passed to each transform.
 
         Returns
         -------
-        dict
-            The transformed data.
+        dict | tuple[dict, dict]
+            The transformed data or tuple of transformed data and jacobians.
```
LarsKue, line 101: You are referring to the log determinant of the Jacobian as simply Jacobians everywhere. I feel like this would be confusing to users, who expect to be returned the full matrix. Can we change that? My preferred terms are `log_det` or `log_det_jac`or "log determinant of the Jacobian" for docs.

LarsKue (changes requested):
Thank you for implementing this! I noted some minor concerns, mostly regarding clarity.

bayesflow/adapters/adapter.py
Kucharssim: Good point, I will use `log_det_jac`

bayesflow/adapters/transforms/rename.py
Kucharssim: Hm I am not sure, because this works only for some of the `Transform` classes, but not all...

bayesflow/adapters/transforms/concatenate.py
Kucharssim: Ok, added a test for this

bayesflow/adapters/transforms/standardize.py
Kucharssim: You are right, but this also revealed a deeper issue with my implementation, thanks! The problem is that the `log_det_jac` here was called before the `forward` pass, and since `standardize` is a stateful layer, the result of `log_det_jac` would be based on values that are out of date - here `None` for the first time the `standardize` is called. I flipped the order so that the forward is always called before `log_det_jac`, but this imo also shows that it would be better to just compute the transform and the `log_det_jac` at the same time as you pointed out earlier... 😅

bayesflow/approximators/continuous_approximator.py
Kucharssim: done

bayesflow/adapters/transforms/rename.py
Kucharssim: Just to expand on this, this only works for `Transform` that only manipulates the keys of `data`, e.g., `drop`, `keep`, `rename`. Other `Transforms`, such as `concatenate`, need their own implementation, when the rest (e.g., `to_dict`) do not do anything wrt the `log_det_jac`.

LarsKue (approved):
Thank you for the changes! I think we are mostly ready to merge this as-is. Can you address the open conversation above and the comments from Copilot? Thanks!

bayesflow/adapters/adapter.py
```diff
@@ -88,22 +90,33 @@ def forward(self, data: dict[str, any], *, stage: str = "inference", **kwargs) -
...
+        if not log_det_jac:
+            for transform in self.transforms:
+                data = transform(data, stage=stage, **kwargs)
+            return data
 
+        log_det_jac = {}
         for transform in self.transforms:
-            data = transform(data, stage=stage, **kwargs)
+            transformed_data = transform(data, stage=stage, **kwargs)
+            log_det_jac = transform.log_det_jac(data, log_det_jac, **kwargs)
```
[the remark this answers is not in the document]
stefanradev93: I don't think that this is correct. The J is a function of the original data.

Kucharssim:
Ok, I think I addressed everything now. If you are happy with this, feel free to merge

PR #422, opened 2025-04-18
Better Loss Plot

jerrymhuang:
bf_new_loss

Fixed and refined `bf.diagnostics.plots.loss`:

* Replace the moving average with an exponential moving average and make it functional and by default;
* Provide the user with a new set of customization, including moving average window (span), color maps, markers, and grid;
* Make the validation loss a viridis gradient line.

This version of loss plot also propagates to the `workflow` object.

### DIFF
--- a/bayesflow/diagnostics/plots/loss.py
+++ b/bayesflow/diagnostics/plots/loss.py
@@ -1,4 +1,4 @@
-from typing import Sequence
+from collections.abc import Sequence
 
 import numpy as np
 import pandas as pd
@@ -14,8 +14,7 @@ def loss(
     history: keras.callbacks.History,
     train_key: str = "loss",
     val_key: str = "val_loss",
-    moving_average: bool = True,
-    moving_average_alpha: float = 0.8,
+    smoothing_factor: float = 0.8,
     figsize: Sequence[float] = None,
     train_color: str = "#132a70",
     val_color: str = None,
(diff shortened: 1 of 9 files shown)


bayesflow/diagnostics/plots/loss.py
```diff
@@ -1,4 +1,4 @@
-from collections.abc import Sequence
+from typing import Sequence
```
LarsKue, line 1: We should prefer `collections.abc`.

bayesflow/diagnostics/plots/loss.py
```diff
@@ -7,21 +7,25 @@
...
 
 def loss(
     history: keras.callbacks.History,
     train_key: str = "loss",
     val_key: str = "val_loss",
-    moving_average: bool = False,
-    per_training_step: bool = False,
-    ma_window_fraction: float = 0.01,
+    moving_average: bool = True,
+    moving_average_alpha: float = 0.8,
```
LarsKue, line 18: Calling this `alpha` is a bit confusing with respect to opacity. I would prefer `smoothing_factor` or `moving_average_smoothing_factor`.

LarsKue (changes requested):
Looks great! I will add the changes that I proposed myself.

LarsKue:
Although it is very pretty, I am removing the gradient line for now, until we have resolved how we can reliably add it to the legend. I also simplified the rest of the code and fixed an issue with the smoothing factor being applied the wrong way around.

LarsKue (approved):

Issue #423, opened 2025-04-20
`TimeSeriesTransformer` summary network not deserializable


vpratz:
Loading the network fails with the error message below (when adding the network to the summary network test suite).
I do not know how to fix this yet, I will open a draft PR with the added test and we can discuss how to resolve it there.

``` verbatim
error_msgs = {139725321209360: (<Dense name=dense_47, built=True>, ValueError("Layer 'dense_47' expected 2 variables, but received 0 variables during loading. Expected: ['kernel', 'bias']"))}
warn_only = False                                                                                                      
                                                                                                                       
    def _raise_loading_failure(error_msgs, warn_only=False):                                                           
        first_key = list(error_msgs.keys())[0]                                                                         
        ex_saveable, ex_error = error_msgs[first_key]                                                                  
        msg = (                                                                                                        
            f"A total of {len(error_msgs)} objects could not "                                                         
            "be loaded. Example error message for "                                                                    
            f"object {ex_saveable}:\n\n"                                                                               
            f"{ex_error}\n\n"                                                                                          
            "List of objects that could not be loaded:\n"                                                              
            f"{[x[0] for x in error_msgs.values()]}"                                                                   
        )                                                                                                              
        if warn_only:                                                                                                  
            warnings.warn(msg)                                                                                         
        else:                                                                                                          
>           raise ValueError(msg)                                                                                      
E           ValueError: A total of 1 objects could not be loaded. Example error message for object <Dense name=dense_47, built=True>:
E                                                                                                                      
E           Layer 'dense_47' expected 2 variables, but received 0 variables during loading. Expected: ['kernel', 'bias']
E                                                                                                                      
E           List of objects that could not be loaded:                                                                  
E           [<Dense name=dense_47, built=True>]                                                                        
                                                                                                                       
/data/Programming/.mamba/envs/bf2/lib/python3.11/site-packages/keras/src/saving/saving_lib.py:631: ValueError          
_ test_save_and_load[summary_dim=4-batch_size=3-set_size=3-feature_size=2-summary_network='time_series_transformer'] __
                                                                                                                       
tmp_path = PosixPath('/tmp/pytest-of-valentin/pytest-4/test_save_and_load_summary_dim27')                              
summary_network = <TimeSeriesTransformer name=time_series_transformer_45, built=True>                                  
random_set = <tf.Tensor: shape=(3, 3, 2), dtype=float32, numpy=                                                        
array([[[-1.0457249 , -0.03069962],                                                                                    
[...]

PR #424, opened 2025-04-20
Serialization of `TimeSeriesTransformer`

vpratz:
This draft PR is intended as a place to fix the deserialization of the `TimeSeriesTransformer`, described in #423.

The deserialization of the `TimeSeriesTransformer` is broken, but it went undetected as this network is not included in the tests yet. I already added the test, but have no fix yet. I suggest we try to come up with a fix quickly and then merge tests and fix into `dev` together.

### DIFF
--- a/bayesflow/networks/transformers/fusion_transformer.py
+++ b/bayesflow/networks/transformers/fusion_transformer.py
@@ -2,8 +2,8 @@ import keras
 from keras import layers
 
 from bayesflow.types import Tensor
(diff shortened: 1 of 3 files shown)


vpratz:
The `config` functions were missing, I included them following the example from `SetTransform`. One difference is that I included `kwargs` in the serialization as well. I'm not sure yet whether we want this, but I think we can discuss this at a future point in time. Self-merging this, as it seems to fix the problem for now.

PR #425, opened 2025-04-21
Adapt inference network tests to support non-invertible architectures

vpratz:
The `ConsistencyModel` class was previously untested. as the inference network test assume the networks are invertible. This PR is a basic solution only relevant to the `ConsistencyModel` class for now, that relies on a `NotImplementedError` to skip the respective tests. If you have other wishes/ideas for this case, please let me know.


LarsKue:

PR #427, opened 2025-04-22
Add advice regarding moving from v1 to v2 to README.

### LINKED ISSUE #388: Compile list of functionality not yet ported to v2
I have repeatedly encountered people asking whether some functionality is available, that was in v1 but is not yet in v2. We should compile a list of those functionalities and display it prominently, so that people can decide whether they already want to make the switch.
Discovering that some functionality does not exist after you already invested some effort into porting code can be very disappointing, so I think this is important to have.

Please add the things you are aware of in this issue. From a quick comparison of the docs, I see

- Sensitivity: functionality from the `sensitivity` module
- MCMC (discontinued): `mcmc` module, this functionality will probably be covered by a PyMC wrapper in the future.
- c2st metric
- from the simulation module: `TwoLevelGenerativeModel`, `TwoLevelPrior`
- networks: `EvidentialNetwork`, `PMPNetwork`

I know there are future plans to cover some of them in a more general setting, but I think it would be great to inform users up front what they are getting and not getting in the new version.

@stefanradev93 @paul-buerkner @LarsKue It would be great if you could complete the list, of give a go when you consider it complete.

vpratz:
As we do not have full feature parity yet, I think it is important to be up-front with the functionality that is still missing, to avoid confusion and disappointment. As we now have the 2.0.0 version on PyPI, some advice on whether to upgrade or not might be helpful.

This PR adds the list of missing functionality from #388, as well as some additional considerations. Feel free to edit, but I think getting something like this out soon would be good.

Closes #388.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,18 +1,53 @@
+name: Publish Python 🐍 distribution 📦 to PyPI
 
-name: Publish to PyPI.org
 on:
   release:
     types: [published]
+
 jobs:
-  pypi:
+  build:
+    name: Build distribution 📦
     runs-on: ubuntu-latest
+
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - run: python3 -m pip install -U build && python3 -m build
-      - name: Publish package
-        uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          password: ${{ secrets.PYPI_API_TOKEN }}
+    - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.x"
+    - name: Install pypa/build
+      run: >-
+        python3 -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: python3 -m build
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v4
+      with:
+        name: python-package-distributions
+        path: dist/
+
+  publish-to-pypi:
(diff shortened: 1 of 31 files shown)


vpratz:
I have added a short phrase regarding the v1 docs, and a link to the migration advice so people can find it.

stefanradev93:
@paul-buerkner What do you think? I believe we can shorten some of the description, as most of the old functionality is probably not relevant. The two level model is probably the most important gap which I intend to close soon.

paul-buerkner:
I like it. About shortening, do you mean the new edits of this PR or older stuff in the vignette? Having some more details cannot hurt I would say. Why would you want to shorten it?

paul-buerkner:
Edit: I now realize this is supposed to be part of the readme not the vignette. Yes, I agree we should shorten it a lot.
Or perhaps better, put the caution sign to the top but all the details down somewhere in the FAQ (and a link to it at the top). It is not helpful for a readme file if I have to read a lot about old stuff without seeing all the cool new stuff first.

vpratz:
I have moved the details to the FAQ. Any thoughts on how to best highlight the section in the top part? I'm not sure whether it needs a separate heading, or some highlighting is enough.
I can see that presenting new features is the exciting part with the new release, but I think at least in the next few weeks we should also focus on our existing user base, so that they do not get the impression that there is a new version and "nothing works anymore".
As the new version is on PyPI now, and some users may upgrade by accident, I think having this information somewhere prominent soon is important.

vpratz:
Again, feel free to edit as you see fit. If you are happy with the text, I will copy the FAQ section to the FAQ on the website as well...

paul-buerkner (approved):
Looks good to me. Thank you! From my side, we can merge as is.

Issue #429, opened 2025-04-22
Error running Two Moons notebook: flow matching default config produces error


vpratz:
If I am not mistaken, the default flow matching configuration on `dev` seems to be broken currently. This probably also concerns the `2.0.0` release.

Running the two moons notebook gives the following error message:

```verbatim
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[12], line 1
----> 1 history = flow_matching_workflow.fit_offline(
      2     training_data, 
      3     epochs=epochs, 
      4     batch_size=batch_size, 
      5     validation_data=validation_data
      6 )

File [~/Programming/IWR/bf2/bayesflow/workflows/basic_workflow.py:714](http://127.0.0.1:8892/lab/tree/examples/bayesflow/workflows/basic_workflow.py#line=713), in BasicWorkflow.fit_offline(self, data, epochs, batch_size, keep_optimizer, validation_data, **kwargs)
    679 """
    680 Train the approximator offline using a fixed dataset. This approach will be faster than online training,
    681 since no computation time is spent in generating new data for each batch, but it assumes that simulations
   (...)
    709     metric evolution over epochs.
    710 """
    712 dataset = OfflineDataset(data=data, batch_size=batch_size, adapter=self.adapter)
--> 714 return self._fit(
    715     dataset, epochs, strategy="online", keep_optimizer=keep_optimizer, validation_data=validation_data, **kwargs
    716 )

File [~/Programming/IWR/bf2/bayesflow/workflows/basic_workflow.py:913](http://127.0.0.1:8892/lab/tree/examples/bayesflow/workflows/basic_workflow.py#line=912), in BasicWorkflow._fit(self, dataset, epochs, strategy, keep_optimizer, validation_data, **kwargs)
    910     self.approximator.compile(optimizer=self.optimizer, metrics=kwargs.pop("metrics", None))
    912 try:
--> 913     self.history = self.approximator.fit(
    914         dataset=dataset, epochs=epochs, validation_data=validation_data, **kwargs
    915     )
    916     self._on_training_finished()
    917     return self.history

File [~/Programming/IWR/bf2/bayesflow/approximators/continuous_approximator.py:200](http://127.0.0.1:8892/lab/tree/examples/bayesflow/approximators/continuous_approximator.py#line=199), in ContinuousApproximator.fit(self, *args, **kwargs)
    148 def fit(self, *args, **kwargs):
    149     """
[...]

PR #431, opened 2025-04-22
Hotfix Version 2.0.1

LarsKue:
See #429


vpratz:
I can hit "Restore branch", you probably can do it as well. Happened to me too, recently. Maybe we can protect dev to not be deletable?

PR #434, opened 2025-04-22
Enable use of summary networks with functional API again

vpratz:
@thegialeo noticed a regression when trying to use BayesFlow summary networks with the functional API. This PR fixes the regression, and adds a test for the summary networks to detect it in the future.

I have extended the `sanitize_input_shape` decorator to support multiple `..._shape` arguments. We might rename it to `sanitize_shape_args`. What do you think?


LarsKue (approved):
Thanks for the PR!

PR #435, opened 2025-04-22
Drastically Improve Speed of Import

LarsKue:
Importing BayesFlow has been very slow. Here is an example:

```py3
%%time
import bayesflow as bf
>>> CPU times: user 13 s, sys: 2.62 s, total: 15.6 s
>>> Wall time: 13 s
```

Some of this time is taken up by importing keras, which differs by the backend used:

```py3
%%time
import keras  # using tensorflow backend
>>> CPU times: user 3.88 s, sys: 294 ms, total: 4.17 s
>>> Wall time: 1.55 s
```

Using a simple line-profiler, I could track the issue to `inspect.stack()`, `inspect.getmodule()` and `inspect.ismodule()`. On import, these are primarily used in 2 places:

1. in `_add_imports_to_all()`
2. in `@serializable`

I refactored these utilities to avoid the expensive calls to `inspect`, making the import time now primarily dominated by the import to the respective deep learning framework:

```py3
%%time
import bayesflow as bf
CPU times: user 4.29 s, sys: 819 ms, total: 5.11 s
Wall time: 2.5 s
```

So, for our library, this is a speed-up of approximately $(13s - 1.55s) / (2.5s - 1.55s) \approx 1200$%.

Since this is a sensitive change to the library, I would like to ask that you take extra care when reviewing this. @vpratz I requested your review primarily for the `_add_imports_to_all` function. Could you add a test that checks if it is working correctly? @stefanradev93 I would like you to sign off on merging this as well.

### DIFF
--- a/bayesflow/utils/_docs/_populate_all.py
+++ b/bayesflow/utils/_docs/_populate_all.py
@@ -1,4 +1,5 @@
-import inspect
+import sys
+import types
 
 
 def _add_imports_to_all(include_modules: bool | list[str] = False, exclude: list[str] | None = None):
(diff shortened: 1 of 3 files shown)


vpratz:
Awesome, looking forward to having this! I have compared the outputs of the doc generation with and without your change to `_populate_all.py`, and the same files are produced, so I think those changes are working as intended.

bayesflow/utils/_docs/_populate_all.py
vpratz, line 1: I have tested the changes in this file via the docs generation, looks good to me.

PR #449, opened 2025-04-27
Document and expose serialization module, update developer docs

vpratz:

PR #450, opened 2025-04-27
Deprecate old serialization (`(de)serialize_value_or_type`) and add developer docs

vpratz:
The functionality was replaced by the more powerful `serialize` and `deserialize` functions. Therefore I would suggest we deprecate the old functions.

It was still use in the point estimation setting, this change would break loading of previously saved models that use those classes. I just noticed that the `serializable` adapter is currently producing wrong packages (I will open and link an issue soon). I first thought that it was related to my changes, and already fixed it in a1b4d19d5. But the problem is also present on `dev` and `main`.
As the fix for this also breaks backwards compatibility, I think doing this deprecation now might be a good.

This PR is a continuation of #449 which also contains the fix, so the diff will simplify once #449 is merged (or we merge both jointly by merging this one).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          pip install -U pip setuptools wheel
+          python -m pip install -U pip setuptools wheel
           pip install .[test]
 
       - name: Install JAX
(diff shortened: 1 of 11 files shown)


vpratz:
I cannot get the PointInferenceNetwork to work with the new functions. @han-ol Do you have time to take a look at this? If not, we have to postpone this until we have figured out how to achieve this...

vpratz:
Thanks for your comments. Regarding your question: Yes, in my first attempt I tried to streamline it a bit, but it failed with error messages I could not learn a lot from, unfortunately. They might well be related to the issue, but I did not explore it further. As the new functions do a very similar thing as the old ones, my guess would be that we cannot resolve this yet.

LarsKue (approved):
Thank you for the PR and for the added docs!

Issue #441, opened 2025-04-25
Different sets of parameters in the `ModelComparisonSimulator`


vpratz:
This bug was reported in this [Discourse thread](https://discuss.bayesflow.org/t/getting-error-using-modelcomparisonsimulator-in-bf-2-0-1/150), thanks to [dizyd](https://discuss.bayesflow.org/u/dizyd) for reporting. The `ModelComparisonSimulator` throws an error when different prior parameters are specified for the different simulators. The minimal reproducible example is:

```python
import bayesflow as bf
import numpy as np

rng = np.random.default_rng()

def prior_1():
    return dict(w=rng.uniform())

def prior_2():
    return dict(c=rng.uniform())


def model_1(w):
    return dict(x = w)


def model_2(c):
    return dict(x = c)

simulator_1 = bf.make_simulator([prior_1,model_1])
simulator_2 = bf.make_simulator([prior_2,model_2])

simulator = bf.simulators.ModelComparisonSimulator(simulators=[simulator_1, simulator_2])
simulator.sample(10)
```

The line that throws the error is `sims = tree_concatenate(sims, numpy=True)`.

I'm not familiar with that code, maybe @Kucharssim, @stefanradev93 or @LarsKue can take a look at it.

Kucharssim:
In situations like these, you would need to set `use_mixed_batches=False` and then it should work. But I think we can do better here and help the user a little bit more. I will have a look.

vpratz:
Do I understand it correctly that we do not use the parameters for inference? If this is the case, then maybe a "smarter" concatenation function (for example one that pads with `nan` if a parameter is not available for a given model) would be an option as well. The user would then be responsible to keep/use only the shared outputs, e.g. by using `Adapter.keep()`. What do you think?

stefanradev93:
@LarsKue @paul-buerkner I think generally our adapters should silently keep only the respective reserved keywords before passing the data to the network. This is also related to https://github.com/bayesflow-org/bayesflow/issues/439 by @arrjon

paul-buerkner:
I agree.

LarsKue:
In general, the Adapters are currently almoat entirely unaware of both the simulator and the Approximator. This makes them easy to write and maintain, but of course, limited in some aspects. We could think about changing this paradigm, although I think the Adapters are currently very clean.

vpratz:
I agree with @LarsKue, and would at the very least make the reserved keywords easily configurable.
[...]

PR #452, opened 2025-04-29
`ModelComparisonSimulator`: handle different outputs from individual simulators

Kucharssim:
fixes https://github.com/bayesflow-org/bayesflow/issues/441

As per https://github.com/bayesflow-org/bayesflow/issues/441#issuecomment-2830285318, this PR implements

> "smarter" concatenation function (for example one that pads with nan if a parameter is not available for a given model). 

However, by default the simulator will just drop (with an info warning) keys that are not common for all simulators, since in most situations we would not need those outputs in the first place.

### DIFF
--- a/bayesflow/simulators/model_comparison_simulator.py
+++ b/bayesflow/simulators/model_comparison_simulator.py
@@ -42,10 +42,12 @@ class ModelComparisonSimulator(Simulator):
             If neither `p` nor `logits` is provided, defaults to uniform logits.
         use_mixed_batches : bool, optional
             Whether to draw samples in a batch from different models.
+
             - If True (default), each sample in a batch may come from a different model.
             - If False, the entire batch is drawn from a single model, selected according to model probabilities.
         key_conflicts : str, optional
             Policy for handling keys that are missing in the output of some models, when using mixed batches.
+
             - "drop" (default): Drop conflicting keys from the batch output.
             - "fill": Fill missing keys with the specified value.
             - "error": An error is raised when key conflicts are detected.
@@ -53,7 +55,7 @@ class ModelComparisonSimulator(Simulator):
             If `key_conflicts=="fill"`, the missing keys will be filled with the value of this argument.
         shared_simulator : Simulator or Callable, optional
             A shared simulator whose outputs are passed to all model simulators. If a function is
-            provided, it is wrapped in a `LambdaSimulator` with batching enabled.
+            provided, it is wrapped in a :py:class:`~bayesflow.simulators.LambdaSimulator` with batching enabled.
         """
         self.simulators = simulators
 
--- a/bayesflow/simulators/model_comparison_simulator.py
+++ b/bayesflow/simulators/model_comparison_simulator.py
@@ -6,6 +6,7 @@ from bayesflow.utils import tree_concatenate
 from bayesflow.utils.decorators import allow_batch_size
 
 from bayesflow.utils import numpy_utils as npu
+from bayesflow.utils import logging
 
 from types import FunctionType
 
@@ -22,6 +23,7 @@ class ModelComparisonSimulator(Simulator):
         p: Sequence[float] = None,
         logits: Sequence[float] = None,
         use_mixed_batches: bool = True,
+        key_conflicts: str | float = "drop",
         shared_simulator: Simulator | Callable[[Sequence[int]], dict[str, any]] = None,
     ):
         """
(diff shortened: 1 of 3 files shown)


vpratz:
Thanks for the PR, I skimmed it and like the idea behind the changes. I'll try to conduct a proper review some time this week.

bayesflow/simulators/model_comparison_simulator.py
```diff
@@ -38,8 +40,14 @@ def __init__(
...
-            If True, samples in a batch are drawn from different models. If False, the entire batch
-            is drawn from a single model chosen according to the model probabilities. Default is True.
+            Whether to draw samples in a batch from different models.
+            - If True (default), each sample in a batch may come from a different model.
+            - If False, the entire batch is drawn from a single model, selected according to model probabilities.
+        key_conflicts : {"drop"} | float, optional
+            Policy for handling keys that are missing in the output of some models, when using mixed batches.
+            - "drop" (default): Drop conflicting keys from the batch output.
+            - float: Fill missing keys with the specified value.
+            - If neither "drop" nor a float is given, an error is raised when key conflicts are detected.
```
LarsKue, line 50: I would make this into `key_conflicts="error"`.

bayesflow/simulators/model_comparison_simulator.py
```diff
@@ -38,8 +40,14 @@ def __init__(
             A sequence of logits corresponding to model probabilities. Mutually exclusive with `p`.
             If neither `p` nor `logits` is provided, defaults to uniform logits.
         use_mixed_batches : bool, optional
-            If True, samples in a batch are drawn from different models. If False, the entire batch
-            is drawn from a single model chosen according to the model probabilities. Default is True.
+            Whether to draw samples in a batch from different models.
+            - If True (default), each sample in a batch may come from a different model.
+            - If False, the entire batch is drawn from a single model, selected according to model probabilities.
+        key_conflicts : {"drop"} | float, optional
```
LarsKue, line 46: I get the idea behind `key_conflicts=float`, but I think it is confusing from the outside. Instead, I would choose something like `key_conflicts="replace"` or `"fill"` and have an additional argument `fill_value=None`.

LarsKue (commented):
Looks good from my side. See individual comments.

Can we also add tests for the (few) missed edge-case lines?

bayesflow/simulators/model_comparison_simulator.py
Kucharssim: Yes, that improves clarity of the interface. Thanks!

bayesflow/simulators/model_comparison_simulator.py
Kucharssim: Done

bayesflow/simulators/model_comparison_simulator.py
```diff
@@ -118,3 +133,58 @@ def sample(self, batch_shape: Shape, **kwargs) -> dict[str, np.ndarray]:
...
+                    shape[0] = batch_sizes[i]
+                    sim[missing_key] = np.full(shape=shape, fill_value=self.fill_value)
+            return sims
+        elif self.key_conflicts == "error":
+            raise ValueError("Key conflicts are found in simulator outputs, cannot combine them into one batch.")
+
+    def _determine_key_conflicts(self, sims):
+        # determine only once
+        if self._keys is not None:
+            return self._keys
```
vpratz, line 168: Can this return "wrong" results if some simulators had `n=0` in line 120 when the function first runs, and `n>0` later on? Is this something we want to safeguard against?

vpratz: I can imagine this function to be quite cheap to compute, would it make sense to run it completely every time (but only logging the info once)?

vpratz (commented):
See my comment on one edge case, I'm not sure if it is a relevant one. What do you think?
Apart from that, the PR looks good to me, I only added minor formatting fixes to the docstring.

bayesflow/simulators/model_comparison_simulator.py
LarsKue: I agree

tests/test_simulators/test_simulators.py
```diff
@@ -52,8 +53,19 @@ def test_fixed_sample(composite_gaussian, batch_size, fixed_n, fixed_mu):
...
+    if multimodel_key_conflicts.key_conflicts == "drop":
+        samples = multimodel_key_conflicts.sample(batch_size)
         assert set(samples) == {"x", "model_indices"}
-    else:
+    elif multimodel_key_conflicts.key_conflicts == "fill":
+        samples = multimodel_key_conflicts.sample(batch_size)
         assert set(samples) == {"x", "model_indices", "c", "w"}
         assert np.sum(np.isnan(samples["c"])) + np.sum(np.isnan(samples["w"])) == batch_size
+    elif multimodel_key_conflicts.key_conflicts == "error":
+        with pytest.raises(Exception):
```
LarsKue, line 70: This is too broad of a check. Use specific exception types.

LarsKue (changes requested):
The changes are looking great. Can we address Valentin's comments before we merge? I also left some minor comments, still.

bayesflow/simulators/model_comparison_simulator.py
Kucharssim: ok, will do that

vpratz (approved):
Thanks a lot for the revised changes, they look good to me now.

PR #453, opened 2025-04-30
Add projectors to DeepSet

han-ol:
This adds projectors after the MLPs in the DeepSet network.

Without these projectors an activation function is applied to the learnable residual.
It is preferable to not bound that residual from below.

I construct the MLPs with all widths but the last, then add a new dense layer of the last width.

Changing the MLPs and adding corresponding new projector attributes both breaks serialization. We could in principle detect checkpoints of the DeepSet old format and translate it, if we think this is worth the effort at this stage.


stefanradev93:
I would simply pass all layers to the inner MLP and keep the projector. Also, you wanted to select better defaults for init / activation. Do you want to include them as part of this PR?

LarsKue (changes requested):
Let's implement Stefan's requested changes before merging this.

LarsKue (approved):

PR #456, opened 2025-05-02
Fix redirects from old docs to stable-legacy

vpratz:
The redirects went to v1.1.6, but we have chosen the name `stable-legacy` for the old docs instead. This PR adjusts the links in the files used for the redirects.


LarsKue (approved):
Thank you for the fix!

vpratz:
@LarsKue @stefanradev93 I think we need to update the list of required checks to reflect that we currently do not test Python 3.11. Could one of you adapt this, please?

PR #457, opened 2025-05-02
serialization: apply new scheme for package (breaking change)

### LINKED ISSUE #451: [BUG] `serializable` decorator produces wrong package for serialization
**To Reproduce**
On `dev` or `main`, run
```python
import bayesflow as bf
import keras

keras.saving.get_registered_name(bf.Adapter)  # or any other class without "package" set in the serializable decorator
```
Output:
```
'bayesflow.utils.decorators>Adapter'
```

**Expected behavior**
The path should be the path to the module, so
```
'bayesflow.adapters.adapter>Adapter'
```

**Solution**
In `utils.serialization.serializable`, increase the `depth` in the `sys._getframe` call to `2`.

**Additional context**
I came across this for #449 and #450, and first assumed my changes to the allow_args decorator had caused it, so I already fixed it in a1b4d19d5, which is included in those PRs. Upon further testing, I noticed the problem already existed in `dev` and `main`.

The fix will break loading of serialized files created while the bug was present, but I think fixing it quickly and putting out an update is probably the best way forward. We might want to combine this with the deprecation of our old serialization helper functions in #450.
I think we should reflect on whether we want to keep the functionality of `serializable` like this. In my opinion, it brings a certain fragility, because moving classes to a different file or folder breaks deserialization, even when the class itself did not change. Requiring `package` to be set manually will remove this footgun, and thereby remove one cause for accidentally breaking serializability.

vpratz:
Closes #451.

- introduces new policy for consistent naming for serilization (see https://github.com/bayesflow-org/bayesflow/issues/451
  for a discussion): standard is the path of a module a class resides
  in, trucated at depth two. So for a class in bayesflow.networks, we
  set package="bayesflow.networks", even if the live in the
  bayesflow.networks.mlp submodule.
- The `serializable` decorator checks this and errors if this is not
  followed. The check can be disabled for certain cases (e.g., classes
  in the experimental module, that might eventually live somewhere
  else).
- After this commit, previously saved models will not be loadable. As we
  introduced a bug regarding this anyway (https://github.com/bayesflow-org/bayesflow/issues/451), we will accept this and
  should inform users about it.
- usage of direct calls to `keras.saving.register_keras_serializable`
  were replaced with our custom decorator.
- update docs and add a disclaimer to README regarding potential breaking changes until 2.1 release (feel free to adapt)

Todo for next patch release
----

- as this breaks loading models from the previous patch versions, make a brief announcement in our community channels, so that people are aware and can pin their installation, if necessary
- include changes from #453 (if we want to include them), which are breaking as well


bayesflow/adapters/transforms/serializable_custom_transform.py
```diff
@@ -2,17 +2,18 @@
 import numpy as np
 from keras.saving import (
     deserialize_keras_object as deserialize,
-    register_keras_serializable as serializable,
     serialize_keras_object as serialize,
     get_registered_name,
     get_registered_object,
 )
```
LarsKue, line 8: Can we switch to using deserialize and serialize from our new pipeline here?

LarsKue (commented):
We should think about whether we want the current module check to be a flag in the serializable decorator. If users are supposed to use our pipeline rather than the one from keras, the check must be off by default.

I think one of the following options is better:
1. We add a check in any `test_serialize_deserialize` for the module structure
2. We leave the flag in the current `serializable` decorator, but turn it off by default. Internal modules can use something like `from bayesflow.internal import serializable` which is a `partial(bayesflow.utils.serializable, disable_module_check=False)`

I am in favor of 1, since it avoids the issue of 2 that new contributors may accidentally use the "wrong" kind of decorator.

bayesflow/adapters/transforms/serializable_custom_transform.py
vpratz: Good point, I have changed it.

vpratz:
Thanks for your comments. I like the idea of moving the test to the test suite, but as far as I can tell (correct me if I'm wrong) the `test_serialize_deserialize` test only explicitly covers a subset of classes. Many others are implicitly part of the test, but I do not know if we can access those for the test we have in mind.
I agree with your concern regarding 2.
A third option would be to limit the check to the case where the module name indicates a class is part of BayesFlow. I will add a commit that implements this, but we can still decide to go with one of the other options.

Would you say that we actually want users to to use our decorator instead of the Keras-provided one? Currently the only thing it does is to supply this check, which is only relevant internally, so we could also think about making it private. Or do we expect to encounter future scenarios where we want to have additional functionality in this decorator? As we cannot rule out that users use the Keras decorator instead, I think this would not be so desirable anyways.

What do you think?

LarsKue:
Thanks for the update. I think users should use our decorator, as this gives us the most flexibility in terms of maneuvering around keras.

PR #462, opened 2025-05-07
Networks docstrings: add summary/inference network indicator

vpratz:
From the table in the `bayesflow.networks` module overview, one cannot tell whether a network is a summary network, an inference network, or neither. This PR adds short labels to indicate inference networks (IN) and summary networks (SN), so one can easily identify them.

The information is redundant, as the classes are also subclasses of InferenceNetwork/SummaryNetwork, but until now, this information isn't easily accessible to end users (you have to click on the class first to see the "Bases: ..." label). This is one possible fix. Do you like it this way or do you have different ideas on how to fix this?


vpratz:
@stefanradev93 @LarsKue Any thoughts/objections to this?

LarsKue (approved):

PR #463, opened 2025-05-07
WIP: User guide (initial structure/content)

vpratz:
As discussed with @han-ol and @LarsKue, a user guide enables us to provide more in-depth information on each topic, concentrated in one place. This is a nice addition to the examples, which touch on all aspects of the library, but only as detailed as necessary for a given application.

The structure provided here is intended as a starter, and roughly follows the structure of the Linear Regression Starter notebook. I have copied some paragraphs from the notebook, and also written a few additional sections. I have opted for Jupyter notebooks, as they allow us to add demo code as well.

What do you think would be a good approach? Should we keep it in a separate branch until we have a reached a somewhat complete/coherent state? Or do we just put it out with an appropriate warning regarding its completeness, and update it as we go?


paul-buerkner:
This looks great to me! I would be in favor of merging it soon (with a warning) and then update as we go. That way, users can already benefit from the existing material and see that we are actively working on documentation and user support.

PR #469, opened 2025-05-08
Stabilize MultivariateNormalScore by constraining initialization in PositiveDefinite link

### LINKED ISSUE #468: MultivariateNormalScore is vulnerable to large input values
Training with `MultivariateNormalScore` occasionally fail in the first step, when inputs (conditions) are high at initialization. Summary networks generally don't ensure that output variance at initialization is near 1. I stubled upon this with DeepSet, but don't expect it to be limited to just that.

**To Reproduce**

Initialize a PointInferenceNetwork with MultivariateNormalScore and try to invert covariance matrix right after building.

```python
import bayesflow as bf
import keras
import numpy as np

# Initialize fresh network
network = bf.networks.PointInferenceNetwork(
    scores=dict(mvn=bf.scores.MultivariateNormalScore())
)

# Random input with high variance in conditions
xz = keras.ops.convert_to_tensor(np.random.normal(0, 1, size=(64, 3)))
conditions = keras.ops.convert_to_tensor(np.random.normal(0, 10, size=(64, 2)))

# Attempt to invert the predicted covariance matrices
covariance = network(xz=xz, conditions=conditions)["mvn"]["covariance"]
keras.ops.inv(covariance)  # fails, not invertible

```
**Expected behavior**
A lack of input normalization should not break training. (We nevertheless can expect performance degradation, which would be a separate issue to solve for any given summary net.)

**Environment**
- OS: Ubuntu
- Python Version: 3.11
- Backend: all
- BayesFlow Version: 2.0.3

han-ol:
Fixes #468.

The PR
* adds a function `positive_diag()` in the utils module that converts a batch of matrices to the same batch of matrices but with positive diagonal. In particular, passing matrices filled with zeros will produce unit matrices.
* adds a `keras.layers.LayerNormalization` to the matrix elements to prevent high values for these, as can occur in the first forward pass after initialization. Center and scale are initialzed as 0 and 0.1, but are learnable parameters.

Together this unchanged expressivity and initialization of covariance matrices near the unit matrix.

### DIFF
--- a/bayesflow/links/positive_definite.py
+++ b/bayesflow/links/positive_definite.py
@@ -14,8 +14,6 @@ class PositiveDefinite(keras.Layer):
 
         self.layer_norm = keras.layers.LayerNormalization()
 
-        self.built = True
-
     def call(self, inputs: Tensor) -> Tensor:
         # normalize the activation at initialization time mean = 0.0, std = 0.1
         inputs = self.layer_norm(inputs) / 10
--- a/bayesflow/links/positive_definite.py
+++ b/bayesflow/links/positive_definite.py
@@ -1,7 +1,7 @@
 import keras
 
 from bayesflow.types import Tensor
-from bayesflow.utils import layer_kwargs, fill_triangular_matrix
+from bayesflow.utils import layer_kwargs, fill_triangular_matrix, positive_diag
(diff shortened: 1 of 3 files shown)


stefanradev93:
Thanks for the improvements, Hans!

PR #471, opened 2025-05-09
Fixed Jacobian computation of standardize transform

han-ol:
I believe the Jacobian computation of standardize should have reversed sign. 

y = (x - mu) / sigma
log p(y) = log p(x) - log(sigma)

Calling the adapter in forward direction should substract log(sigma), while inverse direction should add log(sigma).

Current implementation is switched. The PR fixes that.

BTW: It's really nice that Jacobian tracking adapter exists now in BF, thanks so much!


Kucharssim (approved):
Oops, you are right. Thanks for the fix!

paul-buerkner:
thank you both! feel free to merge :-)

Simon Kucharsky ***@***.***> schrieb am Fr., 9. Mai 2025,
16:42:

Issue #472, opened 2025-05-12
Implement Fusion Summary Networks


vpratz:
Multimodal inputs occur in many settings. They require a more complex structure of `summary_variables` than numpy arrays, as different inputs usually have different dimensionality, or even number of dimensions. They might also necessitate different summary networks, which are tailored to the specific summary. Manual implementation of this is already possible with the current API, but streamlining it would be great for downstream tasks. To achieve this, we need

- a transform to introduce the additional nesting in the adapter: (e.g., `fuse(keys, into="summary_variables")` which creates a dict with the provided keys and corresponding values inside the `summary_variables` entry).
- a `FusionSummaryNetwork` which takes a `dict` of keys and corresponding summary networks, as well as an additional network that combines their outputs to a chosen summary dimension.

Open questions:
- Do we want to allow for transforms in the second level dict, for example by adding an `adapter` argument to `fuse`, which operates on the inner variables?

stefanradev93:
Addressed in https://github.com/bayesflow-org/bayesflow/pull/473

PR #473, opened 2025-05-12
Add classes and transforms to simplify multimodal training

vpratz:
This PR addresses #472.

- Add class `MultimodalSummaryNetwork` to combine multiple summary networks, each for one modality.
- Add transforms `Group` and `Ungroup`, to gather the multimodal inputs in one variable (usually "summary_variables")
- Add tests for new behavior

Open questions:
- Handling of nesting in transforms: As far as I can tell, most transforms cannot handle nested data structures, as they are not designed for it. Changing this is a major undertaking, and probably something we do not want to do now. The multimodal inputs require passing a dict as "summary_variables". To facilitate this, I added the `Group` transform, which takes variables from the shallow level and moves them to a `dict`. I also added `Ungroup` to do the inverse, so that in combination one additional level of depth is available in the adapter. I think this is sufficient for now, but I'm open for other design choices/suggestions.
- Naming: I opted for more generic names, also because we already have a `FusionTransformer`. I have named the final network in the MultimodalSummaryNetwork `fusion_network`, maybe that might lead to confusion as well? If so, do you have ideas for a more generic name?

When we have agreed on an implementation, I will add a small tutorial to showcase how all of this can be used. If you want to see an example before this, just let me know...


stefanradev93:
Great idea!

vpratz:
I have added a notebook to demonstrate how one would use this. What do you think?

stefanradev93:
Hi Valentin, I really like the approach and the added `group` transform allowing for dicts as values for the ultimate keys. This can be used for other custom summary networks as well! A few minor remarks before I do more in-depth testing:

- I believe each network `[self.summary_networks[k](inputs[k]) for k in self._ordered_keys]` should also get the `training` argument.
- I would call the new module `fusion_network` and corresponding network `FusionNetwork`, stating in the doc that the net implements **late fusion**. The modules inside the network may be called `backbones` and `head`. Just a thought.

stefanradev93:
Early fusion would be mixing the raw data somehow (e.g., cross attention) and then having a simple aggregation in the end.

PR #474, opened 2025-05-13
Permit Python version 3.12

vpratz:
Add both versions to test temporarily to test the installation in the CI.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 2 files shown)


vpratz:
Python 3.13 is not supported by TensorFlow yet (see [here](https://www.tensorflow.org/install/pip)), but probably it will be soon. Python 3.12 seems to work with our tests, so I think this is ready to be merged.

Somewhat related: I noticed that version 1 does not have an upper bound on the Python version, so on all newer Python version not supported by BayesFlow 2 yet, version 1 will be installed instead of showing an error. Can/do we want to change that somehow?

PR #482, opened 2025-05-20
Fix #481

### LINKED ISSUE #481: BasicWorkflow: Unstable val_losses due to validation data shuffling
In the `BasicWorkflow` class, the validation set is constructed via `validation_data = OfflineDataset(data=validation_data, batch_size=dataset.batch_size, adapter=self.adapter)` [in the `_fit` method](https://github.com/bayesflow-org/bayesflow/blob/main/bayesflow/workflows/basic_workflow.py#L882). This treats the validation data as a standard `OfflineDataset` and implies shuffling on epoch end. Thus, stochasticity is introduced by modifying the composition of the evaluated batches, which can lead to substantial changes between multiple validation loss calculations during training even for frozen networks. 

E.g., adapting Linear_Regression_Starter.ipynb to use offline training with a learning rate of 0 (and no standardization in the adapter),

```python
validation_data = workflow.simulate(1000)
history = workflow.fit_offline(
    data=training_data,
    epochs=5,
    batch_size=32,
    validation_data=validation_data,
)
```

leads to highly variable validation losses:

```python
[19.465839385986328,
 10.662302017211914,
 16.612285614013672,
 7.247644901275635,
 17.282154083251953]
```

Setting the number of validation sets equal the batch size, either via `validation_data = workflow.simulate(32)` or via `batch_size=1000`, enforces equal batches despite shuffling, resulting in stable validation losses:

```python
[12.364879608154297,
 12.364879608154297,
 12.364879608154297,
 12.364879608154297,
 12.364879608154297]
```

A pull request with a simple fix will follow.

elseml:
Added the option for OfflineDataset to prevent shuffling on epoch end, leading to stable validation losses also when validation_set_size != batch_size (see #481). Not sure if we want explicit tests for this since it would involve quite some code?


stefanradev93:
The change should be covered by the workflow tests.

elseml:
Sure! I liked that the previous solution reused the existing `stage` argument, but this way it is more explicit. I adapted the code using `shuffle_on_epoch_end` instead since shuffling is also applied on dataset creation, which does not relate to the current issue (we could also disable both via a general `shuffle` argument, but I lean towards retaining initial shuffling also for validation sets).

While this fixes validation set shuffling for the workflow, users working with a manual training pipeline now always need to remember to set `shuffle_on_epoch_end=False` when creating `OfflineDataset` validation sets to prevent shuffling during training. We could also delete shuffling by default altogether and let users always specify the argument, but this would not be too elegant. Any ideas for a more principled solution that alleviates this requirement (also pinging @LarsKue)?

vpratz:
Why do you lean towards retaining initial shuffling also for validation sets? My impression would be that we usually aggregate over the validation batches anyway, and computing them should not modify anything, so the order should not matter. Do you have cases in mind where the order of the validation set would matter? If not, I think distinguishing `shuffle` vs. `shuffle_on_epoch_end` is not necessary for now, and going with `shuffle` is sufficient. What do you think?

vpratz:
Are you sure? I think I have seen incomplete batches in the past. At least the datasets produce incomplete batches:
```python
ds = bf.OfflineDataset(data={"data": np.random.normal(size=(66,1))}, batch_size=32, adapter=bf.Adapter())

for i, d in enumerate(ds):
    print(i, d['data'].shape)
# Output:
0 (32, 1)
1 (32, 1)
2 (2, 1)
```

stefanradev93:
Yes, I believe the internal `evaluate` function has `drop_last` or something of sorts. @LarsKue

elseml:
I had the same impression, but might be mislead here. Curious to hear Lars' opinion on this.

LarsKue:
It appears to, but I have been unable to find it in the code. Could be something else going on, like an off-by-one error.

elseml:
Since we could not find evidence that initial shuffling is required for validation/test sets, I followed Valentins suggestion to simplify the interface and control both initial shuffling and shuffling on epoch end with a single argument (using a `shuffle_dataset` argument since there is already a class method `shuffle`). 

I also added equivalent changes to the `DiskDataset` class and disabled validation set shuffling in the two (experimental) example notebooks that currently manually create `OfflineDataset` validation sets so that we don't forget about it. Everything should be ready for merge now from my side.

vpratz:
In addition, I would be in favor of renaming the `shuffle_dataset` parameter to `shuffle`, to be consistent with e.g. [torch Dataloaders](https://docs.pytorch.org/tutorials/beginner/basics/data_tutorial.html). We can store it as `_shuffle` to avoid a collision with the method.

PR #485, opened 2025-05-22
Correctly track train / validation losses

### LINKED ISSUE #481: BasicWorkflow: Unstable val_losses due to validation data shuffling

LarsKue:
This PR:

- Fixes #481 
- Supercedes #483 
- Fixes validation loss tracking in `test_step`
- Removes `loss/..._loss` and other nested metrics
- Temporarily removes the ability to track custom metrics

Future TODO:
- We have to reenable tracking custom metrics by keeping a `keras.metrics.Mean` tracker object on the approximator for each custom metric the user passes. We also need to call `update_state` on each of those trackers in the `compute_metrics` or `train/test_step` method.

@vpratz Do you think you have capacity to take care of the TODO?

### DIFF
--- a/bayesflow/approximators/approximator.py
+++ b/bayesflow/approximators/approximator.py
@@ -142,7 +142,8 @@ class Approximator(BackendApproximator):
         """Obtain the batch size from a batch of data.
 
         To properly weight the metrics for batches of different sizes, the batch size of a given batch of data is
-        required. As the data structure differs between approximators, each approximator has to specify this method.
+        required. As the data structure differs between approximators, each concrete approximator has to specify
+        this method.
 
         Parameters
         ----------
--- a/bayesflow/approximators/approximator.py
+++ b/bayesflow/approximators/approximator.py
@@ -1,5 +1,3 @@
-from collections.abc import Mapping
-
 import multiprocessing as mp
 
 import keras
@@ -22,7 +20,7 @@ class Approximator(BackendApproximator):
         # implemented by each respective architecture
         raise NotImplementedError
 
-    def build_from_data(self, data: Mapping[str, any]) -> None:
+    def build_from_data(self, data: dict[str, any]) -> None:
         self.compute_metrics(**filter_kwargs(data, self.compute_metrics), stage="training")
         self.built = True
 
@@ -137,27 +135,3 @@ class Approximator(BackendApproximator):
             self.build_from_data(mock_data)
 
         return super().fit(dataset=dataset, **kwargs)
-
-    def _batch_size_from_data(self, data: any):
-        """Obtain the batch size from a batch of data.
-
-        To properly weight the metrics for batches of different sizes, the batch size of a given batch of data is
-        required. As the data structure differs between approximators, each concrete approximator has to specify
-        this method.
-
-        Parameters
-        ----------
-        data :
-            The data that are passed to `compute_metrics` as keyword arguments.
-
-        Returns
-        -------
-        batch_size : int
-            The batch size of the given data.
-        """
-        raise NotImplementedError(
-            "Correct calculation of the metrics requires obtaining the batch size from the supplied data "
-            "for proper weighting of metrics for batches with different sizes. Please implement the "
-            "_batch_size_from_data method for your approximator. For a given batch of data, it should "
-            "return the corresponding batch size."
-        )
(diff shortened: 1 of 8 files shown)


vpratz:
Do I read it correctly that this PR does not modify the shuffling and therefore not fix this aspect of #481?

Regarding the TODO: I'm not familiar with this part of the code base yet, so I cannot really judge how complicated it is and when I will find the time, but I can try to a closer look at it in the coming weeks.

stefanradev93:
Shall we still merge this PR, as it provides a pretty important fix and open another one pertaining to general metrics and aggregation.

stefanradev93:
It fixes the incorrect calculation and display of validation losses only on the last validation batch. Additionally, it removes the duplicate printing of loss, inference_net/loss, val_loss / val_inference_net/loss.

vpratz:
Ahh, thanks a lot! I will do a proper review in the next days, then.

vpratz:
I now understand better what is going on, and the changes look good to me. I will try to add the trackers for the other metrics today, and let you know if I encounter any difficulties...

vpratz:
Quick question @LarsKue @stefanradev93 : The individual losses were removed in the PR. I would add them as metrics in the case that more than one loss is present (not if there is only one, to avoid the useless duplicate), so that they are individually tracked. This will also display them, which I think is desirable. Do you agree, or would you rather not display that information?

vpratz:
Short update: I have succeeded in tracking the metrics, but the serialization for custom `inference_metrics` and `summary_metrics` seems to be broken (at least in TensorFlow), and might need some restructuring to work nicely. It might make sense to move this to a separate PR though, I'm currently checking if the issues are separate or interdependent...

vpratz:
@LarsKue @stefanradev93 I think the changes are ready to review, the issues regarding custom metrics are (as far as I can tell) not related to the changes in this PR.

vpratz:
I was able to resolve this properly. The reason for the order-dependency was taking the unweighted mean of means. So in the averaged metrics, average values obtained from batches with batch size had the same weight as values obtained from batch size 2.
By setting the `sample_weight` argument of `Metric.update_state` to the batch size, we can obtain the correct averages. This requires calculating the batch sizes from the data at hand, which is approximator-specific. Therefore I outsourced this calculation into a private method that each approximator has to overwrite. If you have different design-ideas for this, please let me know.

This fixes #481 and supersedes that aspect of #482.
@LarsKue @stefanradev93 Could you please take another look?

Note: `tqdm` will report wrong values for the training loss, as it does naive sampling, the value stored in `history` will be correct, though.

<details><summary>Code to reproduce</summary>

```python
import bayesflow as bf
import keras

workflow = bf.BasicWorkflow(
    inference_network=bf.networks.CouplingFlow(subnet_kwargs={"dropout": 0.0}),
    inference_variables=["parameters"],
    inference_conditions=["observables"],
    simulator=bf.simulators.GaussianMixture(),
    initial_learning_rate=0.0,
    standardize=[]
)

training_data = workflow.simulate(66)
validation_data = workflow.simulate(66)
history = workflow.fit_offline(
    data=training_data,
    epochs=3,
    batch_size=32,
    validation_data=validation_data,
)
print(history.history)
```

</details>

stefanradev93 (commented):
Hi Valentin and Lars, I buffed up the docs for the new changes. I also decided to repeat `_batch_size_from_data` in the backend-specific classes to avoid confusion, since the method is already explicitly needed in these classes. Other than that, I think this PR is ready to merge.

Issue #484, opened 2025-05-22
[BUG] Data sets have a dangerous stage argument


stefanradev93:
**Describe the bug**
Our data sets currently accept a stage argument which is `training` by default. Having `validation` and a stateful `standardize` transform will results in an error due to:

`if self.adapter is not None:
    batch = self.adapter(batch, stage=self.stage)
`

since the running means and standard deviations have never been computed. 

**Expected behavior**
I have come to the realization that `BatchNorm` layers should be part of the approximators and applied to all `inference_conditions`, `summary_variables`, and `inference_variables`. This will have the advantage that adapters will remain stateless and users will not have to deal with standardizing things explicitly. Still, we should keep the `standardize` transform with static means and stds for special cases. 

Let me know what you think and I will provide an implementation.

stefanradev93:
The question is only how to handle the ldj of the transformation.

vpratz:
I think "standard" batch norm is not well suited for this, as during training it uses the _current_ mean and std of a batch instead of the moving average, which is not what we want and will probably hurt performance. We could think about moving the standardization with momentum (as it is currently implemented in the adapter) to the approximator and standardize `summary_conditions`, `inference_variables` and (possibly) the outputs of the summary network.

A few other thoughts:
- we would usually not want to use different adapters for training set and validation set, but use the same adapter. Using different adapters will lead to different standardization, so different inputs to the network. The issue you describe then only arises if we use it on the validation set before the training, which might or might not make sense, depending on the context.
- we could in addition modify the behavior of the `Standardize` transform, so that it uses mean and variance of the batch in cases where it has momentum and was not called with stage="training" yet (maybe with a warning)

I see the problem though, and resolving it would be great. Another avenue we might explore is moving the adapter out of the datasets, e.g. by wrapping the datasets in the approximator. If I recall it correctly, @LarsKue has reflected on the general problem and possible difficulties with it, but I don't remember exactly which difficulties led to putting the adapter into the datsets in the first place. @LarsKue, could you comment on this?

LarsKue:
If the adapter is in the dataset, calling it is "free" in terms of computational cost, since it will be done by a worker process. That is the primary reason why it is there.

vpratz:
Thanks for the comment! Do you know whether the computational cost of the adapter is practically relevant here? I would assume the adapter to be cheap compared to the summary/inference networks in most cases. Did you do any benchmarking regarding this?

stefanradev93:
[...]

PR #486, opened 2025-05-23
[WIP] Move standardization into approximators and make adapter stateless.

stefanradev93:
Creates a `Standardization` layer that is adaptively created and applied to all relevant variables (`summary_variables`, `inference_variables`, `inference_conditions`), along with a few unit tests. 

The layer knows how to compute its inverse and ldj and remains invisible to users. As a consequence, users no longer need to standardize quantity explicitly in the adapter and the adapter can be stateless. Thus, the PR addresses https://github.com/bayesflow-org/bayesflow/issues/484. 

TODOs:
- Add deprecation warning for using standardize in the Adapter.
- Add layer to `ModelCaomparisonApproximator`
- What to do with point estimation networks? @han-ol 
- Adjust all notebooks, removing standardize transforms from the adapter (SIR and LR updated).
- Rigorously ensure `log_prob` remains correct.


bayesflow/approximators/continuous_approximator.py
```diff
@@ -112,6 +113,33 @@ def compile(
...
+        elif isinstance(self.standardize, str):
+            keys = [self.standardize]
+        elif isinstance(self.standardize, Sequence):
+            keys = self.standardize
+        else:
+            keys = []
+
+        if "inference_variables" in adapted_data and "inference_variables" in keys:
+            self.inference_variables_norm = Standardization()
+            self.inference_variables_norm(adapted_data["inference_variables"], stage="training")
```
vpratz, line 129: Can those initializing calls be skipped, as they are called by the initial the self.compute_metrics call (l.140) anyways?

stefanradev93: Good spot, the calls can be skipped!

stefanradev93:
Yes, I wouldn't remove the `standardize` transform from the adapter, only deprecate the stateful version and phase it out as we roll on.

paul-buerkner:
So about the point estimator layer: For safety reasons it could make sense to deactivate the standardization layer by default. However, that may cause a lot of issues with users complaining about unstable networks. 

So what actually breaks? If users call `.sample` everything is fine anyway. If users call `.estimate`, mean and quantiles are just fine. SDs, variances and covariances can "made to be fine" if we add code that knows how unstandarized these point estimates. One could add a warning if standardization was used but `.estimate` is called with a point estimate whose unstandardization behavior is not defined?

stefanradev93:
Yes!

vpratz:
I have added an end-to-end test containing nested inputs which will fail for the current state, the fix is contained in #501.

bayesflow/networks/standardization/standardization.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @classmethod
+    def from_config(cls, config, custom_objects=None):
+        return cls(**deserialize(config, custom_objects=custom_objects))
+
+    def call(
+        self,
+        x: Tensor | dict[str, Tensor],
+        stage: str = "inference",
+        forward: bool = True,
```
vpratz, line 62: Do we want to change `forward` to `inverse`, in line with the convention in the adapter?

stefanradev93: I really don't like the Adapter's convention, which follows PPL conventions rather than DL. Our networks call `forward` during training and `inverse` during inference, so I think we should stick with that, especially now that standardization is part of the networks.

vpratz: Ok, that's fine then, thanks for the comment :)

stefanradev93:
The branch turned out to be a medium refactor of the approximators. I will probably open another one regarding point approximators / point networks. The only remaining part is how to generally and elegantly handle un-standardization of the estimates when `estimate` is called. Right now, the estimates would be valid in standardized space. I would appreciate some help on that @han-ol.

bayesflow/adapters/transforms/standardize.py
```diff
@@ -69,6 +70,14 @@ def __init__(
...
         super().__init__()
 
+        if mean is None or std is None:
+            warnings.warn(
+                "Dynamic standardization is deprecated and will be removed in later versions."
+                "Instead, use the standardize argument of the approximator / workflow instance or provide "
+                "fixed mean and std arguments. You may incur some redundant computations if you keep this transform.",
+                DeprecationWarning,
+            )
+
```
vpratz, line 80: I think it would be nice to have a convenience function that calculates mean and std for a dataset, in the format that would be required here. We could also advertise it in the deprecation warning. What do you think?

bayesflow/approximators/approximator.py
```diff
@@ -11,17 +11,17 @@
 
 
 class Approximator(BackendApproximator):
-    def build(self, data_shapes: any) -> None:
-        mock_data = keras.tree.map_structure(keras.ops.zeros, data_shapes)
+    def build(self, data_shapes: dict[str, tuple[int]]) -> None:
```
vpratz: Our type hints currently do not reflect nested structures. Do we leave it like that for now or do we want to adapt them? What would be a good hint in that case, just `dict[str, tuple[int] | dict[str, dict]]`, or something more involved?

bayesflow/approximators/approximator.py
```diff
@@ -11,17 +11,17 @@
...
 
     @classmethod
     def build_adapter(cls, **kwargs) -> Adapter:
         # implemented by each respective architecture
         raise NotImplementedError
 
-    def build_from_data(self, data: dict[str, any]) -> None:
-        self.compute_metrics(**filter_kwargs(data, self.compute_metrics), stage="training")
+    def build_from_data(self, adapted_data: dict[str, any]) -> None:
+        self.compute_metrics(**filter_kwargs(adapted_data, self.compute_metrics), stage="training")
```
vpratz: This call can error (with a very ugly/uninformative error message) in Jax if the inference and summary networks do not check themselves whether they are already built (see 01aadf108207f55705b376be8e760633eddddb01 for a fix where this was the problem). We could think about using a `NotImplementedError` here and trying to do a proper build in the subclasses, but I'm not sure if it is worth it right now.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -53,9 +55,22 @@ def __init__(
...
+        if isinstance(standardize, str) and standardize != "all":
+            self.standardize = [standardize]
+        else:
+            self.standardize = standardize
+
+        if self.standardize == "all":
+            # we have to lazily initialize these
+            self.standardize_layers = None
+        else:
+            self.standardize_layers = {var: Standardization() for var in self.standardize}
```
vpratz: In the other constructor calls it is `Standardization(trainable=False)`, is the trainable argument missing here?

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -284,7 +323,7 @@ def predict(
         self,
         *,
         conditions: Mapping[str, np.ndarray],
-        logits: bool = False,
+        probs: bool = True,
```
vpratz, line 359: I think this rename should not happen silently. We could accept `logits` and use `prob = not logits` while showing a deprecation warning for a while.

han-ol:
Specification of what I think we want now, regarding point estimates and their transformations (adapter and in-approximator-standardization): 
1. support everything for of anything sample-like (incl. simple point estimates like mean, quantiles)
2. warn when parametric estimate (potentially) cannot be inverse-adapted
3. NEW since this PR: deal with in-approximator-standardization of non-sample-like estimates (e.g. covariance)

current transforms that have trouble with inverse for non-sample-like estimates
* constrain
* log
* sqrt
* scale
* shift
* standardize
If any of those are  applied to data that ends up in inference variables, back-transforming estimates like covariance is not supported currently. On the other hand, if we refrain to use those on inference variables, even non-sample-like estimates should be correctly transformed.

Proposed solution:
We can add a class attribute similar to `NOT_TRANSFORMING_LIKE_VECTOR_WARNING`. I'd suggest `TRANSFORMING_LIKE_TENSOR`.

(We can keep `NOT_TRANSFORMING_LIKE_VECTOR_WARNING` in anticipation of some estimate that neither transforms like a vector nor a tensor.)

 It should be no problem to check `TRANSFORMING_LIKE_TENSOR`, and correspondingly apply the tensor transformation of the standardization. That is:
 estimate * std ** 2.

For implementing this with the `Standardization` layer, one option would be to add a flag `tensor: bool = False` does this if True is passed, but you might have a different idea also.

bayesflow/approximators/point_approximator.py
```diff
@@ -148,17 +152,18 @@ def log_prob(
 
             Log-probabilities have shape (num_datasets,).
         """
-        log_prob = super().log_prob(data=data, **kwargs)
-        # Squeeze log probabilities dictionary if there's only one key-value pair.
-        log_prob = PointApproximator._squeeze_parametric_score_major_dict(log_prob)
-
-        return log_prob
+        return super().log_prob(data=data, **kwargs)
```
vpratz: Could you quickly comment why this call to `_squeeze_parametric_score_major_dict` was removed here? It is still present in the `sample` function.

bayesflow/approximators/point_approximator.py
```diff
@@ -233,7 +238,7 @@ def _squeeze_estimates(
     def _squeeze_parametric_score_major_dict(samples: Mapping[str, np.ndarray]) -> np.ndarray or dict[str, np.ndarray]:
         """Squeezes the dictionary to just the value if there is only one key-value pair."""
         if len(samples) == 1:
-            return next(iter(samples.values()))  # Extract and return the only item's value
+            return next(iter(samples.values()))
```
vpratz, line 262: Can we keep this comment, I think it is helpful and there is no need to remove it

bayesflow/approximators/continuous_approximator.py
```diff
@@ -32,42 +33,51 @@ class ContinuousApproximator(Approximator):
...
-    SAMPLE_KEYS = ["summary_variables", "inference_conditions"]
+    CONDITION_KEYS = ["summary_variables", "inference_conditions"]
 
     def __init__(
         self,
         *,
         adapter: Adapter,
         inference_network: InferenceNetwork,
         summary_network: SummaryNetwork = None,
+        standardize: str | Sequence[str] | None = "all",
```
vpratz, line 55: Do we want to expose `momentum` and `epsilon` for the standardization to users?

vpratz (changes requested):
I think the changes look good overall, see my individual comments for possible changes.

bayesflow/approximators/point_approximator.py
stefanradev93: Because our `log_prob` functions currently only return Tensors. This may change for graphical approximators, but it will require a different design altogether.

bayesflow/approximators/model_comparison_approximator.py
stefanradev93: Good spot. It's not strictly necessary, but I like making it explicit. Fixed.

bayesflow/approximators/model_comparison_approximator.py
stefanradev93: Agreed. Fixed with a deprecation warning.

bayesflow/approximators/point_approximator.py
han-ol: Point inference networks do return a dictionary for log_prob actually, one key-value pair vor every ParametricDistributionScore.

bayesflow/approximators/approximator.py
stefanradev93: The call was also present before. What is essential is the call to `build` on L139 of the `Approximator`, because keras wraps the build function internally. I think I like the idea of re-implementing build in the subclasses. @LarsKue What do you think?

bayesflow/approximators/approximator.py
stefanradev93: I would keep it simple for now as you suggested: `dict[str, tuple[int] | dict[str, dict]]`.

bayesflow/adapters/transforms/standardize.py
stefanradev93: Agreed, but such a function will not be very efficient when the entire data set is not (yet) in memory. I see its use mainly for `OfflineDataset`.

bayesflow/approximators/approximator.py
vpratz: Yes, sorry for not making it more explicit, this problem is older and not related to the changes here, we can also move this to a separate issue if you like

bayesflow/approximators/point_approximator.py
stefanradev93: Fixed. I added it as a separate line, as we agreed to refrain from inline comments.

bayesflow/approximators/approximator.py
stefanradev93: My biggest issue with this right now is that the build_from data runs the networks with a bunch of zeros, which is bad for initializing the running means and stds.

bayesflow/approximators/continuous_approximator.py
stefanradev93: Yes, I would add them to the kwargs for approximators and workflow. Agreed?

bayesflow/approximators/point_approximator.py
vpratz: Thanks. Is there a place to read up on those style agreements? If not, do we want to put them in the `CONTRIBUTING` file?

bayesflow/approximators/approximator.py
stefanradev93: I would probably address it here. This turned out to be one of those monstrous PRs which start with a harmless feature and end up in a gargantuan refactor.

bayesflow/approximators/point_approximator.py
stefanradev93: The definition of `log_prob` in the `PointApproximator` is:

`return super().log_prob(data=data, **kwargs)`

This calls `ContinuousApproximator`'s `log_prob` method, which returns a tensor.

stefanradev93:
@han-ol We can rarely refrain from using those transformations on inference variables, since they not only help convergence but also avoid exploding gradients. I am on board with the idea of adding a flag to the estimators, such that the approximator can handle the transformation accrodingly.

bayesflow/approximators/approximator.py
LarsKue: We need to replace the build method entirely if we want to reliably build on real data.

bayesflow/approximators/point_approximator.py
LarsKue: Eventually, we should make a style guide. We could also add some rules to ruff, but I think this could be too strict in this case.

LarsKue:
@han-ol Could you explain what the idea of separating vectors from tensors is here? Everything should transform like a tensor, including vectors.

bayesflow/approximators/point_approximator.py
han-ol: The problem I mean arises, because the log_prob originates at `PointInferenceNetwork.log_prob()`, which returns a dictionary. In this case we need a dictionary, because different `ParametricDistributionScore` s might imply different log_probs. This was the reason for the more complicated map_structure calls in `ContinuousApproximator.log_prob` that were also removed  in the PR.

I see the following options to keep this functionality intact:
1. fully override log_prob in `PointApproximator` without the `super().log_prob` call
2. restore the previous pattern allowing for log_prob to be a dictionary in `ContinuousApproximator.log_prob` which should only require minor adaptations similar to what we had before.

More generally, we could aim for a cleaner separation of what to inherit by introducing an intermediate parent class shared for `PointApproximator` and `ContinuousApproximator`. However, that would be a bigger change.

han-ol:
@stefanradev93 Ok and yes, I agree. I just wanted to reiterate, what is currently supported and what is not, but we should not focus on the adapter in this PR. Would you pass the information on how a particular estimate transforms to a extended `Standardization.call` method or implement it differently?

stefanradev93:
To me this is a case where inheritance does more harm than good, so it may
make more sense to override the method in the point approximator and
completely decouple it from continuous in the near futurs.

han-ol:
@LarsKue Thanks for bringing this up, you're right. Vectors are rank-1 tensors, so we should choose different names.The difference I am actually talking about is of cause in the transformation behavior: ($D = \text{diag}[\sigma_1, ... \sigma_n]$)

1. sample, quantile, ...: transforms as $D \cdot x' + \mu$
2. covariance: transforms as $D \cdot \Sigma' \cdot D$

I thought of it as vector vs tensor because (1.) is also applying the shift, so it is not a linear transformation anymore but an affine one.

Thinking and reading a bit about this again, I also realized only now, that we should not think of standardization as a basis change, but rather a data change. We'd need to multiply by $D^{-1}$ from both sides for (2.) since the covariance is covariant if we were following the basis change rules.

I would propose the following class attributes to inform inverse standardization.  

```
# in base class ScoringRule
no_shift:tuple = tuple()   # default is empty
rank = {}  # empty, if not given would fallback to rank 1 for convenience

# in MeanScore instead
rank = dict(value=1)  # same as default but explicit

# in MultivariateNormalScore instead
no_shift = ("covariance",)
rank = dict(mean=1, covariance=2)
```

vpratz:
I have added the test case for the log_prob call with a point inference network with multiple parametric scores (i.e., where log_prob is a dictionary).

stefanradev93:
> I have added the test case for the log_prob call with a point inference network with multiple parametric scores (i.e., where log_prob is a dictionary).

Great! @han-ol can you add the new method to the point approximator?

bayesflow/approximators/continuous_approximator.py
```diff
@@ -154,12 +194,37 @@ def compute_metrics(
...
+            raise ValueError("Summary variables are required when a summary network is present.")
+
+        if "summary_variables" in self.standardize:
+            summary_variables = self.standardize_layers["summary_variables"](summary_variables, stage=stage)
+
+        summary_metrics = self.summary_network.compute_metrics(summary_variables, stage=stage)
+        summary_outputs = summary_metrics.pop("outputs")
+        return summary_metrics, summary_outputs
+
+    def _prepare_inference_variables(self, inference_variables: Tensor, stage: str) -> Tensor:
```
han-ol, line 256: `_prepare_inference_variables` is not explicitly transforming inputs to tensor.

han-ol:
Thanks for adding the tests! I added an independent `log_prob` method for the `PointApproximator` and the new tests are passing.

I also refactored the preparation of inference_variables and conditions in general. A new method `ContinuousApproximator._prepare_data` does the job for `sample`, `log_prob` and `estimate` for both `ContinuousApproximator` and `PointApproximator`. I think this is a good solution to have code sharing of this central feature but still leave room for all approximators to be different e.g. in what they output.

@stefanradev93, and everybody: Could you take a look at the change and help think if there might be any side effects from this refactor?

bayesflow/approximators/continuous_approximator.py
stefanradev93: I don't think that this explicit transform is needed anymore, as keras models seem to do an implicit conversion; worth double-checking.

han-ol:
This PR seems to have introduced a problem that manifests as exploding loss during training in the linear regression notebook but not in the SIR or Lotka-Volterra notebooks. I haven't determined the commit at which the problem begins, but it occurred before https://github.com/bayesflow-org/bayesflow/pull/486/commits/4fa1bbbd0fe40437c99f91573b156589113b99be, and the dev branch is not affected.

vpratz:
> This PR seems to have introduced a problem that manifests as exploding loss during training in the linear regression notebook but not in the SIR or Lotka-Volterra notebooks. I haven't determined the commit at which the problem begins, but it occurred before [4fa1bbb](https://github.com/bayesflow-org/bayesflow/commit/4fa1bbbd0fe40437c99f91573b156589113b99be), and the dev branch is not affected.

Thanks! Do you have capacities to track this down (maybe using git bisect)?

han-ol:
I am looking into it right now and learning about git bisect :)

han-ol:
The first commit in this PR for which the notebook is not throwing an error is the fifth commit (45ab9ea0cebdbc5436613c5ba97c0f20d5415cdb), which already has the issue.
So we know it is introduced in those early commits, but I didn't find the actual problem yet.

stefanradev93:
Pretty sure it's related to the "building with a bunch of zeros" problem -> if the actual variable has large numbers (like an N), standardization layers are extremely misinitialized. We should build standardization differently, perhaps by letting the approximator build the inference and summary networks separately.

vpratz:
Good point. Will this case even lead to a std of zero, so that we initialize the std with `epsilon`? Should this be the case, does initializing with standard normal instead of zero in `Approximator.build` change the behavior you see, @han-ol?

han-ol:
Yes, this makes sense! I can confirm that 
1. the std is at its clipping value 1e-6
2. fixing N resolves the issue.
3. if N varies and we change the initialization of mean to normal, the problem is not fixed.


So I think the main problem can be described as follows:

The N in linear regression is fixed for each batch and broadcasted to the shape of x. std(N) = 0 for every individual batch of N. We are clipping this to epsilon from below (defaulting to 1e-6) and therefore also the moving average is fixed at 1e-6.

Meanwhile the mean(N) will actually be the mean N across batches.

Since subsequent batches have different N, the actual variance of values that are passed as inference_condition is on the order of 10e6.

vpratz:
How can we handle such cases? Do we need an `exclude` parameter like we have in the transforms?

stefanradev93:
They need to be handled on a case-by-case basis. Ns or other shape-like
args can simply be log-transformed and perhaps the default standardization
should be inference_variables only. Also, the Standardization may need to
throw an error when 0 var is encountered, as previously suggested by
Valentin.

bayesflow/networks/standardization/standardization.py
```diff
@@ -0,0 +1,116 @@
...
+        if log_det_jac:
+            log_det_jacs = keras.tree.pack_sequence_as(x, log_det_jacs)
+            return outputs, log_det_jacs
+
+        return outputs
+
+    def _update_moments(self, x: Tensor, index: int):
+        mean = keras.ops.mean(x, axis=tuple(range(keras.ops.ndim(x) - 1)))
+        std = keras.ops.std(x, axis=tuple(range(keras.ops.ndim(x) - 1)))
+        std = keras.ops.maximum(std, self.epsilon)
```
vpratz: I'm not so happy with the epsilon solution. Ideally, this should not happen during training. If it happens anyways, I think it might be better to issue a warning (at least once), and not update the std in that case. Another option would be to go from "moving average" to "incremental calculation of dataset mean and std". This might resolve the problem more cleanly. What do you think?

vpratz: An algorithm like [this one](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Weighted_incremental_algorithm) would probably be what we would implement for that. This would also let us properly handle differences in batch size (i.e., smaller last batch in epoch not getting overly large weight).

stefanradev93: I am on it.

stefanradev93: I have changed the standardization to a more robust incremental moments calculation with no parameters.

vpratz: That looks great, thanks a lot for the implementation! I have moved the `_update_moments` call to the front, so that we use the complete information in each call. I also set the init value for M^2 to zero. As it is additive, I think we do not require an initial value of 1, right?

bayesflow/approximators/continuous_approximator.py
vpratz: Resolved by the move to method without parameters.

han-ol:
Nice work with the incremental moments! The issue in the linear regression notebook is resolved and training is stable again.

vpratz:
What defaults do we want to return when the layer is used without prior call with `stage="training"`? Do we want to handle this as a special case and use a standard deviation of 1 when `self.count == 0`? In the current state, it is mean=0, std=0.
Edit: I have modified the code to use an std of one in [0952a29](https://github.com/bayesflow-org/bayesflow/pull/486/commits/0952a29fa60032ee3b14a7547fdd8beff2894f0b).

vpratz:
I have added a few more tests, I think we are on a good way with the PR. I see mainly the following TODOs:

- add test for different transformation types (@han-ol)
- replace `Adapter.standardize` call in `BasicWorkflow` and notebooks, document how to use the standardization in the examples
- (optional for now) create convenience function to compute fixed means and stds from dataset, for use with `Adapter.standardize`

Do you see any other things that are missing? Is there more work required for the point estimates, or are the main changes already in place?

han-ol:
I added the remaining tests for the new transformation types and also gave them more readable names.

For point estimation everything is in place now. 

Furthermore, we were skipping a test for sampling with the MVN score right after building it, because in JAX the Cholesky decomposition would usually fail. Commit https://github.com/bayesflow-org/bayesflow/pull/486/commits/ef97a6c934d6e6bb5d371bb73033de1516f81683 avoids calling it altogether by having the head return the Cholesky factor directly from now on.

stefanradev93:
I have completed the TODOs and will merge tomorrow.

PR #488, opened 2025-05-23
Add Nnpe adapter class

elseml:
Introduces a new adapter class, `Nnpe`, which implements Noisy Neural Posterior Estimation (NNPE) as described in [Ward et al. (2022)](https://proceedings.neurips.cc/paper_files/paper/2022/file/db0eac6747e3631eb91095cd76065611-Paper-Conference.pdf). NNPE augments training data with additive noise from a spike-and-slab distribution (mixture of Cauchy and Normal noise), aiming to improve robustness for noisy real world data by extending the scope of the training data. While the specific noise distributions seem to be mostly motivated by their usage in the non-amortized RNPE method proposed in the same paper, it proved to be a quick-and-easy improvement for situations where one expects the presence of similarly distributed noise [in our recent benchmarks](https://arxiv.org/abs/2502.04949).

I tried to incorporate all best practices from existing adapter classes, but would be thankful for a thorough check of the implemented changes and the clarity of the provided explanations. I used a `seed` argument instead of passing an `rng` for easier serialization, but there might be a more elegant way.

### DIFF
--- a/bayesflow/adapters/transforms/nnpe.py
+++ b/bayesflow/adapters/transforms/nnpe.py
@@ -1,0 +1,61 @@
+import numpy as np
+
+from bayesflow.utils.serialization import serializable, serialize
+
+from .elementwise_transform import ElementwiseTransform
+
+
+@serializable("bayesflow.adapters")
+class Nnpe(ElementwiseTransform):
+    """Implements noisy neural posterior estimation (NNPE) as described in [1], which adds noise following a
+    spike-and-slab distribution to the training data as a mild form of data augmentation to robustify against noisy
+    real-world data (see [1, 2] for benchmarks).
+
+    [1] Ward, D., Cannon, P., Beaumont, M., Fasiolo, M., & Schmon, S. (2022). Robust neural posterior estimation and
+    statistical model criticism. Advances in Neural Information Processing Systems, 35, 33845-33859.
+    [2] Elsemüller, L., Pratz, V., von Krause, M., Voss, A., Bürkner, P. C., & Radev, S. T. (2025). Does Unsupervised
+    Domain Adaptation Improve the Robustness of Amortized Bayesian Inference? A Systematic Evaluation. arXiv preprint
+    arXiv:2502.04949.
+
+    Parameters
+    ----------
+    slab_scale : float
+        The scale of the slab (Cauchy) distribution.
+    spike_scale : float
+        The scale of the spike spike (Normal) distribution.
+    seed : int or None
+        The seed for the random number generator. If None, a random seed is used. Used instead of np.random.Generator
+        here to enable easy serialization.
+
+    Notes
+    -----
+    The spike-and-slab distribution consists of a mixture of a Cauchy (slab) and a Normal distribution (spike), which
+    are applied based on a Bernoulli random variable with p=0.5.
+
+    The default scales follow [1] and expect standardized data (e.g., via the `Standardize` adapter). It is therefore
+    recommended to adapt the scales when using unstandardized training data.
+
+    Examples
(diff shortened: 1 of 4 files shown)


bayesflow/adapters/transforms/nnpe.py
```diff
@@ -0,0 +1,97 @@
...
+        if stage != "training":
+            return data
+
+        # Check data validity
+        if not np.all(np.isfinite(data)):
+            raise ValueError("NNPE.forward: `data` contains NaN or infinite values.")
+
+        # Automatically determine scales if not provided
+        if self.spike_scale is None or self.slab_scale is None:
+            data_std = np.std(data)
```
vpratz, line 81: This will lead to different scales for each batch, I'm not sure if this is desirable. If we choose to do this, we should state it more explicitly in the docstring.

stefanradev93: There is an alternative solution:

- Standardize data with batch_mean and batch_std
- Add unscaled spikes and slabs
- Re-scale back with batch batch_mean and batch_std

elseml: Thanks for checking! Yes, this is the drawback of that solution. I still think it is preferable to more complex solutions, since the method is more about adding some noise at all rather than adding a very specific amount of noise (the default scales by Ward et al. also seem quite heuristically chosen to me). If you agree on this, I can add some more info in the docstring.

stefanradev93: The way Lasse explained to me, the approach explicitly wants that scale(original) < scale(transformed). In that case, I think fluctuations between batches are fine, as the downstream Standardize layer (which will be part of approximators) will take care of that.

elseml:
@stefanradev93: [Ward et al. (2022)](https://proceedings.neurips.cc/paper_files/paper/2022/file/db0eac6747e3631eb91095cd76065611-Paper-Conference.pdf) chose `p=0.5` in the context of the non-amortized RNPE method and ported this to NNPE, so in the context of NNPE it is simply one of many possible noise-generating mechanisms with the advantage of being tested in at least two papers so far. I would stay with the fixed value here to reduce complexity.

bayesflow/adapters/transforms/nnpe.py
elseml: I implemented it globally following the original NNPE implementation, but agree that dimensionwise scaling would be valuable in many situations and will implement it as an option. I think dimensions with zero variation are not problematic since in that case nothings breaks, there will simply be no noise added. Dimensionwise instead of global scaling will increase the variability of the std calculation between batches though.

bayesflow/adapters/transforms/nnpe.py
```diff
@@ -34,22 +44,96 @@ class NNPE(ElementwiseTransform):
...
+        else:
+            if self.per_dimension:
+                arr = np.asarray(passed, dtype=float)
+                if arr.shape != expected_shape or arr.ndim != 1:
+                    raise ValueError(f"{name}: expected array of shape {expected_shape}, got {arr.shape}")
+                return arr
+            else:
+                try:
+                    scalar = float(passed)
+                except Exception:
```
vpratz, line 133: Please use more specific exceptions here, I think ValueError and TypeError are the relevant ones.

elseml:
Thanks for the careful review! The changes added quite some complexity, since the expected shape of passed scales now changes depending on global (expects float) or dimensionwise (expects 1D array of length data.shape[-1]) noise application, and both passed and automatically determined scales should be possible. But the transform should be quite flexible now to handle unstandardized and heterogeneous input data.

Issue #475, opened 2025-05-14
Remove Deprecation Warning for subnet_kwargs


stefanradev93:
I would allow users to specify `subnet_kwargs` **indefinitely** in inference network interfaces and remove the deprecation warning which was added at some point. 

Using `subnet_kwargs` It is much more convenient when we want to change a single settings from the defaults compared to importing the default MLP, initializing it, and passing it to the network (unwieldly in many respects).

@paul-buerkner @LarsKue @vpratz Let me know what you think before I proceed with a PR.

paul-buerkner:
I would tend to agree. can you share code of the two options for direct comparison?

stefanradev93:
Option 1 (`subnet_kwargs` enabled): 

```
inference_net = bf.networks.FlowMatching(subnet_kwargs={"widths": (256, 256)})
```

Option 2 (no `subnet_kwargs`):

```
inference_net = bf.networks.FlowMatching(subnet=bf.networks.MLP(widths=(256, 256)))
```

I think we should keep both options as viable (i.e., redundancy is good).

paul-buerkner:
Does that create any relevant code overhead apart from one if else statement essentially?

stefanradev93:
It does entail an additional if statement and serializing the `subnet_kwargs`.

paul-buerkner:
Sounds okay to me. I would be in favor of keeping the argument.

vpratz:
I agree that this is a nice usability feature, and would also be in favor of keeping it. I think from a technical and interface perspective, we should aim to do the following:

- Serialize only the initialized `subnet`, and not the `subnet_kwargs`, where possible: Makes simple deserialization possible even if defaults have changed. We might have to give more thought to cases where we use the `subnet_kwargs` further downstream, I'm not sure about the exact the behavior for those cases.
- If an object instead of a type is passed, warn or error if `subnet_kwargs` is passed as well.

stefanradev93:
Currently, subnet_kwargs must be serialized only for `CouplingFlow`.

vpratz:
Closed by #489

PR #489, opened 2025-05-24
Align diffusion model with other inference networks and remove deprecation warnings

stefanradev93:
This PR suggests a small refactor of the existing diffusion model implementation to align it a bit better with other inference networks. It also removes the deprecation warning, as discussed per https://github.com/bayesflow-org/bayesflow/issues/475. 

@arrjon @vpratz Can you please ensure that

- The following defaults make sense:
1. noise_schedule: Literal["edm", "cosine"] | NoiseSchedule | type = "edm",
2. prediction_type: Literal["velocity", "noise", "F"] = "F",
3. loss_type: Literal["velocity", "noise", "F"] = "noise",

- The function `get_weights_for_snr` of `edm_noise_schedule.py` is correct (L 102).
- Stop time should indeed be 0 and not some `epsilon` (L331 in `diffusion_model.py`)

I have removed the clipping option from the constructor, as this should be something that is potentially only passed during inference, akin to the number of steps in flow matching.

### DIFF
--- a/bayesflow/experimental/diffusion_model/diffusion_model.py
+++ b/bayesflow/experimental/diffusion_model/diffusion_model.py
@@ -17,8 +17,10 @@ from bayesflow.utils import (
     logging,
     tensor_utils,
 )
-from .dispatch import find_noise_schedule
 from bayesflow.utils.serialization import serialize, deserialize, serializable
+
+from schedules.noise_schedule import NoiseSchedule
+from .dispatch import find_noise_schedule
 
 
 # disable module check, use potential module after moving from experimental
@@ -42,7 +44,7 @@ class DiffusionModel(InferenceNetwork):
     }
 
     INTEGRATE_DEFAULT_CONFIG = {
-        "method": "euler",  # or euler_maruyama
+        "method": "euler",
         "steps": 250,
     }
 
@@ -50,9 +52,12 @@ class DiffusionModel(InferenceNetwork):
         self,
         *,
         subnet: str | type = "mlp",
-        integrate_kwargs: dict[str, any] = None,
-        noise_schedule: Literal["edm", "cosine"] | dict | type = "edm",
+        noise_schedule: Literal["edm", "cosine"] | NoiseSchedule | type = "edm",
         prediction_type: Literal["velocity", "noise", "F"] = "F",
+        loss_type: Literal["velocity", "noise", "F"] = "noise",
+        subnet_kwargs: dict[str, any] = None,
+        schedule_kwargs: dict[str, any] = None,
+        integrate_kwargs: dict[str, any] = None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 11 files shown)


arrjon (approved):
All the changes look goot to me!

bayesflow/experimental/diffusion_model/schedules/edm_noise_schedule.py
```diff
@@ -102,10 +102,11 @@ def derivative_log_snr(self, log_snr_t: Tensor, training: bool) -> Tensor:
     def get_weights_for_snr(self, log_snr_t: Tensor) -> Tensor:
         """Get weights for the signal-to-noise ratio (snr) for a given log signal-to-noise ratio (lambda)."""
         # for F-prediction: w = (ops.exp(-log_snr_t) + sigma_data^2) / (ops.exp(-log_snr_t)*sigma_data^2)
```
arrjon, line 104:
```
def get_weights_for_snr(self, log_snr_t: Tensor) -> Tensor:
         """Get weights for the signal-to-noise ratio (snr) for a given log signal-to-noise ratio (lambda)."""
         # for F-prediction: w = (ops.exp(-log_snr_t) + sigma_data^2) / (ops.exp(-log_snr_t)*sigma_data^2)
```
the comment should be for F-loss

arrjon:
Thanks for your refactor!

> * The following defaults make sense:
> 
> 1. noise_schedule: Literal["edm", "cosine"] | NoiseSchedule | type = "edm",
> 2. prediction_type: Literal["velocity", "noise", "F"] = "F",
> 3. loss_type: Literal["velocity", "noise", "F"] = "noise",

@vpratz checked the defaults on two moons only so far, where EDM with F worked best. I think this might still change in the future...

> * The function `get_weights_for_snr` of `edm_noise_schedule.py` is correct (L 102).
Correct, but the comment is wrong: should say for "F-loss", the weighting depends on the loss not the predicition type

> * Stop time should indeed be 0 and not some `epsilon` (L331 in `diffusion_model.py`)
This is indeed correct, as truncation is generally handled by the noise scheduler and `epsilon` is then schedule dependend. 
 
> I have removed the clipping option from the constructor, as this should be something that is potentially only passed during inference, akin to the number of steps in flow matching.
Agreed!

> Bonus question: Is squared loss (weighted by SNR weights) always the go to or should we allow other losses (e.g., L1,...) as in flow matching?
I have not seen any other loss used so far, might be worth a try?

> * There is a lot of inconsistency is passing the `training` argument around the diffusion model and the schedule. Some functions are explicitly designed to operate only in `training=False` mode, so they shouldn't even be getting the argument, which should generally be reserved only for operations depending on NN-outputs.

bayesflow/experimental/diffusion_model/diffusion_model.py
```diff
@@ -155,26 +160,42 @@ def from_config(cls, config, custom_objects=None):
...
         elif self._prediction_type == "x":
-            x = pred
-        else:  # "score"
-            x = (z + sigma_t**2 * pred) / alpha_t
-
-        if self._clip_x is not None:
-            x = ops.clip(x, self._clip_x[0], self._clip_x[1])
-        return x
+            return pred
+        else:
```
vpratz, line 197: Maybe keep the comment that the remaining option is "score"

bayesflow/experimental/diffusion_model/diffusion_model.py
```diff
@@ -258,9 +320,13 @@ def _forward(
         training: bool = False,
         **kwargs,
     ) -> Tensor | tuple[Tensor, Tensor]:
-        integrate_kwargs = {"start_time": 0.0, "stop_time": 1.0}
-        integrate_kwargs = integrate_kwargs | self.integrate_kwargs
-        integrate_kwargs = integrate_kwargs | kwargs
+        integrate_kwargs = {
+            **self.integrate_kwargs,
+            "start_time": kwargs.pop("start_time", 0.0),
```
vpratz, line 325: This will override the `start_time` and `stop_time` specified in `self.integrate_kwargs` (if there is one), what is the idea/motivation behind this change?

bayesflow/experimental/diffusion_model/diffusion_model.py
```diff
@@ -307,9 +373,12 @@ def _inverse(
         training: bool = False,
         **kwargs,
     ) -> Tensor | tuple[Tensor, Tensor]:
-        integrate_kwargs = {"start_time": 1.0, "stop_time": 0.0}
-        integrate_kwargs = integrate_kwargs | self.integrate_kwargs
-        integrate_kwargs = integrate_kwargs | kwargs
+        integrate_kwargs = {
+            **self.integrate_kwargs,
+            "start_time": kwargs.pop("start_time", 1.0),
```
vpratz, line 378: same as above

bayesflow/experimental/diffusion_model/diffusion_model.py
```diff
@@ -42,78 +44,80 @@ class DiffusionModel(InferenceNetwork):
     }
 
     INTEGRATE_DEFAULT_CONFIG = {
-        "method": "euler",  # or euler_maruyama
-        "steps": 250,
+        "method": "euler",
+        "steps": 100,
```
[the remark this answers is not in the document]
stefanradev93: I didn't see much difference on non-two moons examples; also the default settings + 250 steps lead to very slow sampling. Is this the case for you as well?

bayesflow/experimental/diffusion_model/diffusion_model.py
vpratz: Yes, I mainly meant why you changed up the order, using the `|` operator already allowed overwriting, and the way you rewrote it overwriting using the `self.integrate_kwargs` gets disabled (I think this was on accident)

bayesflow/experimental/diffusion_model/diffusion_model.py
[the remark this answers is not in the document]
vpratz: I don't get a clear picture, but as you wrote this is probably mostly a problem in low-dimensional and complex posteriors, so I'm fine with changing the default for now.

vpratz (approved):

PR #490, opened 2025-05-24
Fix typos in Linear_Regression_Starter.ipynb

enbrown:
Fixed small typos in the linear regression example.

Sorry if the form of this correction doesn't match the contributing guidelines; there was little guidance about how to "correct" existing examples.


paul-buerkner:
Merging now :-)

Issue #491, opened 2025-05-26
[BUG] FlowMatching checkpoints fail to load when residual connections are disabled


simschaefer:
I trained a set of networks with deactivated residual conntections in the inference net (FlowMatching):

```
inference_net = bf.networks.FlowMatching(subnet_kwargs={"residual": False})

summary_net = bf.networks.SetTransformer(dropout=0.01, num_seeds=2, summary_dim=32,embed_dim=(128, 128))

workflow = bf.BasicWorkflow(
simulator=simulator,
adapter=adapter,
initial_learning_rate=0.0005,
inference_network=inference_net,
summary_network=summary_net,
checkpoint_filepath=network_dir,
checkpoint_name=network_name,
inference_variables=model_specs['simulation_settings']['param_names']
)

history = workflow.fit_online(epochs=200, num_batches_per_epoch=1000, batch_size=16, validation_data=val_data)
```

After training, ineference works fine, but if I try to load the checkpoints:
`
approximator = keras.saving.load_model(network_dir)
`
I get this error:

```

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[70], line 12
      1 workflow = bf.BasicWorkflow(
      2     simulator=simulator,
      3     adapter=adapter,
   (...)      9     inference_variables=model_specs['simulation_settings']['param_names']
     10 )
---> 12 approximator = keras.saving.load_model(network_dir)

File ~/miniforge3/envs/bf_new/lib/python3.11/site-packages/keras/src/saving/saving_api.py:189, in load_model(filepath, custom_objects, compile, safe_mode)
    186         is_keras_zip = True
    188 if is_keras_zip or is_keras_dir or is_hf:
--> 189     return saving_lib.load_model(
    190         filepath,
    191         custom_objects=custom_objects,
    192         compile=compile,
    193         safe_mode=safe_mode,
    194     )
    195 if str(filepath).endswith((".h5", ".hdf5")):
    196     return legacy_h5_format.load_model_from_hdf5(
    197         filepath, custom_objects=custom_objects, compile=compile
    198     )

File ~/miniforge3/envs/bf_new/lib/python3.11/site-packages/keras/src/saving/saving_lib.py:367, in load_model(filepath, custom_objects, compile, safe_mode)
    362     raise ValueError(
    363         "Invalid filename: expected a `.keras` extension. "
    364         f"Received: filepath={filepath}"
    365     )
    366 with open(filepath, "rb") as f:
[...]

PR #492, opened 2025-05-26
Quickfix: ensure that Dense layers in Sequential are built

vpratz:
This is a quick fix to the problem described in #491. I think the problem originates from using the Ellipsis in the input shape to `Sequential`, which `Sequential` does not seem to process correctly, which `Dense` is apparently able to do.
@LarsKue Do you think this is a good way to fix this? We could also try to avoid the use of `Sequential` entirely, and use a custom class to avoid the issue you raised in https://github.com/keras-team/keras/issues/21158.


stefanradev93:
I would probably avoid Sequential...

vpratz:
Superseded by #493

PR #503, opened 2025-06-01
Handle nesting for ConvertDType, ToArray, adapt concatenate dispatch

vpratz:
This PR generalizes the transforms from the `Approximator.create_default` method, so that they can be used with nested inputs as well. For ConvertDType and ToArray, it supplies a generalization that works on dictionaries.
Edit: For Concatenate, it adds a minor change in the Adapter.concatenate dispatch function to detect the special case of only one key being present also if it is nested in a sequence of length one.

The recursive structure of nested inputs requires some extra utility functions, which I put into `bayesflow.utils.tree`.


LarsKue:
@vpratz Isn't the concat issue already addressed by this if-else in the adapter dispatch?
https://github.com/bayesflow-org/bayesflow/blob/449a79a07b2db749771e24dad4b176a0621a79aa/bayesflow/adapters/adapter.py#L485-L486

vpratz:
Thanks for the response. I didn't see that, I encountered an error in the transform, probably because I passed a list of length one instead of a string. With that additional context, I think it's better to just add that case in the dispatch as well, and revert the changes in the transform.

bayesflow/adapters/adapter.py
```diff
@@ -482,6 +482,9 @@ def concatenate(self, keys: str | Sequence[str], *, into: str, axis: int = -1):
         axis : int, optional
             Along which axis to concatenate the keys. The last axis is used by default.
         """
+        if isinstance(keys, Sequence) and len(keys) == 1:
+            # unpack string if only one key is supplied, so that Rename is used below
+            keys = keys[0]
```
LarsKue: Nit: I think this would be more robust with some code duplication, since we ideally don't want to rely on the interaction of two unconnected if statements:

```py3
if isinstance(keys, str):
    transform = Rename(keys, to_key=into)
elif len(keys) == 1:
    transform = Rename(keys[0], to_key=into)
```

bayesflow/adapters/transforms/convert_dtype.py
```diff
@@ -32,7 +33,7 @@ def get_config(self) -> dict:
         return serialize(config)
 
     def forward(self, data: np.ndarray, **kwargs) -> np.ndarray:
-        return data.astype(self.to_dtype, copy=False)
+        return map_structure(lambda d: d.astype(self.to_dtype, copy=False), data)
```
LarsKue, line 36: This makes no sense to me. `data` is typed as `np.ndarray`, why would you need to `map_structure` here? If this transform ever receives non-arrays, something else is going wrong, perhaps in the `FilterTransform` or `MapTransform`.

bayesflow/adapters/transforms/to_array.py
```diff
@@ -35,13 +36,36 @@ def get_config(self) -> dict:
...
+            def restore_original_type(path, value):
+                try:
+                    original_type = get_value_at_path(self.original_type, path)
+                    return original_type(value)
+                except KeyError:
+                    pass
+                except TypeError:
+                    pass
+                except ValueError:
+                    # separate statements, as optree does not allow (KeyError | TypeError | ValueError)
```
LarsKue: This is likely not an issue with optree, but your syntax. The syntax to catch any one of multiple error types is:

```py3
try:
    ...
except (ValueError, RuntimeError):
    ...
```

bayesflow/adapters/transforms/to_array.py
LarsKue, line 1: I think the added complexity to these transforms is not worth it. Why do we need to support sampling nested dictionaries?

bayesflow/utils/tree.py
LarsKue, line 1: Utilities like these are very general, and need to be extensively documented and tested. I find it somewhat opaque for now what these are trying to achieve, and why this functionality isn't already present in `optree` or `keras.tree`.

LarsKue (changes requested):
Thank you for the PR. I think we should discuss these changes a bit more before we move on with an implementation. See comments.

vpratz:
Thanks a lot for the review! I'm not sure if the motivation behind the changes is totally clear yet. I'll try to describe it once more and we can exchange ideas on how (and if) we want to proceed.
The scenario I would like to improve is the following:

- a user wants to do inference on multi-modal data using a `FusionNetwork`, and in their simulator directly specifies `summary_variables` as a dictionary that will work with the `FusionNetwork`.
- The user then wants to use the `BasicWorkflow` and sets `summary_variables=["summary_variables"]`.
- This results in errors, as the transforms in the default adapter cannot handle the dictionary that was provided as `summary_variables`.

We already have ways to work around this using `group` and `ungroup`, but this requires manually specifying the adapter and knowing how to achieve this. The main motivation for the changes at hand is to make the defaults general enough that users do not encounter errors, and that the question what is and isn't possible with the `Adapter` does not get in their way if they only use the defaults.

@stefanradev93 mentioned that in the future we might see more summary networks that require dictionaries as inputs, so I figured this might become a more common scenario and worth resolving.

Most of the complexity is due to trying to keep `ToArray` invertible. If we do not consider this important, this could be simplified.
Other avenues (like adapting the creation of the default adapter) would be possible as well.

What do you think? Is this a change we want, and if so, what would be the best way to get there?

stefanradev93:
Yes, dictionary inputs and composite summary networks will be increasingly important.

bayesflow/adapters/transforms/convert_dtype.py
vpratz: See below for the setting, I have not (yet) adapted the type hints.

bayesflow/adapters/transforms/to_array.py
vpratz: Thanks, yes, I used the wrong syntax here.

LarsKue:
I understand the motivation, but I don't think the Adapter should handle the increased complexity. Instead, I think we should look moreso into the Approximator and how we can facilitate having multiple summary networks there.

vpratz:
@LarsKue This is not only about multiple summary networks, but any summary network that requires multiple inputs that do not fit in one tensor (see Stefan's comment above). Do you have any ideas regarding that?

vpratz:
I have reduced the complexity a lot by dropping the inverse in `ToArray`, and reduced the tree utility to the minimum necessary. I think with those changes the code would be manageable.
I'm not yet convinced there is a better place to handle this. Manually handling dictionaries everywhere they might come up will probably be more complex and prone to inconsistencies than supporting them here.

stefanradev93:
Yes, mappings and sequences should be handled by the adapter. Downstream networks should only consume what the adapter provides.

vpratz:
Closing this for now, as it is still not clear what kind of input to the adapter we want to support. We might reopen this if we decide to support passing dictionaries to the adapter as well.

Edit: Add reason for closing the PR for now.

Issue #430, opened 2025-04-22
`diagnostics.pairs_posterior()` plots targets behind other elements


thegialeo:
The `diagnostics.pairs_posterior()` function currently plots `targets` behind the prior and posterior distributions, making them hard to see. Additionally, there is limited customization for the appearance and legend handling of the targets, and no way to adjust their rendering order due to separate axes being used in diagonal subplots. This does not affact non-diagonal subplots.

### Minimal example to reproduce

```python
import numpy as np
import bayesflow as bf

dummy_prior = np.random.multivariate_normal([0, 0], np.eye(2), size=100)
dummy_posterior = np.random.multivariate_normal([0, 0], 0.1 * np.eye(2), size=100)
dummy_targets = np.array([0, 0])

fig = bf.diagnostics.plots.pairs_posterior(
    estimates=dummy_posterior,
    priors=dummy_prior,
    targets=dummy_targets,
)

fig.savefig("test_plots.png")
```

Output: test_plots.png

![Image](https://github.com/user-attachments/assets/acfeb308-c05f-4455-bf29-31d5123fee10)

### Suggested improvements

- Plot `targets` on top of prior and posterior plots.
- Ensure `targets` appear in both diagonal and non-diagonal subplots.
- Allow user to specify the color of the `targets`.
- Return or attach the diagonal axes (from `histplot_twinx`) so users can adjust them or add custom elements.
- Provide more control over legends, such as adding `targets` to legend.

### Root cause / possible fix

The issue seems to originate from the `histplot_twinx` function used in `_pairs_samples()`. It creates a new twin axis which is not included in the returned `g` object. As a result, `targets` and prior/posterior elements end up on different axes, preventing user control (e.g., via `zorder`). This issue does not affect non-diagonal plots since they share the same axes. 

```python
import numpy as np
import bayesflow as bf

dummy_prior = np.random.multivariate_normal([0, 0], np.eye(2), size=100)
dummy_posterior = np.random.multivariate_normal([0, 0], 0.1 * np.eye(2), size=100)
dummy_targets = np.array([0, 0])

fig = bf.diagnostics.plots.pairs_posterior(
    estimates=dummy_posterior,
    priors=dummy_prior,
)

for i in range(2):
    for j in range(2):
        ax = fig.axes[i, j]
        if i == j:
            ax.axvline(dummy_targets[i], color="red", linestyle="--", label="Target")
        else:
            ax.scatter(
                dummy_targets[j],
[...]

PR #505, opened 2025-06-03
Better pairplots

jerrymhuang:
This pull request addresses Issue #430 , where `targets` do not show up properly because they are masked by other elements in the pair plots, especially `pairs_posterior()`. 

In addition, the legends of the pair plots are refactored so that they are created separately from the plots (as opposed to the previously intertwined configuration), and `targets` are now included as part of the legends.

Example usage:

```python
import numpy as np
import bayesflow as bf

test_prior = np.random.multivariate_normal([0, 1, 2, 3], np.eye(4), size=100)
test_posterior = np.random.multivariate_normal([0, 1, 2, 3], 0.1 * np.eye(4), size=100)
test_targets = np.array([0, 1, 2, 3])

fig = bf.diagnostics.plots.pairs_posterior(
    estimates=test_posterior,
    priors=test_prior,
    targets=test_targets,
)
```

image

### DIFF
--- a/bayesflow/diagnostics/plots/pairs_posterior.py
+++ b/bayesflow/diagnostics/plots/pairs_posterior.py
@@ -21,6 +21,7 @@ def pairs_posterior(
     height: int = 3,
     post_color: str | tuple = "#132a70",
     prior_color: str | tuple = "gray",
+    target_color: str | tuple = "red",
     alpha: float = 0.9,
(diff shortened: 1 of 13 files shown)


stefanradev93:
Hi Jerry, thanks for fixing this! Please, direct the merge to `dev`, as per our convention.

PR #510, opened 2025-06-13
Release 2.0.4

stefanradev93:
## Summary  
This PR introduces the next set of improvements and qualitative changes for the upcoming **v2.0.4** release.

---

## Major Changes  
- Added several diffusion model implementations  
- Rewritten and refactored all `Approximators` classes  
- Introduced new adapter transforms, including support for nested tensors (e.g., dictionaries of tensors) to better handle multimodal simulations  
- Added a new fusion network for multimodal amortized Bayesian inference (ABI)  
- Integrated standardization layers into each `Approximation`, eliminating the need for stateful adapters  
  - **Note**: this paves the way for fully stateless adapters in **v2.0.5**; all stateful functionality from the adapter will be removed in **v2.0.5**.
- Enabled automatic standardization of point estimates  
- Enhanced support for tracking multiple loss terms  
- Enabled sampling of tensors with arbitrary shapes  
- Extended and improved tests

---

## Bug Fixes  
- Resolved serialization issues in edge cases and when using model comparison approximators  
- Fixed incorrect aggregation of validation losses  
- Corrected `pairplot` rendering issues


README.md
```diff
@@ -132,9 +99,47 @@ export KERAS_BACKEND=jax
 
 This way, you also don't have to manually set the backend every time you are starting Python to use BayesFlow.
 
-**Caution:** Some development environments (e.g., VSCode or PyCharm) can silently overwrite environment variables. If you have set your backend as an environment variable and you still get keras-related import errors when loading BayesFlow, these IDE shenanigans might be the culprit. Try setting the keras backend in your Python script via `import os; os.environ["KERAS_BACKEND"] = "<YOUR-BACKEND>"`.
```
han-ol, line 135: Is this info moved to somewhere else, or not relevant anymore for some reason?

bayesflow/datasets/online_dataset.py
```diff
@@ -18,19 +20,72 @@ def __init__(
...
+            A simulator object with a `.sample(batch_shape)` method to generate data.
+        batch_size : int
+            Number of samples per batch.
+        num_batches : int
+            Total number of batches in the dataset.
+        adapter : Adapter or None
+            Optional adapter to transform the simulated batch.
+        stage : str, default="training"
+            Current stage (e.g., "training", "validation", etc.) used by the adapter.
+        augmentations : dict of str to Callable or Callable, optional
```
han-ol: We can mention augmentation support in the release notes as well.

bayesflow/experimental/cif/cif.py
```diff
@@ -12,7 +12,7 @@
 # disable module check, use potential module after moving from experimental
 @serializable("bayesflow.networks", disable_module_check=True)
 class CIF(InferenceNetwork):
-    """Implements a continuously indexed flow (CIF) with a `CouplingFlow`
+    """(IN) Implements a continuously indexed flow (CIF) with a `CouplingFlow`
     bijection and `ConditionalGaussian` distributions p and q. Improves on
```
han-ol, line 16: I am curious what's the idea behind the convention to precede the doc string with (IN) for inference networks?

stefanradev93:
@arrjon / @vpratz The diffusion model seems to be failing the cycle consistency test with an absolute deviation of around `7.98e-03` only on JAX. Should we simply increase the tolerance for the approximate cycle consistency of DMs or is there something more fundamental at play?

bayesflow/experimental/cif/cif.py
stefanradev93: The idea was to clearly demarcate inference from summary networks until we separate them into separate modules.

README.md
stefanradev93: We haven't encountered any of these issues recently, but they may be relevant in highly dev-centric environments. I'd prefer to keep the length of the README in check.

bayesflow/datasets/online_dataset.py
stefanradev93: Good idea. Done.

arrjon:
We had problems with jax before, but only in the stochastic solver. Everything else was fine. There the problem was the random number generator. Could this be a similar issue here?

stefanradev93:
> We had problems with jax before, but only in the stochastic solver. Everything else was fine. There the problem was the random number generator. Could this be a similar issue here?

It's specifically `diffusion_model_cosine_velocity` that is accumulating more than 1e-3 error in the cycle consistency test.

vpratz:
@stefanradev93 I'm not on a desktop to verify myself, but I think the number of integration steps might be too low. For EDM+F, I had to increase it to 250 steps to get the test to pass. The other configs are markes as "slow", so the issue didn't show up in the `dev` CI.
Could you please try if increasing the number of steps in `test_networks/contest.py` resolves the issue? (Note: this is mainly a problem because the networks here are untrained, not a deeper problem with the performance of diffusion models, at least as far as I can tell)

stefanradev93:
> @stefanradev93 I'm not on a desktop to verify myself, but I think the number of integration steps might be too low. For EDM+F, I had to increase it to 250 steps to get the test to pass. The other configs are markes as "slow", so the issue didn't show up in the `dev` CI. Could you please try if increasing the number of steps in `test_networks/contest.py` resolves the issue? (Note: this is mainly a problem because the networks here are untrained, not a deeper problem with the performance of diffusion models, at least as far as I can tell)

I will do some local experiments. The trained diffusion model is performing quite well. The strange thing is that the test only fails with a JAX backend....

vpratz:
Is there a reason to not include #503 in this release, @stefanradev93? Or is it only missing for a lack of reviewing capacity? I think it would be nice to have those changes included...

vpratz:
> I will do some local experiments. The trained diffusion model is performing quite well. The strange thing is that the test only fails with a JAX backend....

@stefanradev93 I cannot reproduce the failure locally, can you? I think skipping this specific test for the diffusion models (e.g. via a marker) would be acceptable, as we do not expect cycle consistency to strictly hold for the number of steps we want to put in here, and the errors can accumulate in more or less arbitrary ways. We might want to use (at least somewhat) trained models in the future, where we can expect more consistent results.

bayesflow/adapters/transforms/random_subsample.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(
+        self,
+        sample_size: int | float,
+        axis: int = -1,
+    ):
+        super().__init__()
+        if isinstance(sample_size, float):
+            if sample_size <= 0 or sample_size >= 1:
+                ValueError("Sample size as a percentage must be a float between 0 and 1 exclusive. ")
```
LarsKue: This is missing a `raise` (see copilot review)

bayesflow/adapters/transforms/random_subsample.py
```diff
@@ -0,0 +1,48 @@
...
+        axis = self.axis
+        max_sample_size = data.shape[axis]
+
+        if isinstance(self.sample_size, int):
+            sample_size = self.sample_size
+        else:
+            sample_size = np.round(self.sample_size * max_sample_size)
+
+        # random sample without replacement
+        sample_indices = np.random.permutation(max_sample_size)[0 : sample_size - 1]
```
LarsKue: should be `[:sample_size]` (see copilot review)

LarsKue:
@vpratz I added a review to #503. I think we should not include it here, yet.

README.md
LarsKue: Bob Carpenter at Flatiron suggested that we try to hide the environment variables from users anyway, to which I strongly agree. Ideally, we should move to an implementation that can select the correct backend without having to expose users to how keras handles things.

bayesflow/adapters/transforms/standardize.py
```diff
@@ -69,6 +70,14 @@ def __init__(
     ):
         super().__init__()
 
+        if mean is None or std is None:
+            warnings.warn(
+                "Dynamic standardization is deprecated and will be removed in later versions."
+                "Instead, use the standardize argument of the approximator / workflow instance or provide "
+                "fixed mean and std arguments. You may incur some redundant computations if you keep this transform.",
+                FutureWarning,
```
LarsKue, line 78: Should be a `DeprecationWarning` since it is already deprecated.

bayesflow/datasets/__init__.py
```diff
@@ -1,4 +1,4 @@
-"""
+r"""
```
LarsKue, line 1: Did we check whether this changes anything in the generated docs?

bayesflow/adapters/transforms/nnpe.py
```diff
@@ -0,0 +1,187 @@
...
+                if arr.shape != expected_shape or arr.ndim != 1:
+                    raise ValueError(f"{name}: expected array of shape {expected_shape}, got {arr.shape}")
+                return arr
+            else:
+                try:
+                    scalar = float(passed)
+                except TypeError:
+                    raise TypeError(f"{name}: expected a scalar convertible to float, got type {type(passed).__name__}")
+                except ValueError:
+                    raise ValueError(f"{name}: expected a scalar convertible to float, got value {passed!r}")
```
LarsKue: When would this ever happen? `ValueError` implies the type is correct, but the value is wrong. All real scalar values have the correct type, and by definition have a value that is castable to float.

bayesflow/adapters/transforms/nnpe.py
```diff
@@ -0,0 +1,187 @@
...
+        self.spike_scale = spike_scale
+        self.slab_scale = slab_scale
+        self.per_dimension = per_dimension
+        self.seed = seed
+        self.rng = np.random.default_rng(seed)
+
+    def _resolve_scale(
+        self,
+        name: str,
+        passed: float | np.ndarray | None,
```
LarsKue: ideally, we would simply allow `ArrayLike` here and always cast to array

bayesflow/adapters/transforms/nnpe.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # Apply spike-and-slab noise
+        mixture_mask = self.rng.binomial(n=1, p=0.5, size=data.shape).astype(bool)
+        noise_spike = self.rng.standard_normal(size=data.shape) * spike_scale
+        noise_slab = self.rng.standard_cauchy(size=data.shape) * slab_scale
+        noise = np.where(mixture_mask, noise_slab, noise_spike)
+        return data + noise
+
+    def inverse(self, data: np.ndarray, **kwargs) -> np.ndarray:
+        """Non-invertible transform."""
```
LarsKue: Should be a comment, not a docstring.

bayesflow/adapters/transforms/squeeze.py
```diff
@@ -0,0 +1,43 @@
...
+    >>> a = np.array([[1], [2], [3]])
+
+    >>> sq = bf.adapters.transforms.Squeeze(axis=-1)
+    >>> sq.forward(a).shape
+    (3,)
+
+    It is recommended to precede this transform with a :class:`~bayesflow.adapters.transforms.ToArray` transform.
+    """
+
+    def __init__(self, *, axis: int | tuple):
```
LarsKue: `axis` should be type-hinted as `int | Sequence[int]`. We can convert to tuple internally.

bayesflow/adapters/transforms/standardize.py
```diff
@@ -123,7 +132,7 @@ def inverse(self, data: np.ndarray, **kwargs) -> np.ndarray:
 
     def log_det_jac(self, data, inverse: bool = False, **kwargs) -> np.ndarray:
         std = np.broadcast_to(self.std, data.shape)
-        ldj = np.log(np.abs(std))
+        ldj = -np.log(np.abs(std))
         if inverse:
             ldj = -ldj
```
LarsKue, line 137: Nit: A bit ugly that inverse negates the ldj twice.

bayesflow/adapters/transforms/ungroup.py
```diff
@@ -0,0 +1,83 @@
...
+        self.prefix = prefix
+        self._ungrouped = None
+
+    def get_config(self) -> dict:
+        return serialize({"key": self.key, "prefix": self.prefix, "_ungrouped": self._ungrouped})
+
+    @classmethod
+    def from_config(cls, config: dict, custom_objects=None):
+        config = deserialize(config, custom_objects)
+        _ungrouped = config.pop("_ungrouped")
```
LarsKue, line 32: This is an undisclosed side-effect with potentially fatal consequences. Consider the following snippet:

```py3
config = dict(...)
adapter_a = deserialize(config)
adapter_b = deserialize(config)  # KeyError: "_ungrouped"
```

bayesflow/approximators/__init__.py
```diff
@@ -1,4 +1,4 @@
-"""
+r"""
```
LarsKue, line 1: I'm certain this docstring worked before, why the addition of the raw string modifier now?

bayesflow/approximators/backend_approximators/jax_approximator.py
```diff
@@ -19,17 +50,34 @@ def stateless_compute_metrics(
...
-        1. Accept trainable variables as the first argument
-            (can be at any position as indicated by the argnum parameter
-             in autograd, but needs to be an explicit arg)
-        2. Accept, potentially modify, and return other state variables
-        3. Return just the loss tensor as the first value
-        4. Return all other values in a tuple as the second value
-
-        This ensures:
-        1. The function is stateless
-        2. The function can be differentiated with jax autograd
```
LarsKue, line 32: I think this doscstring was actually very helpful. Changing the structure of compute_metrics in any other way breaks the functionality. I have no problem with rewording it, but I think we should not scrap it.

bayesflow/approximators/backend_approximators/numpy_approximator.py
LarsKue, line 1: I don't think anyone uses numpy as the backend. We could scrap this approximator altogether. We never advertise usability of BayesFlow with purely Numpy anyway.

bayesflow/approximators/continuous_approximator.py
```diff
@@ -342,23 +444,68 @@ def sample(
...
+
+        if log_det_jac:
+            data, ldj = adapted
+            ldj_inference = ldj.get("inference_variables", 0.0)
+        else:
+            data = adapted
+            ldj_inference = None
+
+        # Standardize conditions
+        for key in ContinuousApproximator.CONDITION_KEYS:
```
LarsKue: use `self.CONDITION_KEYS` instead

bayesflow/approximators/continuous_approximator.py
```diff
@@ -367,73 +514,63 @@ def _sample(
...
         **kwargs,
     ) -> Tensor:
-        if self.summary_network is None:
-            if summary_variables is not None:
-                raise ValueError("Cannot use summary variables without a summary network.")
-        else:
-            if summary_variables is None:
-                raise ValueError("Summary variables are required when a summary network is present.")
+        if (self.summary_network is None) != (summary_variables is None):
+            raise ValueError("Summary variables and summary network must be used together.")
```
LarsKue: I think the more expressive errors are worth the few extra lines here.

bayesflow/approximators/continuous_approximator.py
```diff
@@ -367,73 +514,63 @@ def _sample(
...
 
         return self.inference_network.sample(
-            batch_shape,
-            conditions=inference_conditions,
-            **filter_kwargs(kwargs, self.inference_network.sample),
+            batch_shape, conditions=inference_conditions, **filter_kwargs(kwargs, self.inference_network.sample)
         )
 
-    def summaries(self, data: Mapping[str, np.ndarray], **kwargs):
+    def summaries(self, data: Mapping[str, np.ndarray], **kwargs) -> np.ndarray:
```
LarsKue: Should we call this `summarize` or `compute_summaries` instead?

bayesflow/approximators/continuous_approximator.py
```diff
@@ -468,25 +605,25 @@ def _log_prob(
...
-    ) -> Tensor | dict[str, Tensor]:
-        if self.summary_network is None:
-            if summary_variables is not None:
-                raise ValueError("Cannot use summary variables without a summary network.")
-        else:
-            if summary_variables is None:
-                raise ValueError("Summary variables are required when a summary network is present.")
+    ) -> Tensor:
+        if (self.summary_network is None) != (summary_variables is None):
+            raise ValueError("Summary variables and summary network must be used together.")
```
LarsKue: See above. Better errors > fewer lines of code

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -53,9 +59,50 @@ def __init__(
...
+        # Build classifier network and logits projector if needed
+        if not self.classifier_network.built:
+            self.classifier_network.build(classifier_conditions_shape)
+        if not self.logits_projector.built:
+            output_shape = self.classifier_network.compute_output_shape(classifier_conditions_shape)
+            self.logits_projector.build(output_shape)
+
+        # Set up standardization layers if requested
+        if self.standardize == "all":
+            self.standardize = [var for var in ModelComparisonApproximator.CONDITION_KEYS if var in data_shapes]
```
LarsKue: use `self.CONDITION_KEYS` instead

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -53,9 +59,50 @@ def __init__(
...
+            self.standardize_layers = {var: Standardization(trainable=False) for var in self.standardize}
+
+        # Build all standardization layers
+        for var, layer in getattr(self, "standardize_layers", {}).items():
```
LarsKue: Big code smell. The attribute should *always* be present.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -134,46 +180,69 @@ def compute_metrics(
...
 
         metrics = {"loss": loss} | classifier_metrics | summary_metrics
-
         return metrics
 
     def fit(
         self,
         *,
-        adapter: Adapter = "auto",
+        adapter: Adapter | str = "auto",
```
LarsKue: Do we want to allow more strings to be passed? Otherwise, the type hint should be either removed (no manual passing of `adapter="auto"`) or changed to `Literal["auto"]`.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -284,7 +356,7 @@ def predict(
         self,
         *,
         conditions: Mapping[str, np.ndarray],
-        logits: bool = False,
+        probs: bool = True,
```
LarsKue, line 360: Why deprecate this? I also dislike the abbreviation of probabilities.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -295,86 +367,111 @@ def predict(
...
+            )
+            probs = not logits
+
         # Apply adapter transforms to raw simulated / real quantities
         conditions = self.adapter(conditions, strict=False, stage="inference", **kwargs)
 
         # Ensure only keys relevant for sampling are present in the conditions dictionary
-        conditions = {k: v for k, v in conditions.items() if k in ModelComparisonApproximator.SAMPLE_KEYS}
-
+        conditions = {k: v for k, v in conditions.items() if k in ModelComparisonApproximator.CONDITION_KEYS}
```
LarsKue: use self

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -295,86 +367,111 @@ def predict(
...
-            summary_outputs = self.summary_network(
-                summary_variables, **filter_kwargs(kwargs, self.summary_network.call)
-            )
-
-            if classifier_conditions is None:
-                classifier_conditions = summary_outputs
-            else:
-                classifier_conditions = keras.ops.concatenate([classifier_conditions, summary_outputs], axis=1)
+        # Optionally standardize conditions
+        for key in ModelComparisonApproximator.CONDITION_KEYS:
```
LarsKue: use self

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -295,86 +367,111 @@ def predict(
...
+            if key in conditions and key in self.standardize:
+                conditions[key] = self.standardize_layers[key](conditions[key])
 
-        output = self.classifier_network(classifier_conditions)
-        output = self.logits_projector(output)
+        output = self._predict(**conditions, **kwargs)
 
-        return output
+        return keras.ops.convert_to_numpy(keras.ops.softmax(output) if probs else output)
```
LarsKue: please avoid inline if statements

bayesflow/approximators/point_approximator.py
```diff
@@ -55,13 +55,27 @@ def estimate(
             Each estimator output (i.e., dictionary value that is not itself a dictionary) is an array
             of shape (num_datasets, point_estimate_size, variable_block_size).
         """
+        # Adapt, optionally standardize and convert conditions to tensor.
+        conditions = self._prepare_data(conditions, **kwargs)
+        # Remove any superfluous keys, just retain actual conditions.  # TODO: is this necessary?
+        conditions = {k: v for k, v in conditions.items() if k in ContinuousApproximator.CONDITION_KEYS}
```
LarsKue: use self

bayesflow/approximators/point_approximator.py
```diff
@@ -55,13 +55,27 @@ def estimate(
...
 
-        conditions = self._prepare_conditions(conditions, **kwargs)
         estimates = self._estimate(**conditions, **kwargs)
+
+        if "inference_variables" in self.standardize:
+            for score_key, score in self.inference_network.scores.items():
+                for head_key in estimates[score_key].keys():
+                    trafo_type = score.TRANSFORMATION_TYPE.get(head_key, "location_scale")
```
LarsKue: "trafo" is not a common abbreviation for "transformation" in English (only German). Let's also avoid abbreviations.

stefanradev93:
> @stefanradev93 I cannot reproduce the failure locally, can you? I think skipping this specific test for the diffusion models (e.g. via a marker) would be acceptable, as we do not expect cycle consistency to strictly hold for the number of steps we want to put in here, and the errors can accumulate in more or less arbitrary ways. We might want to use (at least somewhat) trained models in the future, where we can expect more consistent results.

The error is stochastic and I think you are right. We should not rely on this test for now. It is still a bit strange that it's only the case for JAX.

bayesflow/approximators/point_approximator.py
```diff
@@ -107,10 +121,20 @@ def sample(
             Each output (i.e., dictionary value that is not itself a dictionary) is an array
             of shape (num_datasets, num_samples, variable_block_size).
         """
-        conditions = self._prepare_conditions(conditions, **kwargs)
+        # Adapt, optionally standardize and convert conditions to tensor.
+        conditions = self._prepare_data(conditions, **kwargs)
+        # Remove any superfluous keys, just retain actual conditions.  # TODO: is this necessary?
+        conditions = {k: v for k, v in conditions.items() if k in ContinuousApproximator.CONDITION_KEYS}
```
LarsKue: use self

bayesflow/approximators/point_approximator.py
```diff
@@ -55,13 +55,27 @@ def estimate(
...
+                    )
+
         estimates = self._apply_inverse_adapter_to_estimates(estimates, **kwargs)
+
         # Optionally split the arrays along the last axis.
         if split:
             estimates = split_arrays(estimates, axis=-1)
+
         # Reorder the nested dictionary so that original variable names are at the top.
         estimates = PointApproximator._reorder_estimates(estimates)
```
LarsKue: I'm not a big fan of class-scoped static methods. Sure, your current implementation might not rely on `self`, but that could change with future requirements, at which point you have to go back and change all of these calls to `self._reorder_estimates`.

If we really want to keep this as a staticmethod, we could at least call it using `self` here.

bayesflow/approximators/point_approximator.py
```diff
@@ -148,19 +172,20 @@ def log_prob(
...
 
-    def _prepare_conditions(self, conditions: Mapping[str, np.ndarray], **kwargs) -> dict[str, Tensor]:
-        """Adapts and converts the conditions to tensors."""
+        # Change of variables formula, respecting log_prob to be a dictionary
+        if log_det_jac is not None:
+            log_prob = keras.tree.map_structure(lambda x: x + log_det_jac, log_prob)
 
-        conditions = self.adapter(conditions, strict=False, stage="inference", **kwargs)
-        conditions = {k: v for k, v in conditions.items() if k in ContinuousApproximator.SAMPLE_KEYS}
+        log_prob = PointApproximator._squeeze_parametric_score_major_dict(log_prob)
```
LarsKue: Same comment as above on class-scoped staticmethods.

bayesflow/datasets/disk_dataset.py
```diff
@@ -42,27 +81,39 @@ def __init__(
         self.files = list(map(str, self.root.glob(pattern)))
         self.stage = stage
 
-        self.shuffle()
+        self.augmentations = augmentations
```
LarsKue: let's make this `augmentations or []` to drop the if-statement below

bayesflow/datasets/disk_dataset.py
```diff
@@ -42,27 +81,39 @@ def __init__(
...
-        batch = []
-        for file in files:
-            batch.append(self.load_fn(file))
+        batch = [self.load_fn(file) for file in files]
 
         batch = tree_stack(batch)
 
+        if self.augmentations is None:
+            pass
+        elif isinstance(self.augmentations, Mapping):
```
LarsKue, line 102: If we support this we could also support a simple sequence. I would be a bigger fan of just requiring that there is only a single augmentation, and then we outsource the iteration entirely.

bayesflow/datasets/online_dataset.py
```diff
@@ -18,19 +20,72 @@ def __init__(
...
+        Returns
+        -------
+        dict of str to np.ndarray
+            A batch of simulated (and optionally augmented/adapted) data.
+        """
         batch = self.simulator.sample((self.batch_size,))
 
+        if self.augmentations is None:
+            pass
+        elif isinstance(self.augmentations, Mapping):
```
LarsKue, line 82: See comment for disk dataset.

bayesflow/datasets/rounds_dataset.py
LarsKue, line 1: We should at least deprecate this.

bayesflow/experimental/diffusion_model/diffusion_model.py
```diff
@@ -0,0 +1,500 @@
...
+            Additional keyword arguments passed to the noise schedule constructor. Default is None.
+        integrate_kwargs : dict[str, any], optional
+            Configuration dictionary for integration during training or inference. Default is None.
+        **kwargs
+            Additional keyword arguments passed to the base class and internal components.
+        """
+        super().__init__(base_distribution="normal", **kwargs)
+
+        if prediction_type not in ["noise", "velocity", "F", "x"]:
+            raise TypeError(f"Unknown prediction type: {prediction_type}")
```
LarsKue: This is already covered by the type hint, do we really want to double-guard against stupidity?

bayesflow/experimental/free_form_flow/free_form_flow.py
```diff
@@ -86,13 +84,6 @@ def __init__(
         """
         super().__init__(base_distribution, **kwargs)
 
-        if encoder_subnet_kwargs or decoder_subnet_kwargs:
-            warnings.warn(
-                "Using `subnet_kwargs` is deprecated."
-                "Instead, instantiate the network yourself and pass the arguments directly.",
-                DeprecationWarning,
-            )
```
LarsKue, line 94: Why was this removed?

bayesflow/links/positive_definite.py
LarsKue, line 1: No deprecation?

bayesflow/metrics/functional/root_mean_squared_error.py
LarsKue, line 1: Not against this change, but I'm curious as to why it's necessary?

bayesflow/networks/coupling_flow/coupling_flow.py
```diff
@@ -40,7 +40,7 @@ class CouplingFlow(InferenceNetwork):
 
     def __init__(
         self,
-        subnet: str | type = "mlp",
+        subnet: str | type | keras.Layer = "mlp",
```
LarsKue: No, this is not correct. The subnet needs to be constructed individually for each coupling layer. For even dimensions, you can use the same one across all layers, but it would be a terrible network.

LarsKue: I'm honestly in favor of just changing the argument name to `subnet_constructor` just to be clear to users and devs alike.

bayesflow/networks/flow_matching/flow_matching.py
```diff
@@ -53,7 +51,7 @@ class FlowMatching(InferenceNetwork):
 
     def __init__(
         self,
-        subnet: str | keras.Layer = "mlp",
+        subnet: str | type | keras.Layer = "mlp",
```
LarsKue, line 54: What is the benefit of allowing `type` here?
This:
```py3
FlowMatching(subnet=MLP, subnet_kwargs=dict(widths=[64, 64]))`
```
seems much worse to me than this:
```py3
FlowMatching(subnet=MLP(widths=[64, 64]))
```

At the very least due to intellisense and static type checking.

bayesflow/utils/dispatch/find_noise_schedule.py
LarsKue, line 1: Can delete this.

bayesflow/adapters/transforms/random_subsample.py
vpratz: Fixed in a611f7060917dca05a9ed35d6a7afed95251022.

bayesflow/adapters/transforms/random_subsample.py
vpratz: Fixed in a611f7060917dca05a9ed35d6a7afed95251022.

bayesflow/datasets/__init__.py
vpratz: Yes, as far as I could tell it does not make difference for the rendering.

bayesflow/adapters/transforms/ungroup.py
vpratz: This is not correct, as `deserialize` returns a copy of the config. If you want to make it more explicit, feel free to add a `copy` call.

bayesflow/approximators/__init__.py
vpratz: Escaped spaces produce a `DeprecationWarning: invalid escape sequence '\ '` (see [here](https://github.com/bayesflow-org/bayesflow/actions/runs/15443901547/job/43468243163#step:9:1753)), which pollutes the output and leads to problem with some tools (e.g., some tools fail to process the default coverage report due to this). Moving to a raw string mitigates this and does not seem to have effects on the rendered output.

bayesflow/adapters/transforms/standardize.py
stefanradev93: This is ignored by the default warning filters, except in the __main__ module (PEP 565). As per guidelines, FutureWarning should be used for warnings intended for end users.

bayesflow/links/positive_definite.py
stefanradev93: @han-ol can take care of that.

bayesflow/experimental/free_form_flow/free_form_flow.py
vpratz: For consistency with the other networks, see https://github.com/bayesflow-org/bayesflow/issues/475 for reference.

bayesflow/metrics/functional/root_mean_squared_error.py
stefanradev93: Not sure what the comment is referencing exactly.

bayesflow/networks/flow_matching/flow_matching.py
stefanradev93: We had type for some but not for others. We either remove it everywhere and also from `find_network` or allow if everywhere.

bayesflow/networks/coupling_flow/coupling_flow.py
stefanradev93: I would definitely not touch the name of the argument, as it has been central since 2.0.0. See point below about type args.

bayesflow/metrics/functional/root_mean_squared_error.py
vpratz: I think it's not necessary anymore. Deserialization with `keras.metrics.MeanMetricWrapper` (see below) required a serializable function, but I have since changed it to fully manual deserialization, which should work without this change.

bayesflow/utils/dispatch/find_noise_schedule.py
stefanradev93: Again, not sure what the comment references.

bayesflow/networks/flow_matching/flow_matching.py
vpratz: I think it is only really necessary in `CouplingFlow`, where we need to hand around a type for downstream initialization. Do we want to keep the hint in the other inference nets for consistency, or do we remove it everywhere else?

bayesflow/utils/dispatch/find_noise_schedule.py
vpratz: The file is completely empty and can be deleted.

bayesflow/networks/coupling_flow/coupling_flow.py
stefanradev93: Fixed type.

bayesflow/approximators/continuous_approximator.py
elseml: `compute_summaries` would be most consistent imo - then `log_prob` in `ContinuousApproximator` and `PointApproximator` would be the last remaining outlier function across all Approximator classes where a noun is not preceded by a verb. What do you all think, is the consistency worth the hassle here (given that `log_prob` is quite central)?

bayesflow/networks/flow_matching/flow_matching.py
stefanradev93: I removed it from coupling flows. We may still keep it for other networks, since it is, at the very least, possible.

bayesflow/links/positive_definite.py
han-ol: I will restore the PositiveDefinite class as it were, with an additional DeprecationWarning. Or would you prever a FutureWarning?

EDIT: taken care of in https://github.com/bayesflow-org/bayesflow/pull/510/commits/a8e9a72fb551c37ad74a82f63a42c0eedb551df2

bayesflow/adapters/transforms/nnpe.py
elseml: Was catching the rather exotic case of passing non-numeric strings, now gone with #514.

bayesflow/adapters/transforms/nnpe.py
elseml: Implemented in #514.

bayesflow/adapters/transforms/nnpe.py
elseml: Implemented in #514.

bayesflow/adapters/transforms/nan_to_num.py
```diff
@@ -0,0 +1,91 @@
...
+    ----------
+    default_value : float
+        Value to substitute wherever data is NaN.
+    return_mask : bool, default=False
+        If True, a mask array will be returned under a new key.
+    mask_prefix : str, default='mask_'
+        Prefix for the mask key in the output dictionary.
+    """
+
+    def __init__(self, key: str, default_value: float = 0.0, return_mask: bool = False, mask_prefix: str = "mask"):
```
elseml, line 26: It should be clear from the docstring that `default_value` already comes with a default value of `0.0`. Something like `replacement_value` might be clearer than `default_value`.

bayesflow/workflows/basic_workflow.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         inference_variables: Sequence[str] | str = None,
         inference_conditions: Sequence[str] | str = None,
         summary_variables: Sequence[str] | str = None,
-        standardize: Sequence[str] | str = "inference_variables",
+        standardize: Sequence[str] | str | None = "inference_variables",
```
elseml, line 40: I expected a default of `standardize="all"` for the `BasicWorkflow` since users might not be too familiar with the intricacies of standardization in deep learning. Can we switch to that or are there critical reasons to restrict the default to `standardize="inference_variables"`?

elseml (approved):

README.md
vpratz: Sounds good, I have opened an issue for this, #515.

bayesflow/approximators/continuous_approximator.py
vpratz: I believe this is mostly a question of personal taste, at least I couldn't find anything in a style guide. As it does not concern the end user, I would leave it as is for this PR.

bayesflow/approximators/continuous_approximator.py
vpratz: I'm not sure, `summarize` would work, but I would be against switching to the longer `compute_...`. Given that we have `log_prob` as well I do not think that this is very pressing, but feel free to open an issue for further discussion. See [here](https://github.com/bayesflow-org/bayesflow/pull/384#issuecomment-2839934160) for the original discussion.

bayesflow/experimental/diffusion_model/diffusion_model.py
vpratz: Yes, I think we can do that here, typos etc. can happen as well, and with this we get a meaningful error message. But it should be a value error...

bayesflow/approximators/point_approximator.py
han-ol: Good point, I changed it to `transformation_type`.

bayesflow/workflows/basic_workflow.py
vpratz: @stefanradev93 What do you think?

bayesflow/approximators/point_approximator.py
han-ol: The specific functionality it brings is not likely to require changing instance state in the future, so I don't see particular reason to act here.

On the other hand, more generally we should have a repo-wide convention regarding instance-scoped versus class-scoped attributes and methods and I am fine with both (it seems like a taste-question).

bayesflow/workflows/basic_workflow.py
stefanradev93: I would keep `inference_variables` for compatibility. Also, observables often require a different type of preprocessing than targets.

bayesflow/adapters/transforms/nan_to_num.py
stefanradev93: Added description to doc.

bayesflow/approximators/continuous_approximator.py
LarsKue: I'm fine with leaving it as-is for now. The reason you want to avoid writing out the class name is simply robustness: If the user overwrites the class attribute on the instance, they expect this to change the behavior, but fetching it explicitly from the class voids this.

It's also resistant against changes to the class name, and subclassing.

bayesflow/approximators/point_approximator.py
LarsKue: There are *very* few reasons usually to use class-scoped variables or functions over instance-scoped. Unless you make the choice for a specific reason, opt for instance-scoped.

bayesflow/approximators/model_comparison_approximator.py
stefanradev93: It's pretty standard in other libs and nobody cares about logits in Bayesian model comp (statisticians don't call them logits as well).

bayesflow/approximators/model_comparison_approximator.py
stefanradev93: agreed

bayesflow/approximators/model_comparison_approximator.py
stefanradev93: agreed.

bayesflow/approximators/continuous_approximator.py
stefanradev93: I would be up for `summarize`.

bayesflow/approximators/backend_approximators/numpy_approximator.py
stefanradev93: Agreed.

bayesflow/adapters/transforms/standardize.py
stefanradev93: the -log is the correct forward ldj, unfortunately, given the somewhat confusing inverse / forward convention of adapter transforms that goes against the lit. on invertible nets.

bayesflow/datasets/disk_dataset.py
stefanradev93: Please, do not change this, as there are use cases for having a list of augmentations.

bayesflow/datasets/rounds_dataset.py
stefanradev93: This should be removed, as it has no users or use cases.

bayesflow/approximators/continuous_approximator.py
vpratz: Do we want to allow users changing this on an instance? Then I agree that we might want to change this. My impression was that the current form is specifically written as is to not allow this, also indicated by the uppcase, which marks it as constant.

bayesflow/approximators/continuous_approximator.py
vpratz: @stefanradev93 Do we want to make this change already for this release? If so let me know, and I will deprecate the old function in favor of the new name.

stefanradev93: Yes. I don't think that the function was part of any release, in which case we do not need to deprecate it.

vpratz: It was already part of [2.0.3](https://github.com/bayesflow-org/bayesflow/blob/v2.0.3/bayesflow/approximators/continuous_approximator.py#L403), so I would opt for a proper deprecation.

vpratz: Covered by #516.

bayesflow/approximators/continuous_approximator.py
LarsKue: Generally, users should not want to change this, but preventing them from doing it is unreasonable restrictive, imo

LarsKue:
Thanks everyone for the massive work in contributing these features and also in cleaning up the parts that I commented so quickly!

LarsKue (approved):

PR #511, opened 2025-06-14
Add utility function for batched simulations

### LINKED ISSUE #478: Add convenience function for batched sampling from simulator
As far as I can tell, `Simulator.sample` does not support a sample size argument, which would be nice for more memory-intensive simulators when a large number of samples is necessary. The functionality is already available via the `Simulator.rejection_sample` method, but requires specifying a valid `predicate` that returns `True` for all entries, like in this example:

```python
simulator.rejection_sample(
    num_simulations, sample_size=1000, predicate=lambda x: np.full((x["observables"].shape[0],), True)
)
```

The easiest thing would be to wrap this somehow, e.g. introduce a special case where the predicate is ignored, and use this in a `sample_batched` function. It's not a large change, but I'm not certain what the best choices are here. @LarsKue could you take a look when you find the time?

vpratz:
The implementation is a simple wrapper leveraging the batching capabilities of `rejection_sample`. It provides batched simulations along the first dimension.

Closes #478.

### DIFF
--- a/bayesflow/simulators/simulator.py
+++ b/bayesflow/simulators/simulator.py
@@ -66,3 +66,32 @@ class Simulator:
                 result = tree_concatenate([result, samples], axis=axis, numpy=True)
 
         return result
+
+    @allow_batch_size
+    def sample_batched(
+        self,
+        batch_shape: Shape,
+        *,
+        sample_size: int,
+        **kwargs,
+    ):
+        """Sample the desired number of simulations in smaller batches.
+
+        Limited resources, especially memory, can make it necessary to run simulations in smaller batches.
+        The number of samples per simulated batch is specified by `sample_size`.
+
+        Parameters
+        ----------
+        batch_shape : Shape
+            The desired output shape, as in :py:meth:`sample`. Will be rounded up to the next complete batch.
+        sample_size : int
+            The number of samples in each simulated batch.
+        kwargs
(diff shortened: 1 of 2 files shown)


stefanradev93:
Hi Valentin, thanks for the implementation! While it totally makes sense code-wise, I am currently struggling to see use cases for having a method that is ostensibly equivalent to:

`simulator.sample(batch_size * num_samples)
`

But I am pretty sure I am missing something.

vpratz:
Yes, what you missed is that this is about sampling in smaller batches, while what you wrote will lead the simulator to produce one large batch. When the simulations are memory-intensive and `batch_size * num_samples` from your example is large, this will lead to an out-of-memory error.

The implementation here is instead approximately equivalent to (pseudo-code)

```
sims = []
for i in range(ceil(batch_size / sample_size)):
    sims += simulator.sample(sample_size)
```

This requires only simulating a batch of size `sample_size` at once, and therefore has a lower memory footprint.
Note that the jargon here can add to the confusion. I have kept `batch_size=num_samples` the total number of simulations in line with `simulator.sample`, and `sample_size` the smaller number in the inner loop, in line with `rejection_sample`. If you have ideas to improve upon this, I'm happy to hear them.

stefanradev93:
I see, thanks for the clarification!

Issue #517, opened 2025-06-18
Design Discussion: Incorporating multi-input networks


vpratz:
For more complex settings, dealing with multi-modal inputs is an important requirement. This requires specifying multiple inputs, which is usually handled in a named (dictionaries) or unnamed (lists) way of passing the inputs. We have already included a few changes to allow passing dictionaries to summary networks (`Adapter.group`, `Adapter.ungroup`, handling of dictionary inputs in the new `Standardization` layer), but @LarsKue voiced concerns that the current approach leads to a cluttered/inconsistent user interface. Therefore, we want to open the discussion again to gather the requirements and needs, as well as ideas and directions for the implementation.

I see mainly the following discussion points, feel free to add more if you see other aspects:

1. Importance: Is this an edge case that only experienced users need, or do we want it to be a first-class feature just as single-input networks?
2. Which parts of the pipeline should handle the increased complexity?

Areas with models with multi-modal data include psychology (see e.g. [1]), and physical models, where the same simulator can produce data for different experimental settings, which can then be combined for more exact inference.

Some related discussion has already taken place in #503, but I think it would be good to restart the discussion here.

I will provide my current thoughts below, also tagging @LarsKue and @stefanradev93 for input.

[1] https://link.springer.com/article/10.1007/s42113-023-00167-4

vpratz:
My current thoughts on the questions:

Re 1: I think it is a common enough setting that it would be good to ensure the normal workflows work, and you get a meaningful warning/error when you try to call functionalities with a dictionary that do not support it. Full support would be great if we find a way to integrate it consistently and without too large a cost regarding maintainability.

Re 2: For now I see two main options, both have their advantages and disadvantages.

a) We allow handling dictionaries in the adapter (and have to decide how far we take it).

Pro: The approximator does not have to care about the contents of `summary_variables`, and can expect to get the correct output from the summary network without worrying how the networks and its input look like, as long as they are compatible with each other.
Con: The adapter, which is currently quite clean as it (more or less) only operates on arrays, gets substantially more complex, as it has to handle nested inputs as well. We would have to see how much we can smooth this out with flatten-restructure wrappers around transforms.

b) We try to keep the adapter "pure" (no handling of nested inputs), and assemble the required structures further downstream in the pipeline. This could be either in the approximator, or in an additional dedicated class between adapter and approximator.

Pro: The adapter is simpler, adding/maintaining transforms is easier.
Con: Depends on the exact implementation.
[...]

PR #521, opened 2025-06-24
Make diffusion model conditioning more flexible

arrjon:
I introduced a new keyword `concatenated_input`  to the `subnet_kwargs` in the diffusion model. This keyword controls how inputs—such as parameters, noise, and condition—are fed into the model’s subnet.

Previously, the model assumed all inputs were 1D vectors and concatenated them directly for the default MLP subnet. However, for more flexible architectures—such as subnets designed to preserve or induce spatial structures—this assumption doesn't hold. Now we can also return all inputs separately directly to the subnet.

### DIFF
--- a/tests/test_networks/conftest.py
+++ b/tests/test_networks/conftest.py
@@ -61,12 +61,12 @@ class ConcatenateMLP(Sequential):
 
 
 @pytest.fixture()
-def diffusion_model_edm_F_subnet_concatenate():
+def diffusion_model_edm_F_subnet_separate_inputs():
     from bayesflow.networks import DiffusionModel
 
     return DiffusionModel(
         subnet=ConcatenateMLP([8, 8]),
-        integrate_kwargs={"method": "rk45", "steps": 250},
+        integrate_kwargs={"method": "rk45", "steps": 4},
         noise_schedule="edm",
         prediction_type="F",
         concatenate_subnet_input=False,
@@ -144,11 +144,11 @@ def flow_matching():
 
 
 @pytest.fixture()
-def flow_matching_subnet_concatenate():
+def flow_matching_subnet_separate_inputs():
     from bayesflow.networks import FlowMatching
 
     return FlowMatching(
-        subnet=ConcatenateMLP([8, 8]), integrate_kwargs={"method": "rk45", "steps": 100}, concatenate_subnet_input=False
+        subnet=ConcatenateMLP([8, 8]), integrate_kwargs={"method": "rk45", "steps": 4}, concatenate_subnet_input=False
     )
 
 
@@ -160,10 +160,10 @@ def consistency_model():
 
 
 @pytest.fixture()
-def consistency_model_subnet_concatenate():
+def consistency_model_subnet_separate_inputs():
     from bayesflow.networks import ConsistencyModel
 
-    return ConsistencyModel(total_steps=100, subnet=ConcatenateMLP([8, 8]), concatenate_subnet_input=False)
+    return ConsistencyModel(total_steps=4, subnet=ConcatenateMLP([8, 8]), concatenate_subnet_input=False)
 
 
 @pytest.fixture()
@@ -230,12 +230,9 @@ def typical_point_inference_network_subnet():
         "affine_coupling_flow",
         "spline_coupling_flow",
         "flow_matching",
-        pytest.param("flow_matching_subnet_concatenate"),
         "free_form_flow",
         "consistency_model",
-        pytest.param("consistency_model_subnet_concatenate"),
         pytest.param("diffusion_model_edm_F"),
-        pytest.param("diffusion_model_edm_F_subnet_concatenate"),
         pytest.param("diffusion_model_edm_noise", marks=pytest.mark.slow),
         pytest.param("diffusion_model_cosine_velocity", marks=pytest.mark.slow),
         pytest.param("diffusion_model_cosine_F", marks=pytest.mark.slow),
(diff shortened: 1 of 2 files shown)


vpratz:
Thanks for the PR, I think this is a reasonable idea for advanced use cases. As this is another instance of multi-input networks (even though it's inside the inference network this time, and does not involve the adapter), we might want to include this in the discussion in #517. We might also think about:

- how to pass inputs: named (as a dictionary), or via position (as a tuple like you propose in the PR)
- Consistency in our multi-step models. If we offer this possibility here, we might want to do the same in flow matching and consistency models.

Tagging @LarsKue for comment as well.

stefanradev93:
One very general approach would be to break free from the fixed names, such as "inference_variables", and actually allow for:

- Marking different simulator outputs as either target variables, summary variables or inference conditions
- Selecting a strategy for how different outputs of the same **type** are handled (e.g., concatenated, passed as a tuple, or passed as keyword arguments)
This can be handled with another abstraction, such as `SimulatorOutput` with a flexible scheme.

arrjon:
> Consistency in our multi-step models. If we offer this possibility here, we might want to do the same in flow matching and consistency models.

I agree @vpratz. I added the same logic to both models. I am open to suggestions how the tensors should be passed to the subnet.

To @stefanradev93 comment: I think, this flexibility is only needed for advanced users. So maybe we should not follow this general approach for now, as the fixed names help users to get started with BayesFlow.

stefanradev93:
> > Consistency in our multi-step models. If we offer this possibility here, we might want to do the same in flow matching and consistency models.
> 
> I agree @vpratz. I added the same logic to both models. I am open to suggestions how the tensors should be passed to the subnet.
> 
> To @stefanradev93 comment: I think, this flexibility is only needed for advanced users. So maybe we should not follow this general approach for now, as the fixed names help users to get started with BayesFlow.

Absolutely, this is definitely a 2.>1.x idea.

arrjon:
For now, it is okay as it is @stefanradev93?

bayesflow/experimental/diffusion_model/diffusion_model.py
```diff
@@ -197,6 +200,35 @@ def convert_prediction_to_x(
             return (z + sigma_t**2 * pred) / alpha_t
         raise ValueError(f"Unknown prediction type {self._prediction_type}.")
 
+    def _subnet_input(
```
vpratz: I think the name and docstring do not match the behavior, it does not indicate that the input is already passed through the subnet and returns the output. I think only assembling the inputs without passing might make be the option that is easier to read. What do you think?

vpratz:
I think we need to document somewhere how this can be used (i.e., which inputs are passed to the network if `concatenate_subnet_input` is False`), as this currently only exists in the code itself. I'd suggest we pass the inputs by name, as this eases up communication (only names, no order).

It would be good to have a test in place for the `concatenate_subnet_input is False` case.

bayesflow/experimental/diffusion_model/diffusion_model.py
arrjon: I fixed the docstring and the name. I think only assembling makes the rest of the code harder to read, and in this way it is the same for the different free form approximators.

arrjon:
Thanks @vpratz for the suggestions. I added the documentation.

Regarding the test: As we do not have a network at the moment which can handle multiple inputs, I do not know a useful test for the `concatenate_subnet_input=False` case. Any suggestions?

vpratz:
@arrjon Thanks for the changes!
I would propose to add a simple wrapper network that can handle the case to the test suite, similar to other dummy networks we have. It could just take the separate inputs, concatenate them and pass them to any other network as usual. The main point for me is that we are able to notice if we accidentally break the functionality, so a basic dedicated test should be sufficient.

arrjon:
@vpratz I added the test. Now the merge should be ready! :)

vpratz:
Thanks a lot @arrjon . Sorry for being slow to review, and to always coming up with new things I didn't notice before.

The `build` functions do not take the `concatenate_subnet_input` parameter into account when building the subnet. I'm not sure in when this is problematic and when not, but I think it would be good to pass the correct shapes there as well, to avoid weird problems down the line.

arrjon:
I corrected the shape for the `build` functions. Please verify the implementation, locally the tests passed, but I am not too confident with the implementation.

bayesflow/networks/consistency_models/consistency_model.py
```diff
@@ -161,18 +168,29 @@ def build(self, xz_shape, conditions_shape=None):
...
+            self.subnet.build(input_shape)
+            out_shape = self.subnet.compute_output_shape(input_shape)
+        else:
+            # Multiple separate inputs
+            main_input_shape = xz_shape
+            time_shape = xz_shape[:-1] + (1,)  # same batch/sequence dims, 1 feature
 
-        input_shape = tuple(input_shape)
+            # Build subnet with multiple input shapes
+            input_shape = [main_input_shape, time_shape]
```
vpratz: I think if we pass by key we would use a dictionary here.

arrjon:
@vpratz now also the tests passed :)

vpratz:
Great, I'll take a final look tomorrow and then merge it.

vpratz:
I have changed the naming for the shapes, and created a reduced test for testing the setting, as the inference network tests are really slow.

arrjon:
Thanks @vpratz!

PR #522, opened 2025-06-26
add log_gamma diagnostic

daniel-habermann:
This pull request adds the log gamma diagnostic from

>     [1]  Martin Modrák. Angie H. Moon. Shinyoung Kim. Paul Bürkner. Niko Huurre.
>     Kateřina Faltejsková. Andrew Gelman. Aki Vehtari.
>     "Simulation-Based Calibration Checking for Bayesian Computation:
>     The Choice of Test Quantities Shapes Sensitivity."
>     Bayesian Anal. 20 (2) 461 - 488, June 2025. https://doi.org/10.1214/23-BA1404

In brief, deviation from uniformity of ranks in SBC is detected by computing the likelihood of the most extreme point on the empirical CDF if the rank distribution is indeed uniform (called "gamma", see equation 7 in the above reference).

We can then simulate gamma values under the null hypothesis of equal ranks and compare the empirical gamma to this null distribution. This is achieved by computing log(gamma/null_gamma), where null_gamma is the 5% quantile of the null distribution.

log_gamma < 0 implies a rejection of the hypothesis of uniform ranks at the 5% level.

### DIFF
--- a/bayesflow/diagnostics/metrics/sbc.py
+++ b/bayesflow/diagnostics/metrics/sbc.py
@@ -43,6 +43,18 @@ def log_gamma(
     quantile : float in (0, 1), optional, default 0.05
         The quantile from the null distribution to be used as a threshold.
         A lower quantile increases sensitivity to deviations from uniformity.
+
+    Returns
+    -------
+    result : dict
+        Dictionary containing:
+
+        - "values" : float or np.ndarray
+            The log gamma values per variable
+        - "metric_name" : str
+            The name of the metric ("Log Gamme").
+        - "variable_names" : str
+            The (inferred) variable names.
     """
     samples = dicts_to_arrays(
         estimates=estimates,
--- a/bayesflow/diagnostics/metrics/sbc.py
+++ b/bayesflow/diagnostics/metrics/sbc.py
@@ -52,7 +52,7 @@ def log_gamma(
         - "values" : float or np.ndarray
             The log gamma values per variable
         - "metric_name" : str
-            The name of the metric ("Log Gamme").
+            The name of the metric ("Log Gamma").
         - "variable_names" : str
             The (inferred) variable names.
     """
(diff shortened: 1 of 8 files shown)


daniel-habermann:
Sorry for the email spam, I was fighting a bit with the CI. This is ready to be reviewed now :)

bayesflow/diagnostics/metrics/sbc.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Returns
+    -------
+    result : dict
+        Dictionary containing:
+
+        - "values" : float or np.ndarray
+            The log gamma values per variable
+        - "metric_name" : str
+            The name of the metric ("Log Gamme").
```
paul-buerkner: typo in Gamme

bayesflow/diagnostics/metrics/sbc.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # rank statistics
+    ranks = np.sum(samples["estimates"] < samples["targets"][:, None], axis=1)
+
+    # null distribution and threshold
+    null_distribution = gamma_null_distribution(num_ranks, num_post_draws, num_null_draws)
+    null_quantile = np.quantile(null_distribution, quantile)
+
+    # compute log gamma for each parameter
+    log_gammas = []
```
paul-buerkner: is it common python practice to initialize a zero list and then append instead of initializing a vector of the right length and then assigining via indexing? It seems in other places you do it via np.empty

daniel-habermann: Thank you for the review!

Yeah, afaik it's not nearly as bad as in R and the performance penalty is negligible. But I agree that in this place it's a bit silly because we convert it to a numpy array anyway.

paul-buerkner:
After discussing with Stefan, I will merge it now. Thank you!

PR #525, opened 2025-06-30
Breaking changes: Fix bugs regarding counts in standardization layer

### LINKED ISSUE #524: Standardization: count is shared and gets updated multiple times, leading to wrong results
`Standardization._update_moments` is called for each flattened input. While each input has its own mean and m2, the count is shared, and therefore updated multiple times, leading to an increase by `n_inputs*batch_size` instead of only `batch_size`. This leads to wrong results.

Possible fixes are:
1. Update `count` only once in the end of the updates
2. Have one count object for each input

The first is a bit more hacky, but does not break serialization. The latter would be in line with how we handle the mean and m2, and would fit better in the existing structure in my opinion, but would break the serialization.

I will provide a fix for 1., but 2. would be easy to implement as well.

vpratz:
Fixes #524 and a related bug.

- count was accidentally updated multiple times when multiple inputs were present, leading to wrong values (#524)
- count was calculated wrongly, as only the batch size was used. Correct is the product of all reduce dimensions. This lead to wrong standard deviations.

While the batch dimension is the same for all inputs, the size of the second dimension might vary. For this reason, we need to introduce an input-specific `count` variable. This change is serialization breaking.

I added the test `test_nested_accuracy_forward` to reproduce the bugs. It fails on `dev` and is fixed by this PR.

### DIFF
--- a/bayesflow/networks/standardization/standardization.py
+++ b/bayesflow/networks/standardization/standardization.py
@@ -40,7 +40,7 @@ class Standardization(keras.Layer):
         """
         return keras.ops.where(
             self.moving_m2[index] > 0,
-            keras.ops.sqrt(self.moving_m2[index] / self.count),
+            keras.ops.sqrt(self.moving_m2[index] / self.count[index]),
             1.0,
         )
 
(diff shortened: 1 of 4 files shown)


paul-buerkner:
Thanks for the fix! It looks good to me. Still, someone else (@stefanradev93 or @LarsKue) needs to review this too as I was not involved in the initial feature and may overlook relevant aspects.

PR #528, opened 2025-07-01
Custom test quantity support for calibration_ecdf

han-ol:
This PR adds a keyword argument to calibration_ecdf for custom test quantities in simulation-based calibration checking (SBC). 

The benefits of custom test quantities are detailed in
Modrák, M., Moon, A. H., Kim, S., Bürkner, P., Huurre, N., Faltejsková, K., Gelman, A., & Vehtari, A. (2025). Simulation-Based Calibration Checking for Bayesian Computation: The Choice of Test Quantities Shapes Sensitivity. Bayesian Analysis, 20(2). https://doi.org/10.1214/23-BA1404

# Example

```python
# after training a workflow and sampling val_sim from a simulator

f = bf.diagnostics.calibration_ecdf(
    workflow.sample(conditions=val_sims, num_samples=num_post_samples),
    val_sims,
    test_quantities={  # this argument is new
        r"$\theta_1 + \theta_2$": lambda data: np.sum(data["parameters"], axis=-1),
        r"$\theta_1 - \theta_2$": lambda data: np.sum(data["parameters"] * np.array([[1, -1]]), axis=-1),
        r"$\theta_1 \cdot \theta_2$": lambda data: np.prod(data["parameters"], axis=-1),
    }, 
    difference=True, variable_keys=["parameters"], variable_names=[r"$\theta_1$", r"$\theta_2$"],
)
```
image

### DIFF
--- a/bayesflow/diagnostics/plots/calibration_ecdf.py
+++ b/bayesflow/diagnostics/plots/calibration_ecdf.py
@@ -1,6 +1,7 @@
-from collections.abc import Mapping, Sequence
+from collections.abc import Callable, Mapping, Sequence
 
 import numpy as np
+import keras
 import matplotlib.pyplot as plt
 
 from ...utils.plot_utils import prepare_plot_data, add_titles_and_labels, prettify_subplots
@@ -13,6 +14,7 @@ def calibration_ecdf(
     targets: Mapping[str, np.ndarray] | np.ndarray,
     variable_keys: Sequence[str] = None,
     variable_names: Sequence[str] = None,
+    test_quantities: dict[str, Callable] = None,
     difference: bool = False,
     stacked: bool = False,
     rank_type: str | np.ndarray = "fractional",
@@ -78,6 +80,18 @@ def calibration_ecdf(
     variable_names    : list or None, optional, default: None
         The parameter names for nice plot titles.
         Inferred if None. Only relevant if `stacked=False`.
+    test_quantities   : dict or None, optional, default: None
+        A dict that maps plot titles to functions that compute
+        test quantities based on estimate/target draws.
+
+        The dict keys are automatically added to ``variable_keys``
+        and ``variable_names``.
+        Test quantity functions are expected to accept a dict of draws with
+        shape ``(batch_size, ...)`` as the first (typically only)
+        positional argument and return an NumPy array of shape
+        ``(batch_size,)``.
+        The functions do not have to deal with an additional
+        sample dimension, as appropriate reshaping is done internally.
     figsize           : tuple or None, optional, default: None
         The figure size passed to the matplotlib constructor.
         Inferred if None.
@@ -119,6 +133,32 @@ def calibration_ecdf(
     ValueError
         If an unknown `rank_type` is passed.
     """
+
+    # Optionally, compute and prepend test quantities from draws
+    if test_quantities is not None:
+        # Prepare empty mapping to hold test quantity values
+        test_quantities_estimates = {}
+        test_quantities_targets = {}
+
+        for key, test_quantity_func in test_quantities.items():
+            # Apply test_quantity_func to draws
+            tq_targets = test_quantity_func(data=targets)
+            test_quantities_targets[key] = np.expand_dims(tq_targets, axis=1)
+
+            # We assume test_quantity_func can only handle a 1D batch_size, so estimates
+            # which have shape (num_conditions, num_post_samples, ...) must be flattend first.
+            num_conditions, num_post_samples = next(iter(estimates.values())).shape[:2]
+            flattened_estimates = keras.tree.map_structure(lambda t: np.reshape(t, (-1, *t.shape[2:])), estimates)
+            flat_tq_estimates = test_quantity_func(data=flattened_estimates)
+            test_quantities_estimates[key] = np.reshape(flat_tq_estimates, (num_conditions, num_post_samples, 1))
+
+        # Add custom test quantities to variable keys and names for plotting
+        variable_keys = list(test_quantities.keys()) + variable_keys
+        variable_names = list(test_quantities.keys()) + variable_names
+
+        # Prepend test quantities to draws
+        estimates = test_quantities_estimates | estimates
+        targets = test_quantities_targets | targets
 
     plot_data = prepare_plot_data(
         estimates=estimates,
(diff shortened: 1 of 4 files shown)


stefanradev93:
Thank you, Hans! This is an extremely useful feature. Will review in a bit.

bayesflow/diagnostics/plots/calibration_ecdf.py
```diff
@@ -120,6 +134,32 @@ def calibration_ecdf(
         If an unknown `rank_type` is passed.
     """
 
+    # Optionally, compute and prepend test quantities from draws
+    if test_quantities is not None:
+        # Prepare empty mapping to hold test quantity values
+        test_quantities_estimates = {}
+        test_quantities_targets = {}
+
+        for key, test_quantity_func in test_quantities.items():
```
paul-buerkner: could it make sense to only show test quantities when specified? Some of the quantities may need lots of parameters which need to be present for computation but we don't want to look at them. I am happy to discuss!

bayesflow/diagnostics/plots/calibration_ecdf.py
```diff
@@ -120,6 +134,32 @@ def calibration_ecdf(
...
+            # Apply test_quantity_func to draws
+            tq_targets = test_quantity_func(data=targets)
+            test_quantities_targets[key] = np.expand_dims(tq_targets, axis=1)
+
+            # We assume test_quantity_func can only handle a 1D batch_size, so estimates
+            # which have shape (num_conditions, num_samples, ...) must be flattend first.
```
paul-buerkner: can you elaborate what you do here with flattening and which logic that follows?

han-ol: Sure, so typically `targets[<variable_name>]` is an array of shape `(batch_size, num_block)`, whereas `estimates[<variable_name>]` is an array of shape `(batch_size, num_posterior_samples, num_block)`.

A test quantities of interest that we want to add to the SBC thus is either a 1D array of shape (batch_size,) for the former and a 2D array of shape `(batch_size, num_posterior_samples)` for the latter.

This means the test quantity functions that the user provides, must somehow be applied to two different input shapes.

 I thought it would be convenient if users are not required to make a test quantity function that handles different batch_shapes and it is easy enough to just flatten the estimates to a shape `(batch_size * num_posterior_samples, num_block)` before applying the user provided test quantity function [`(batch_size * num_posterior_samples)`] and reshaping afterwards [`(batch_size, num_posterior_samples)`].
 
 [The implementation does not actually require num_block to be a number, can also be a shape, I just simplified that here for clarity.]

bayesflow/diagnostics/plots/calibration_ecdf.py
han-ol: Is there a situation you have in mind where you couldn't just remove the test quantity function from the `test_quantities` dictionary?

Instead of automatically adding the `test_quantities.keys()` to the `variable_keys`, we could have the user specify test quantities and then repeat their keys (but here potentially only a subset) in `variable_names`. This would parallel what regular targets/estimates experience, but also be less straight forward from the user's point of view in my opinion.

paul-buerkner: Perhaps we misunderstood each other. I meant if we shall still show the plots of the *parameters* even when test quantities are specified (rather than only the test quantities)?

bayesflow/diagnostics/plots/calibration_ecdf.py
paul-buerkner: I see. okay that makes sense. I like this approach.

bayesflow/diagnostics/plots/calibration_ecdf.py
han-ol: Yes, you are right, thanks for your clarification.

I added a new commit so that variable_keys=[] leads to just the test quantities being plotted, while by default (variable_keys=None) they are shown.
 
```python
# after training a workflow and sampling val_sim from a simulator

f = bf.diagnostics.calibration_ecdf(
    workflow.sample(conditions=val_sims, num_samples=num_post_samples),
    val_sims,
    test_quantities={
        r"$\theta_1 + \theta_2$": lambda data: np.sum(data["parameters"], axis=-1),
        r"$\theta_1 - \theta_2$": lambda data: np.sum(data["parameters"] * np.array([[1, -1]]), axis=-1),
        r"$\theta_1 \cdot \theta_2$": lambda data: np.prod(data["parameters"], axis=-1),
    }, 
    difference=True, variable_keys=[], # setting variable_keys to an empty list ensures that ONLY the test quantities are plotted
)
```

image

```python
# after training a workflow and sampling val_sim from a simulator

f = bf.diagnostics.calibration_ecdf(
    workflow.sample(conditions=val_sims, num_samples=num_post_samples),
    val_sims,
    test_quantities={
        r"$\theta_1 + \theta_2$": lambda data: np.sum(data["parameters"], axis=-1),
        r"$\theta_1 - \theta_2$": lambda data: np.sum(data["parameters"] * np.array([[1, -1]]), axis=-1),
        r"$\theta_1 \cdot \theta_2$": lambda data: np.prod(data["parameters"], axis=-1),
    }, 
    difference=True,
)
```
image


```python
# after training a workflow and sampling val_sim from a simulator

f = bf.diagnostics.calibration_ecdf(
    workflow.sample(conditions=val_sims, num_samples=num_post_samples),
    val_sims,
    test_quantities={
        r"$\theta_1 + \theta_2$": lambda data: np.sum(data["parameters"], axis=-1),
        r"$\theta_1 - \theta_2$": lambda data: np.sum(data["parameters"] * np.array([[1, -1]]), axis=-1),
        r"$\theta_1 \cdot \theta_2$": lambda data: np.prod(data["parameters"], axis=-1),
    }, 
    difference=True, variable_names=[r"$\theta_1$", r"$\theta_2$"],
)
```
image

paul-buerkner:
This does look good to me now! Any changes you would want to do from your side? 

I believe tests are still missing. could you add some?

bayesflow/diagnostics/plots/calibration_ecdf.py
paul-buerkner: cool, I like!

han-ol:
Ok, great, I'll extend the tests, but apart from that I have no changes in mind.

han-ol:
Tests are extended and passing, so this one is ready to be merged in dev.

paul-buerkner:
Cool! @stefanradev93 will you review this quickly before merging?

LarsKue:
@stefanradev93 Is this ready to merge? Looks alright, judging from a quick glance.

PR #532, opened 2025-07-04
Ensembles of Approximators

han-ol:
This draft-PR is the result of discussions with @elseml and @stefanradev93.

The goal is fast and convenient support of approximator ensembles and the first steps for this are taken already.

1. We envision `ApproximatorEnsemble` as the abstraction at the heart of future workflows using ensembles.
    * fundamentally, it is a wrapper of a dictionary of arbitrary `Approximator` objects.
    * it overwrites the central methods compute_metrics, build, sample and passes inputs on to the respective ensemble member's methods

2. Since ensembles should cover the sensitivity on all randomness in approximators, which are nut just initialization, but also the random order of training batches, we need slightly modified datasets. 
     * For now only `OfflineEnsembleDataset` is implemented, which makes sure that training batches have an additional dimension at the second axis, containing multiple independent random slices of the available offline samples.

A few things are missing, among them are

* [ ] predict/estimate methods for `ApproximatorEnsemble` (currently `sample` exists)
* [ ] tests for `ApproximatorEnsemble`
* [ ] doc strings for `ApproximatorEnsemble`
* [ ] `OnlineEnsembleDataset`
* [ ] `DiskEnsembleDataset`
* [ ] tests for ensemble datasets
* [ ] some Workflow


han-ol:
You can check out the example notebook here: https://github.com/bayesflow-org/bayesflow/blob/ensembles/examples/ApproximatorEnsemble%20example.ipynb

elseml:
Good job Hans! Fyi, commit [955ac79](https://github.com/bayesflow-org/bayesflow/pull/532/commits/955ac794ee08e5a1b923783d224cc5b2de3f11f5) uncovered a bug in `ModelComparisonApproximator`'s `build_from_data` method which [d8d84c8](https://github.com/bayesflow-org/bayesflow/pull/532/commits/d8d84c8c4bd8395665d780ffc7ebca19144a43d2) addresses.

vpratz:
Nice work :) While trying it out a bit, the additional dimension for each ensemble member in the data caught me by surprise, and it took me a while to figure out why a dimension was missing from my data during training.
I'm not sure what the best interface would be here, but I think it would be good to think about the design for a bit. Maybe the following questions will get us closer to what we want to have:

1. Should the ordinary dataset classes be supported? If yes, which mode should they operate in (does it need a warning?), if no, we might want to override `fit` to check for them and raise an error if they are passed.
2. Should there be multiple "modes" for the `...EnsembleDatasets`, i.e., identical data vs. different data? Might be especially relevant for online training, as different data increases the required compute.
3. How do we handle custom dataset classes, what shapes do we expect from them?

As I did not take part in the discussions, maybe you already talked this through. In any way, I would be happy to hear your thoughts on this...

vpratz:
I have added serialization support, but deserialization fails when multiple approximators use the same weights, e.g. when they share a summary network. I'm not sure yet how this can be resolved, and if we want to enable serialization if we can not resolve it...

paul-buerkner:
@han-ol could you perhaps provide a minimal code example for how the ensembles should work? On that basis, it might be then also easier to discuss interface questions, including those of @vpratz.

vpratz:
@paul-buerkner You can find one as part of the PR: https://github.com/bayesflow-org/bayesflow/blob/ensembles/examples/ApproximatorEnsemble%20example.ipynb

vpratz:
@han-ol and I had a discussion on the dataset question. One approach would be a general `EnsembleDatasetWrapper` with the following properties:

- it takes in an arbitrary `dataset`, e.g. an instance of `OnlineDataset` or `OfflineDataset`
- it determines the batch size either by reading `dataset.batch_size` or by sampling a batch from the dataset
- it has a parameter like `unique_data_fraction` to control if all ensemble members get the same data (0.0) or every member gets different data (1.0). For in between values, a bootstrap procedure can be used.
- the required number of simulations can be obtained by repeatedly sampling batches from `dataset`, or, for our own simulators, by changing the `dataset.batch_size` parameter. The latter would be a bit hacky, we would have to see if we want this.

In addition, in the approximator ensemble class, we can determine by the shape of `inference_variables` if a standard dataset (like `OnlineDataset`) was used. If so, we default to showing the same data to all ensemble members.

This gives us the possibility to use our existing datasets, and only requires this one additional class to add the capability to pass different data to different approximators.

stefanradev93:
I am strongly in favor of this idea. We also discussed in the past that this is a more elegant and catch-all solution.

han-ol:
I am picking this up again now and documenting my goal here before offering an implementation.

`EnsembleDatasetWrapper` as @vpratz suggested, should wrap an arbitrary existing dataset (Online/Offline/Disk) and add an ensemble axis to the returned batches (at axis=1).

To control how independently ensemble members see data, we want to offer a parameter which Valentin previously named `unique_data_fraction`. The idea is that independent data for different ensemble members is costly and simultaneously we expect there to be diminishing returns for higher independence. A bootstrapping approach seems promising.
Instead of `unique_data_fraction` I would suggest the name `batch_independence`, because it does not give users a false expectation of true uniqueness of part of the training data.

The parameter `batch_independence` would linearly interpolate between the extreme cases. Writing $u$ for short, $u \in \[0,1\]$. The interpolation happens only for $u \in \(0,1\)$, the edges are handled separately as listed below.  

For a wrapped dataset with batch size `B` and ensemble size `E`, the wrapper collects

$$
k = \lceil (1 - u) + E \cdot u \rceil
$$

whole batches per step (no truncation, no wasted simulations), forming a pool of size $k \cdot B$. Each ensemble member then independently draws a batch of size `B` from this pool without replacement within each member batch.
This way, training samples are not repeated for individual ensemble members, but also not guaranteed to be unique across all ensemble members (which would be wasteful).

* `u = 0.0`: `k = 1`, default behavior is that all ensemble members see the same batch.
* `u = 1.0`: `k = E`, `E` independent batches are simulated for each member, every member sees only different data.
* `0.0 < u < 1.0`: `1 < k <= E`, data is different for each member, but simulation cost does not explode.

(u is not strictly the fraction of unique data, rather you could think of high $u$ reducing the expected overlap between ensemble member batches, hence I would call it `batch_independence` instead.)

The wrapper does not modify the underlying dataset (as @vpratz said, that would be hacky and I believe always rounding up $k$ sidesteps this nicely), works uniformly for online/offline/disk datasets.

At construction time, I would log how many underlying batches (`k`) will be drawn per step so the cost is transparent.

elseml:
I like it! Also the idea of providing more clarity here via logging. Since the wrapper will necessarily introduce more complexity, `batch_overlap` (reversed then ofc) might be more intuitively understood by users? In any case, I think it would be very useful to provide concrete examples like yours explicitly in the docstring to ease understanding of what is actually going on under the hood.

I think it is great to provide this functionality, but we should also keep in mind that random initialization alone already provides major benefits in deep ensembles (e.g., https://arxiv.org/abs/1912.02757).

han-ol:
I pushed a version (db21979fd5b1373c4d3dd85622a73145567c1cfb) I was working on that is close to what I specified in my previous message but with the following improvement:
Instead of sampling a batch for each member from a shared pool without replacement, we can actually stay deterministic and have overlapping windows into a shared pool. This way we can force some overlap / independence rather than getting it only in expectation. This part is already implemented now.

BUT the implementation has an important drawback still which I will address next: If the wrapper wraps `Offline/DiskDataset` the independence / overlap is only enforced within one epoch. At the end of an epoch, we shuffle the whole index array and make any previous overlap meaningless. Since we almost always want to train multiple epochs for online data, this is unexceptable...

Dividing the index sets upon construction of the wrapper object is the only way forward here I believe, and that requires the wrapper to reach into the wrapped dataset (and meddle with the self.indices attribute). To avoid this hack, but still solve the problem, I would propose to refactor the `Offline/DiskDataset` classes to have and use a `_get_batch_by_sample_indices method`. Because `Offline/DiskDataset` are quite similar already, this refactor is sensible anyway. Then the wrapper can rely on this specification.

I would proceed to that refactor, and if all preexisting tests are passing, fix the `EnsembleDatasetWrapper` for `Offline/DiskDataset`.

han-ol:
Thanks for taking a look! I like `batch_overlap`, and agree it is more intuitive, but am not sure if this might even be too intuitive. The way this works now is that the parameter linearly interpolates a `pool_factor` (`=batch_independence*num_ensembles + (1-batch_independence)`) between `1` and `num_ensembles`. Choosing windows into that pool implies some overlap. Then `1-batch_independence` is close to this overlap, but rounding artefacts can mess it up I think.

Essentially, `batch_independence` would be less suggestive in a good way? What do you think?

Also with the epoch persistant windows described above for Offline/Disk it should rather be called `data_overlap/independence` since it is not enforced at the batch but at the data level.

> I think it is great to provide this functionality, but we should also keep in mind that random initialization alone already provides major benefits in deep ensembles (e.g., https://arxiv.org/abs/1912.02757).

Yes! The goal is that we can easily investigate the uncertainty contribution of all the different levels by fixing/varying:
* data
* network initialization, different weights
* down to just changing the inference networks and sharing the summary network weights, like a last "layer" Bayesian neural network

This should help proving in practice (for any inverse problem at hand) that for example random initialization alone already provides major benefits, without always paying the fully multiplied (computational and practical implementation) cost of replicating for all levels in separate training scripts.

han-ol:
To get sensible Dataset classes and make the approximator interfaces follow consistent rules, I ended up with a few changes across approximators, datasets and their tests.

I hope the ideas are legible from the code and docs when reviewing the changes, but to give some orientation below is an **overview of the most important aspects**:

The **EnsembleDataset** class is intended as a wrapper for all bayesflow datasets. It delegates to different implementations depending on whether the wrapped dataset is online or offline/disk. The idea here is that we want different things for online and indexed datasets, but also want to serve an easy and consistent interface with a short name.  

**`sample`, `log_prob` (and `estimate`)** are now abiding the same output type for `EnsembleApproximator` as `ContinuousApproximator` and `PointApproximator`. When a `EnsembleApproximator` or `PointApproximator` has multiple distributions, it will marginalize over them in `sample` and `log_prob`. The marginalization supports weights (uniform by default); `member_weights` in the `EnsembleApproximator` and `score_weights` in the `PointApproximator`.

To sample/evaluate the distributions separately I introduced corresponding `[sample/log_prob]_separate` methods.

Along with that I renamed `ApproximatorEnsemble` -> `EnsembleApproximator`.

To make the high level sample/log_prob methods work well while keeping the flexibility of

1. not requiring all ensemble members to have sample/log_prob capabilities and
2. not requiring all scoring rules to have sample/log_prob capabilities,

I also introduced the **approximator attribute `self.distribution_keys`** (in all approximators that can have multiple distributions) **and `self.has_distribution`** in all approximators (not in `ModelComparisonApproximator` since it is generally quite different from the rest at the moment, but all others).

han-ol:
I definitely will address these **simple todos** before I would merge:

- [x]  rerun the example notebook
- [x] `member_weights` and `score_weights` should be dicts rather than lists
- [x] normalize `member_weights` and `score_weights` instead of requiring summing to 1. Only require positivity.
- [ ] reach test coverage goal (currently at 81%, target is 85%)
- [x] PR passing all tests, no conflicts with dev

And then there are changes that I would like to see soon but that are harder to do and at this point I am open to leaving them to another PR after this one has merged.

- [ ] `EnsembleApproximator`: Support for passing simulator and adapter to `EnsembleApproximator.fit`.

> Constructing a `EnsembleDataset(OnlineDataset(), num_ensembles=len(self.approximators))` is easy, but currently the `Approximator.build_dataset` method is a classmethod, so it can't access self.

A review of the committed changes would be very helpful at this point! In the meantime, I will address the simpler todos.

stefanradev93:
Excellent work, Hans! I have pushed some small revisions and want to bring a few discussion points to the table (@paul-buerkner, @LarsKue):
- Should we really allow the dataset wrapper to support `ensemble_size=1`? This introduced some check clutter down the line, so I suggest simply rejecting the creation of an ensemble with a single member (nonsensical).
- We are currently missing a workflow that can deal with ensembles. Do we want to support this?
- Do we really need two `sample` and `log_prob` functions (bulk and separate)? I would support a single function each (mirroring the names of a standard approximator) with the added option to merge the outputs (e.g., `merge_samples=False` by default) which covers the current `sample`. I would also allow the `members` argument in the singular functions along with weights.
- NRE tests are inexplicably failing on the branch despite being up to date with `dev` (@LarsKue).

han-ol:
Thanks a lot for the changes and review, Stefan!

1. Reject ensemble_size=1: agree!
2. Offering a workflow would be nice, ideally it can inherit from BasicWorkflow since the EnsembleApproximator is so very similar to the (basic) approximators. If we only overwrite the constructor (to handle more than one summary and inference network, we already have something working and can add ensemble specific plotting, met rics later.
3. sample_separate: the arguments for keeping two methods around are: 
  * The output type signature is stable
  * The keyword arguments are all used straightforwardly (member_weights would not be used in *_separate). Docstrings for methods where arguments behave differently conditional on other arguments are confusing and and hard to write well.
  * approximator.sample has always the same output structure, be it a Continuous-, a Point- or an EnsembleApproximator (this would not be the case if we set the merge_samples by default to False)
4. I think this is due to build_from_data not being implemented in NREApprox yet.

stefanradev93:
Regarding 3: Agreed, let's opt for the **default** behavior of `sample` to match that of the continuous approx. However, I am still against the idea of having multiple sample functions. My suggestion is to allow modifiers via the extra keyword args. It may not be intuitive to search for two sample functions (as was the case for v1) and it is very simple to integrate `sample_separate` into `sample`.

Regarding 4 (@LarsKue): the version of NRE here is exactly the same as the one on dev. Perhaps you can quickly check what's wrong with the test?

paul-buerkner:
I am fine with using only one sample and log_prob function but I would
prefer the default to have the same output structure as the other
approximators as I think you both agree. So then let’s merge the two
function with the default being the current sample (and log_prob)?

Stefan Radev ***@***.***> schrieb am Do. 5. Feb. 2026 um
15:15:

bayesflow/approximators/approximator.py
```diff
@@ -133,7 +133,6 @@ def fit(self, *, dataset: keras.utils.PyDataset = None, simulator: Simulator = N
             logging.info("Building on a test batch.")
             mock_data = dataset[0]
             mock_data = keras.tree.map_structure(keras.ops.convert_to_tensor, mock_data)
-            mock_data_shapes = keras.tree.map_structure(keras.ops.shape, mock_data)
-            self.build(mock_data_shapes)
+            self.build_from_data(mock_data)
```
LarsKue, line 135: Looks like this is the culprit for the failing tests. This was changed to be just `build` on the `dev` branch somewhat recently - looks like a merge from dev into this branch is missing before we merge the other way around. We could also implement `build_from_data` for the `RatioApproximator`, but I think the former solution would be preferred.

han-ol: Using build_from_data keeps the test lean. EnsembleDatasets have this additional dimension which gets dropped in the EnsembleApproximator.build_from_data. If we use build here the test has to take care of that.

It is nice to test all approximators here and not have to exclude or specifically handle ensemble.

I vote to implement build_from_data for NRE.

LarsKue: I don't mind designing code such that it is testable, but this is not a case where I feel like designing the library code around a test is the right thing to do, particularly because calling `build` should still be possible with the `EnsembleApproximator`.

That said, I will push a version of `build_from_data` for `NRE`, although this incurs unnecessary overhead.

han-ol: Thanks for the build_from_data method. I overlooked that the line you are commenting on is not actually in a test, but there are similar lines in different places in the tests as well. The point is not only about testing, sorry for the mix up!

What overhead does `build_from_data` incur? Why do we have `build_from_data` at all? For me it is only a matter of consistency - we can implement it in all approximators or not have it at all.

For the issue at hand, let me step back a bit, as I fear there is some room for missunderstanding each other.

To control which ensemble member receives which data the `EnsembleDatasets` return an additional dimension:
`(batch_size, ensemble_size, ...)`.

Whichever of the `build` and `build_from_data` methods we call in `approximator.fit` should expect shapes/tensors with this added dimension.

**Before the recent change on dev, we used `build_from_data` there, which made room for letting build_from_data expect the ensemble dimension while build does not.** Thus, the pattern below still worked even is `approximator` is an `EnsembleDatasets`.

```python
batch = adapter(simulator.sample((num_batches * batch_size,)))
batch = keras.tree.map_structure(keras.ops.convert_to_tensor, batch)
batch_shapes = keras.tree.map_structure(keras.ops.shape, batch)
approximator.build(batch_shapes)
```

Since this overrides `EnsembleApproximator.build_from_data` the two blocks below are not exchangeble anymore:
```python
approximator.build_from_data(batch)
```
and

```python
batch_shapes = keras.tree.map_structure(keras.ops.shape, batch)
approximator.build(batch_shapes)
```

**We have to decide between:**
1. break the pattern of build(tree_shapes(adapter(simulator))) (pseudo code)
2. keeping `EnsembleApproximator.build_from_data == ContinuousApproximator.build_from_data`

Currently the PR goes with 1. Choosing 2 is simple in terms of implementation. Just move the shape wrangling from `build_from_data` to `build` as shown below:

```python
    def build_from_data(self, adapted_data: dict[str, any]):
        # unchanged as in approximator.py
        data_shapes = keras.tree.map_structure(keras.ops.shape, adapted_data)
        self.build(data_shapes)

    def build(self, data_shape: dict[str, tuple[int] | dict[str, dict]]) -> None:
        # Remove the ensemble dimension from data_shape. This expects data_shape are the shapes of a
        # batch of training data, where the second axis corresponds to different approximators.
        data_shape = keras.tree.map_shape_structure(lambda shape: shape[:1] + shape[2:], data_shape)

        for approximator in self.approximators.values():
            approximator.build(data_shape)
```

han-ol:
Thanks for your reviews and discussions everyone!

**Updated TODO**

* [x] 1. Raise ValueError instead of creating an ensemble with just one member
* [x]  2. Workflow
* [x] 3. fold [sample/log_prob]_separate methods into sample/log_prob
* [x] 4. fix issue related to failing tests with more sensible build_from_data/build logic.
* [x] 5. sufficiently increase test coverage
* [x] 6. enable automatic dataset creation for `EnsembleApproximator.fit(simulator, adapter, ...)` (implement instancemethod `build_dataset`)

**Changed interface between EnsembleDataset and EnsembleApproximator:**

Instead of relying on an extra ensemble dimension, i.e. `(batch_size, ensemble_size, ...)` for each variable dict entry, ensemble data is now represented as a nested dict keyed by ensemble member (matching EnsembleApproximator.approximators)

```python
{"inference_variables":{
    "member_1": np.ndarray of shape (batch_size, ...),
    "member_2": np.ndarray of shape (batch_size, ...)
},
"summary_variables":{"member_1": ...}
}
```

This is cleaner in general and specifically lets us 
* support the  `build(tree_shapes(adapter(simulator.sample(...))))` pattern like in other approximators by checking whether a ensemble nesting is present in the build method,
* removes the need to subset batches based on the order of `ensemble.approximators.keys()` in `compute_metrics`)
* without overriding the `build_from_data` or `fit` methods.

stefanradev93:
I have proposed some refactoring changes and made an attempt on the brutal `estimate` methods, but I didn't gain much, so I reverted it. Perhaps you can draw some inspiration? Also fixed the missing parts from the notebook, which now runs head to tail. Perhaps for the sake of the tutorial we should train the final models well to show good ensemble predictions? The rest looks good to me otherwise.

han-ol:
CI failed because `self.members` was added to `get_config` which is not actually a constructor arg.

Your `_map_members` refactor is a great idea! I extended what you started, so we can remove the `*_separate` methods entirely. Previously, `_sample_separate` could not be expressed with `_map_members`, but adding a `name` arg to the `fn` helps.

Also minor docstring and comment changes.

**Open Questions**

1. I see we need an agreement about `typing.Any`/`any` for typing. Currently, we are not consistent about this in the code base. IMO lower case is hacky (since `any` is a function on iterables not a class/type) although I know it is not flagged by some type checkers. Why not use `typing.Any`?

2. `EnsembleApproximator` overwrites the classmethod `build_dataset` with an instance method. Should we make `build_dataset` an instance method also for parent classes? Tagging @LarsKue.

3. Notebook could be better. Address this pre or after merge?

stefanradev93:
Thanks, Hans!
1. Let's go for `typing.Any`!
3. I would merge with a good notebook.

PR #535, opened 2025-07-11
Log gamma test fix

daniel-habermann:
This fixes the failure probability of the `calibration_log_gamma_end_to_end()` unit test. The culprit was that each individual run of `run_sbc` was set to succeed with a probability of 99%, but the overall acceptance interval of 100 replications of them expected an individual success probability of 95%.

### DIFF
--- a/tests/test_diagnostics/test_diagnostics_metrics.py
+++ b/tests/test_diagnostics/test_diagnostics_metrics.py
@@ -127,7 +127,7 @@ def test_calibration_log_gamma_end_to_end():
         return in_interval
 
     sbc_calibration = [run_sbc(N=N, S=S, D=D) for _ in range(100)]
-    lower_expected, upper_expected = binom.ppf((0.0001, 0.9999), 100, 0.95)
+    lower_expected, upper_expected = binom.ppf((0.0005, 0.9995), 100, 0.95)
 
     # this test should fail with a probability of 0.1%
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks! I will merge as soon as the checks pass.

han-ol:
Thanks for looking at this. My involvement with this is that I saw the test failing improbably often. I am not sure to be honest how the changes in the PR correspond to your description of the original culprit above.
Nonetheless, I will not take issue with merging, since you two are the stats experts here. Just raising this since you requested my review.

daniel-habermann:
> Thanks for looking at this. My involvement with this is that I saw the test failing improbably often. I am not sure to be honest how the changes in the PR correspond to your description of the original culprit above. Nonetheless, I will not take issue with merging, since you two are the stats experts here. Just raising this since you requested my review.

It was probably the second test that was failing. Under the null hypothesis, 100 trials of a Bernoulli experiment with 95% success probability should yield between 84 and 100 successes with an alpha of 0.001. But for a 99% success probability, the corresponding interval is [94, 100].

The second test is checking that sbc fails appropriately often for a *slightly* biased posterior (e.g. a Beta(17, 17) distribution instead of a Beta(16, 17) distribution) by testing that the number of successful SBC runs is not within the null interval. Because that interval was too wide for the given probability, the null was not rejected often enough.

han-ol:
Thanks for the clarification!

PR #536, opened 2025-07-11
Stateless adapters

stefanradev93:
This PR implements the transition towards stateless adapters, as announced in v2.0.4. In particular, it:

- Removes the `stage` argument from data sets and all data set builders.
- Remove the `stage` argument from the `Adapter` and all callers.
- Removes adaptive mean computation from the `standardize` transform, as the functionality is now superseded by the `Standardize` layers in approximators. The remaining `standardize` transform should now be used only with fixed means / stds.
- Removes the `stage` argument from NNPE transform. @elseml / @vpratz / @paul-buerkner This is actually a problem for the 1% of the use cases, as we don't want the noise during inference. As I thought about it, NNPE actually fits the philosophy of an `augmentation` (introduced in 2.0.3) rather than a transform (i.e., something strictly needed only during training). Thus, my proposed solution is to introduce an `augmentations` module and start collecting these transforms there (more to come). Let me know what you think.


vpratz:
I agree that NNPE is a better fit for an augmentation than for a transform. Moving it as part of this PR (with a deprecation for the transform) sounds sensible to me.

elseml:
Agree, that should fit the use case perfectly well. Could you provide a first outline of the `augmentation` module structure for `NNPE` that you have in mind (that we can then iterate on if needed)?

bayesflow/adapters/transforms/nnpe.py
```diff
@@ -37,23 +37,6 @@ class NNPE(ElementwiseTransform):
...
-    For automatic determination, the standard deviation is determined either globally (if `per_dimension=False`) or per
-    dimension of the last axis of the input data (if `per_dimension=True`). Note that automatic scale determination is
-    applied batch-wise in the forward method, which means that determined scales can vary between batches due to varying
-    standard deviations in the batch input data.
-
-    The original implementation in [1] can be recovered by applying the following settings on standardized data:
-    - `spike_scale=0.01`
-    - `slab_scale=0.25`
-    - `per_dimension=False`
-
```
elseml, line 56: It would be nice to retain this info (of course dependent on potential changes in the implementation in the `augmentations` module), since NNPE comes with some subtle decisions/effects that can easily be missed.

bayesflow/adapters/transforms/nnpe.py
```diff
@@ -136,27 +119,18 @@ def _resolve_scale(
                     raise ValueError(f"{name}: expected scalar, got array of shape {arr.shape}")
                 return arr
 
-    def forward(self, data: np.ndarray, stage: str = "inference", **kwargs) -> np.ndarray:
+    def forward(self, data: np.ndarray, **kwargs) -> np.ndarray:
         """
-        Add spike‐and‐slab noise to `data` during training, using automatic scale determination if not provided (see
-        “Notes” section of the class docstring for details).
+        Add spike‐and‐slab noise to `data` using automatic scale determination if not provided.
+        See “Notes” section of the class docstring for details).
```
elseml: The notes section is currently removed in this PR.

bayesflow/adapters/transforms/nnpe.py
```diff
@@ -136,27 +119,18 @@ def _resolve_scale(
                     raise ValueError(f"{name}: expected scalar, got array of shape {arr.shape}")
                 return arr
 
-    def forward(self, data: np.ndarray, stage: str = "inference", **kwargs) -> np.ndarray:
+    def forward(self, data: np.ndarray, **kwargs) -> np.ndarray:
```
elseml: I'd argue for a full removal of all `NNPE` adapter transforms after moving to `augmentations`, since the current in-between state after removing the `stage` argument is very dangerous (noise is **always** applied, also during inference).

elseml (changes requested):
Everything else besides the open `NNPE` stuff looks uncritical to me. Leaving a reminder for us to shift the  `NNPE` tests once the `augmentation` is implemented.

bayesflow/adapters/transforms/nnpe.py
stefanradev93: Done.

bayesflow/adapters/transforms/nnpe.py
stefanradev93: Done

vpratz:
I like the changes overall, I'm currently removing the `.standardize()` calls from the experimental notebooks and will push the changes when I'm done.

The signature of `Adapter.standardize()` is now a bit awkward in my opinion, as `mean` and `std` are required, but are not listed in neither signature nor docstring. They are only passed via `kwargs`. Do we want to change this, @stefanradev93?

stefanradev93:
Good spot! Let's introduce them as they are described in the transform.

elseml (approved):

vpratz (approved):

PR #543, opened 2025-07-22
Auto-select backend

### LINKED ISSUE #515: Automatically set `KERAS_BACKEND` environment variable
If the user does not specifically set it, we can check for the installed backends and select one automatically. This way, users who only use one backend only have to handle installation, and do not have to worry about setting the backend correctly.

See [this comment](https://github.com/bayesflow-org/bayesflow/pull/510/files/ca9e245fbf2cd641b003a84507d988399a60340a#r2147031432).

LarsKue:
Fixes #515. I would be open to moving the functionality to some other file, but then we need to be careful about imports so that keras cannot be imported before this code is run.

### DIFF
--- a/bayesflow/__init__.py
+++ b/bayesflow/__init__.py
@@ -35,7 +35,7 @@ def setup():
                 message += f"{backend.display_name}\n"
             message += "\n"
 
-            message += f"If you continue to see this error, please file an bug report at {issue_url}.\n"
+            message += f"If you continue to see this error, please file a bug report at {issue_url}.\n"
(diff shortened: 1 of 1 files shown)


stefanradev93:
I have proposed to not error when multiple backends are detected but instead use a priority ordering + an informative warning. @LarsKue check it out.

LarsKue:
@stefanradev93 Looks good! Ready to merge from my side 🚀

han-ol:
Cool! Thanks for this.

PR #544, opened 2025-07-23
Test quantities Linear Regression Starter notebook

han-ol:
This PR contains a draft implementation of the log likelihood test quantity for SBC in the Linear Regression Starter notebook as discussed.

Feel free to edit everything, I tried to be very explicit about what happens and what changes, but it needs polishing, checking that the likelihood is actually implemented correctly wouldn't hurt and finally also explanation + citing Modrak et al.

Here is the plot that is added by the first commit of this PR:

image

See it in context here: https://github.com/bayesflow-org/bayesflow/blob/test-quantities-lin-reg-tutorial/examples/Linear_Regression_Starter.ipynb


paul-buerkner:
Thank you! will finalize this PR and then merge :-)

paul-buerkner:
I have updated the notebook. Can you proof read and edit if needed? Feel free to merge then :-)

han-ol:
Great, thank you! Looks good, so I will merge now.

PR #545, opened 2025-07-26
Breaking: parameterize MVNormalScore by inverse cholesky factor to improve stability

vpratz:
The log_prob of `MultivariateNormalScore` can be calculated using only the inverse Cholesky factor L^{-1}, but currently the Cholesky factor L is used as well, requiring the calculation of the inverse. Calculating the inverse can lead to errors due to numerical instabilities in the matrix inversion, so parameterizing the inverse directly and dropping L from the log_prob is beneficial. This also stabilizes the initial loss, and speeds up computation.

This commit contains two further optimizations. Moving the computation of the precision matrix into the einsum, and using the sum of the logs instead of the log of a product.

Open question:
- Is the transformation behavior "left_side_scale" still correct for the inverse matrix? @han-ol 
- Is "cov_chol_inv" a good name, or do we want to change it?

As the parameterization of the methods changes, this is a breaking change. As it resolves major stability problems for usage of the score in higher-dimensional problems, I thing it is worth including them anyway.

Thanks to @han-ol for support in fixing this problem.

### DIFF
--- a/bayesflow/networks/standardization/standardization.py
+++ b/bayesflow/networks/standardization/standardization.py
@@ -117,6 +117,9 @@ class Standardization(keras.Layer):
                     case "left_side_scale":
                         # x_ij = sigma_i * x_ij'
                         out = val * keras.ops.moveaxis(std, -1, -2)
+                    case "right_side_scale_inverse":
+                        # x_ij = x_ij' / sigma_j
+                        out = val / std
                     case _:
                         out = val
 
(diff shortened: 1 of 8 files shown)


han-ol:
Thanks Valentin!

> Is the transformation behavior "left_side_scale" still correct for the inverse matrix? @han-ol

No, but I added a `"right_side_scale_inverse"` that does the appropriate transformation. I also added test for it and checked if `sample` and `estimate` are consistent in a notebook for a full workflow object, so we are good here.

While updating the docstrings I noticed we currently don't include the `Standardization` layer in the API reference. I can imagine this might be intentional since it is not part of the user experience, but want to check with you anyways.

Remaining open questions:
* Do you like the new head name "precision_chol"?
* Is the exclusion of the `Standardization` class from the docs a bug?

vpratz:
Thanks a lot! I have removed the abbreviation in favor of the more comprehensive `precision_cholesky_factor`, is this fine for you?

The exclusion of the `Standardization` class from the public API is intentional for now, as end users should not have to deal with it, but we might change this if there are arguments for including it...

PR #547, opened 2025-07-27
Fix broken link to point estimation example

vpratz:
The example currently does not show in the docs due to the wrong link. Opening this PR to `main` so that we can quickly make it available on the website again.

Also fixes the examples section header in PositiveDefinite docstring.


han-ol (approved):
Thanks!

PR #548, opened 2025-07-28
fix unconditional sampling in ContinuousApproximator

vpratz:
- batch shape was calculated from `inference_conditions` even when they are `None`, leading to an error
- add approximator test for unconditional setting


han-ol:
Nice, thanks!

Issue #508, opened 2025-06-11
[FEATURE] `Adapter.reshape()`


jerrymhuang:
**Feature Description**
Very frequently, users have to change the shape of their simulation outputs as part of the configuration for generating the right data for training the approximator. Therefore, `reshape()` would be a great feature to have for the `Adapter`.

A (very rough) example usage is provided as below:

```python
sim = {"x": np.arange(6)}

adapter = (
    bf.adapters.Adapter()
    # ...
    .reshape("x", to_shape=(3, 2))
    # ...
)
```

PR #558, opened 2025-08-19
`Adapter.reshape()`

jerrymhuang:
Related to #508 .

Example usage: 

```python
sim = {"x": np.arange(6)}

adapter = (
    bf.adapters.Adapter()
    # ...
    .reshape("x", to=(3, 2))
    # ...
)
```

For this to be stateless, the original shape of data should not be stored. Therefore, the forward and inverse operations are identical.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,3097 @@
+version = 1
+revision = 3
+requires-python = ">=3.10, <3.13"
+resolution-markers = [
+    "python_full_version >= '3.12'",
+    "python_full_version == '3.11.*'",
+    "python_full_version < '3.11'",
+]
+
+[[package]]
+name = "absl-py"
+version = "2.3.1"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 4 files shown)


stefanradev93:
Thanks Jerry! The current operation can potentially break an inference pipeline. One alternative compatible with varying shape sizes) is to store and update the shape of the **most recent array** after each forward transform. This renders the adapter "semi-stateful", since it needs to be called at least once in a forward manner to be able to go in reverse, but this is a given for both the training and the inference phases.

uv.lock
vpratz, line 1: I think this uv.lock file slipped into the PR by accident.

vpratz:
I agree with Stefan that storing a state is fine here.
In my opinion, the functionality should also be able to handle batches of data of varying size. I'm not sure what the best interface for this would be. Maybe we can in/exclude axis and only reshape over the specified dimensions, keeping the others fixed? So for example to reshape over the last dimension, we could do `.reshape("key", (2,4), include_axis=[-1])` to reshape a batch of shape `(batch_size, set_size, 8)` into a shape of `(batch_size, set_size, 2, 4)`.

jerrymhuang:
Closing this because there will be a new PR about this incoming. Thanks for the feedback!

PR #562, opened 2025-08-22
Fix inconsistent type hints for `height` argument in `pairs_posterior()`

thegialeo:
This PR fixes a type inconsistency for the `height` argument in `pairs_posterior()`.

### Problem
- The `height` parameter is type hinted as `int` in the function signature.  
- The docstring lists it as `float`.  
- The helper function `_pairs_samples()` assumes `float`.  
- This mismatch causes static type checkers (e.g., mypy, pyright) to raise errors.

### Solution
- Updated the type hint in `pairs_posterior()` to `float` for consistency with both the docstring and `_pairs_samples()`.  
- Ensures type checking tools behave correctly without requiring ignores.  

### Changes
- `pairs_posterior()` → `height: float`

### DIFF
--- a/bayesflow/diagnostics/plots/pairs_posterior.py
+++ b/bayesflow/diagnostics/plots/pairs_posterior.py
@@ -1,7 +1,6 @@
-from collections.abc import Sequence, Mapping
+from collections.abc import Mapping, Sequence
 
 import matplotlib.pyplot as plt
-
 import numpy as np
 import pandas as pd
 import seaborn as sns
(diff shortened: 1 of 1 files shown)


thegialeo:
Additionally, some parameters (e.g., `dataset_id`, `variable_keys`, etc.) are currently assigned `None` by default but are type hinted as non-nullable types (`int`, `Sequence[...]`, etc.).  
Strictly speaking, these should use `Optional[...]` or the union operator (`int | None`) for full correctness.  

However, updating all of these signatures would require broader changes across the codebase.  
For now, this PR focuses only on resolving the `height` inconsistency.

PR #563, opened 2025-08-23
Multiple summary networks with identical inputs

vpratz:
In recent experiments, I found that combining multiple summary networks of the same type is an easy way to increase the capacity, without the complications of tuning the internal hyperparameters of a summary network. @paul-buerkner also indicated interest in this functionality. Therefore, I would like to enable this with an easy to use interface.

There are two options:
1. extending the `FusionNetwork` class to handle this case as well
2. Creating a new class (e.g., `SummaryNetworkEnsemble/Group`), that is similar to the `FusionNetwork`, but passes the same inputs to each summary network

I did not come up with a good, non-confusing name for option 2, so I decided to opt for option 1 and wait for your feedback and ideas on how we handle this best. Switching to option 2 would not be a lot of effort, so we can just decide what we like better.

### DIFF
--- a/bayesflow/networks/fusion_network/fusion_network.py
+++ b/bayesflow/networks/fusion_network/fusion_network.py
@@ -1,4 +1,4 @@
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 from ..summary_network import SummaryNetwork
 from bayesflow.utils.serialization import deserialize, serializable, serialize
 from bayesflow.types import Tensor, Shape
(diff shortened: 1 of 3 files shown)


stefanradev93:
Hi Valentin, thanks for the PR! I like option 1, even though the case of "single data source, multiple summaries" (slightly) defies the notion of "fusion".

vpratz:
Hi, I have implemented option 1, does one of you have capacity to review this PR, @stefanradev93 @paul-buerkner ?

paul-buerkner:
I may not yet have the expertise to fully review this PR but the implementation with either a named or unnamed set of backbones sounds sensible to me. Should it be mentioned in the doc that the Sequence option requires just the standard "summary_variables" key?

LarsKue:
@han-ol Is the proposed functionality here covered by #532?

han-ol:
The proposed functionality is closest to but different from an `EnsembleWorkflow` with multiple summary networks and a shared inference network.
The difference is that in the former summaries are concatenated, while in the latter the shared inference net receives the different summaries separately and is scored with the same loss.

This Fusion approach to increase capacity makes intuitive sense. 
To the extent that its value comes from increasing the robustness to the initialization of summary networks, I am not sure whether an EnsembleWorkflow could end up having a similar effect.

LarsKue:
Then I am in favor of closing this PR without merging. Thoughts @paul-buerkner @stefanradev93 ?

paul-buerkner:
I think so, since we don't have time to work on it now. We can reopen it should we have the capacity (or just make a new PR). @stefanradev93 what do you think?

stefanradev93:
Agreed. But then, on the other hand, it's just an extension of the `FusionNetwork`, so no harm in fetching it.

PR #565, opened 2025-08-26
Add New Flow Matching Schedules

arrjon:
Two papers in SBI context showed empirically that flow matching trained on a slightly different schedule for time improved performance. Instead of uniformly sampling t, they used a power law distribution.
[Flow Matching Posterior Estimation for Simulation-Based Atmospheric Retrieval of Exoplanets](https://ieeexplore.ieee.org/abstract/document/11106493)  
[Flow Matching for Scalable Simulation-Based Inference](https://arxiv.org/abs/2305.17161)

I added the option to flow matching to use this schedule and indeed it improves performance on two moons (512 batches, batch size 64, 50 epochs and standard settings in BayesFlow).

output

"pl" here refers to power law with alpha=-0.6, and "ot" to optimal transform


stefanradev93:
Thanks, Jonas! Should we expose the argument instead of keeping it in `kwargs`?

arrjon:
Yes, we can expose it!

PR #568, opened 2025-09-02
Enable Python 3.13 support, allow numpy 2.x

vpratz:
TensorFlow has released a Python 3.13 compatible version, which was blocking us the last time we wanted to enable Python 3.13 support. Also, I have read the [NumPy 2.0 migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html), and there was nothing that stood out to me that would require action. With the test coverage we have now, I hope anything that could cause problems will show up in the tests.

I have adapted the test to run version 3.13, I will undo this when the tests have successfully passed.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


vpratz:
The tests pass, so I think we can make this change. Feel free to merge if you agree, @stefanradev93.

PR #569, opened 2025-09-04
add parallel simulator

arrjon:
I added support for parallel sampling in `SequentialSimulator` using `joblib`. This allows simulations to be executed across multiple cores without introducing `joblib` as a hard global dependency (it is only imported locally inside the parallel method).

Do we want to support parallel execution natively in BayesFlow?

### DIFF
--- a/bayesflow/simulators/sequential_simulator.py
+++ b/bayesflow/simulators/sequential_simulator.py
@@ -70,3 +70,112 @@ class SequentialSimulator(Simulator):
             }
 
         return data
+
+    def _single_sample(self, batch_shape_ext, **kwargs) -> dict[str, np.ndarray]:
+        """
+        For single sample used by parallel sampling.
+
+        Parameters
+        ----------
+        **kwargs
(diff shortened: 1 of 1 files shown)


stefanradev93:
>Do we want to support parallel execution natively in BayesFlow?

I would be very much up for it. Can you post a small time benchmark for a model of choice?

arrjon:
For the SIR model in our examples, 32.000 simulations are done in 1.8s in parallel using 10 cores or 4.6s sequentially. For very fast simulators, using parallelization will be slower than sequential simulations. However, if simulations are performed on a computing cluster, it might be nice for users to have parallelization already ready to use for more expensive simulators.

stefanradev93:
Thanks! I would be up for merging this PR with `joblib` as a non-core dependency for now (as is) and perhaps add it as a core dependency when we parallelize all sims. What do you think @paul-buerkner?

paul-buerkner:
Agreed. I think it's nice to have it and doesn't hurt as long as it's a non-core dependency.

PR #571, opened 2025-09-06
allow tensor in DiagonalNormal dimension

arrjon:
This pull request updates the `DiagonalNormal` distribution implementation to support multi-dimensional (non-vector) distributions instead of being limited to vectors. The main changes involve replacing the single `dim` attribute with a tuple `dims`, and updating all relevant logic to handle arbitrary shapes.

This is necessary if we pass not a vector but something else to inference networks like diffusion, flow matching etc.

### DIFF
--- a/bayesflow/approximators/continuous_approximator.py
+++ b/bayesflow/approximators/continuous_approximator.py
@@ -536,8 +536,12 @@ class ContinuousApproximator(Approximator):
                 inference_conditions, (batch_size, num_samples, *keras.ops.shape(inference_conditions)[2:])
             )
 
-            target_dim = self.inference_network.base_distribution.dims
-            batch_shape = keras.ops.shape(inference_conditions)[: -len(target_dim)]
+            if hasattr(self.inference_network, "base_distribution"):
+                target_shape_len = len(self.inference_network.base_distribution.dims)
+            else:
+                # point approximator has no base_distribution
+                target_shape_len = 1
+            batch_shape = keras.ops.shape(inference_conditions)[:-target_shape_len]
         else:
             batch_shape = (num_samples,)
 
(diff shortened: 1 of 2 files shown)


stefanradev93 (commented):
Hi Jonas, thanks for the PR. See individual comments.

bayesflow/approximators/continuous_approximator.py
stefanradev93, line 1: This eliminates the possibility to sample for multiple batch dimensions, for instance, `(batch_size, time, num_samples,...)` where the target is still a vector.

bayesflow/distributions/diagonal_normal.py
stefanradev93, line 1: I would use `ops.sum`.

bayesflow/approximators/continuous_approximator.py
stefanradev93: I guess to differentiate the cases, we need to fetch the tensor dim of the target which is known by the latent distro.

bayesflow/distributions/diagonal_normal.py
arrjon: I did that before and it raised an error with jax as far as I can remember

bayesflow/approximators/continuous_approximator.py
arrjon: True, I will change this

bayesflow/distributions/diagonal_normal.py
stefanradev93: Oh, never mind, it's a constant so `sum` is good enough.

arrjon:
I changed the batch shape now to
```
target_dim = self.inference_network.base_distribution.dims
batch_shape = keras.ops.shape(inference_conditions)[: -len(target_dim)]
```
that should cover now more use cases as the shape is inferred from the conditions, but the base shape is also considered.

PR #572, opened 2025-09-08
Compositional sampling diffusion

arrjon:
This pull request introduces compositional sampling support to the BayesFlow framework, enabling diffusion models to handle multiple compositional conditions efficiently. The main changes span the continuous approximator, diffusion model, and inference network modules, adding new methods and refactoring existing ones to support compositional structures in sampling, inference, and diffusion processes.

Larger changes include:
* Added a new `compositional_sample` method to `ContinuousApproximator`, which generates samples with compositional structure and handles flattening, reshaping, and prior score computation for multiple compositional conditions. Supporting internal method `_compositional_sample` was also introduced.
* In `DiffusionModel`, implemented compositional diffusion support including:
  - New `compositional_bridge` and `compositional_velocity` methods for compositional score calculation.
  - `_compute_individual_scores` helper for handling multiple compositional conditions.
  - `_inverse_compositional` method for inverse compositional diffusion sampling.
  
 
The idea is that the workflow now has the method `compositional_sample`, which expects conditions in the form (n_datasets, n_conditions, ...). Then we can perform compositional sampling with diffusion models.
`compositional_sample` allows to set a `mini_batch_size` for memory efficient computation of the compositional score, which does not work with `jax` backend however, as `jax` does not like stochasticity in its integrators which cannot be precomputed. We could support here only fixed step sizes though?

To compute the compositional score we need access to the score of the prior. Here we need to handle the adapter carefully so that we compute the correct score. In the current draft, I was not able to compute the prior score correctly. Some ideas would be great, currently it fails because the adpater is converting stuff to numpy back and forth.

### DIFF
--- a/bayesflow/approximators/continuous_approximator.py
+++ b/bayesflow/approximators/continuous_approximator.py
@@ -731,7 +731,7 @@ class ContinuousApproximator(Approximator):
         if inference_conditions is not None:
             # Reshape conditions for compositional sampling
             # From (n_datasets * n_comp, ...., dims) to (n_datasets, n_comp, ...., dims)
-            condition_dims = keras.ops.shape(inference_conditions)[2:]
+            condition_dims = keras.ops.shape(inference_conditions)[1:]
             inference_conditions = keras.ops.reshape(
                 inference_conditions, (n_datasets, n_compositional, *condition_dims)
             )
--- a/bayesflow/approximators/continuous_approximator.py
+++ b/bayesflow/approximators/continuous_approximator.py
@@ -730,19 +730,19 @@ class ContinuousApproximator(Approximator):
 
         if inference_conditions is not None:
             # Reshape conditions for compositional sampling
-            # From (n_datasets * n_comp, dims) to (n_datasets, n_comp, dims)
-            condition_dims = keras.ops.shape(inference_conditions)[-1]
+            # From (n_datasets * n_comp, ...., dims) to (n_datasets, n_comp, ...., dims)
+            condition_dims = keras.ops.shape(inference_conditions)[2:]
             inference_conditions = keras.ops.reshape(
-                inference_conditions, (n_datasets, n_compositional, condition_dims)
+                inference_conditions, (n_datasets, n_compositional, *condition_dims)
             )
 
             # Expand for num_samples: (n_datasets, n_comp, dims) -> (n_datasets, n_comp, num_samples, dims)
             inference_conditions = keras.ops.expand_dims(inference_conditions, axis=2)
             inference_conditions = keras.ops.broadcast_to(
-                inference_conditions, (n_datasets, n_compositional, num_samples, condition_dims)
+                inference_conditions, (n_datasets, n_compositional, num_samples, *condition_dims)
             )
 
-            batch_shape = (n_datasets, n_compositional, num_samples)
+            batch_shape = (n_datasets, num_samples)
         else:
             raise ValueError("Cannot perform compositional sampling without inference conditions.")
 
@@ -769,18 +769,19 @@ class ContinuousApproximator(Approximator):
         n_datasets, n_compositional = original_shapes[first_key][:2]
 
         # Reshape samples to match compositional structure if needed
-        if len(sample_shape) == 3:  # (n_datasets * n_comp, num_samples, dims)
-            num_samples, dims = sample_shape[1], sample_shape[2]
-            inference_samples = inference_samples.reshape(n_datasets, n_compositional, num_samples, dims)
+        if len(sample_shape) == 3:  # (n_datasets * n_comp, num_samples, ..., dims)
+            num_samples, dims = sample_shape[1], sample_shape[2:]
+            inference_samples = inference_samples.reshape(n_datasets, n_compositional, num_samples, *dims)
             samples["inference_variables"] = inference_samples
 
         # For back-transformation, we might need to flatten again temporarily
         # depending on how the adapter expects the data
         flattened_samples = {}
         for key, value in samples.items():
-            if len(value.shape) == 4:  # (n_datasets, n_comp, num_samples, dims)
-                n_d, n_c, n_s, dims = value.shape
-                flattened_samples[key] = value.reshape(n_d * n_c, n_s, dims)
+            if len(value.shape) == 4:  # (n_datasets, n_comp, num_samples, ...,  dims)
+                n_d, n_c, n_s = value.shape[:3]
+                dims = value.shape[3:]
+                flattened_samples[key] = value.reshape(n_d * n_c, n_s, *dims)
             else:
                 flattened_samples[key] = value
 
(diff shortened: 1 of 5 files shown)


stefanradev93:
Hi Jonas, this is great! One general design question that I would like to discuss is whether to add the new capabilities to the existing classes or inherit from the existing classes and add the new methods there , e.g., as in `CompositionalApproximator`, `CompositionalDiffusionModel`, ... etc. The latter has the advantage that the existing interfaces remain more compact but introduces the need for new classes. @vpratz @paul-buerkner Since the interface is already working well (except for JAX), I think it's a good time to discuss.

paul-buerkner:
Where can I see examples of it's use and how it would alternatively look if the structure was different?

arrjon:
So at the moment, the compositional part is only relevant during inference. You train a diffusion model, and then you can do the following:
```
# training_data.shape = (n_datasets, ...), so no conditions
# sim_data.shape = (n_datasets, n_condtions, ...)

workflow.approximator.inference_network.integrate_kwargs = {'method': 'euler_maruyama',
                                                                   'steps': 200,
                                                                   'mini_batch_size': 2,  # how many conditions in each step are used to compute the estimate of the compositional score
                                                                   'compositional_d1': 0.05,  # density bridge 
                                                                   }

posterior_samples = workflow.compositional_sample(num_samples=100,
                                                                  conditions={'sim_data': test_data_comp_trials['sim_data']},
                                                                  compute_prior_score=prior_score)
# posterior_samples.shape = (n_datasets, n_samples, n_parameters)
```
This implementation is based on the compositional approach in [here](https://arxiv.org/abs/2505.14429).

Defining `CompositionalApproximator`, `CompositionalDiffusionModel` would essentially only change the code organization and that you have to specify the correct approximator already during training (even though there is no difference) or load the trained diffusion model into the `CompositionalApproximator` after training before you can do inference (which would mean define a new workflow and load the model, so not too difficult). A nice point about `CompositionalDiffusionModel` could be, if we want to have specific standard settings which differ to the other diffusion model, e.g., one should use a stochastic sampler rather than the deterministic one.

paul-buerkner:
Thank you. That makes sense. Is there any practical use-case where we could want to use the same diffusion model for both standard and compositional sampling?

arrjon:
I think the usual case is you know from the beginning that you want to do a compositional model. Only in rare cases, where you get new data after you trained a network, you might consider switching from diffusion to compositional diffusion.

However, at the moment a `CompositionalApproximator` might be overkill as we only have a single inference network suitable for this task.

paul-buerkner:
So you would suggest a new diffusion model class but not a new approximator class. I would personally be fine with that. That said, is there anything else beyond the `approximator.compositional_sample` method that would have to be added to the approximator class? If not, I think keeping the existing approximator is fine and checking inside of its compositional sampling method if the employed inference network supports this method.

stefanradev93:
@arrjon Can you post a minimal interface example for the latest version (model definition and sampling) to discuss with the others?

arrjon:
Current stauts: `jax` backend does not work for computing the prior score at the moment due to conflicts between the adapter (which expects `numpy`) and tracing in `jax`. The other backends are working.

LarsKue:
@arrjon Is this still in active development?

stefanradev93:
Yes.

LarsKue:
Due to some issues with the commit history on the `main`branch, we had to overwrite `dev` with the current state of `main`. This may cause unrelated commits and merge conflicts to appear in your PR. Please follow these steps to fix conflicts for your PR:

1. Check out your PR's branch with `git checkout <pr-branch>`
2. Either stash or commit and push your local changes.
3. Create a backup of the branch with `git branch <pr-branch>-backup`
4. Overwrite your PR's branch with the current state of dev using `git push origin dev:<pr-branch> --force`
5. Check the status of your local version with `git status`. If it shows many (200+) unpushed commits, reset the local state with `git reset --hard origin/<pr-branch>`
6. [`cherry-pick`](https://git-scm.com/docs/git-cherry-pick) relevant commits from the backup branch into the PR branch
7. `git push` your changes
8. If you are happy with the final state, delete the backup branch.

We can then squash-merge your PR into `dev`. Thank you for your patience.

arrjon:
I fixed the merge conflicts, the branch is now ready for a proper review

LarsKue:
@arrjon I followed my guide above to remedy the commit history for this PR. Please let me know if you think any code is missing or incorrect. Apologies for the inconvenience.

If you have local changes, here is how you can re-apply them correctly:

1. `git stash` your changes
2. reset your local branch to the remote branch with `git reset --hard origin/compositional_sampling_diffusion`
3. `git stash apply` your changes back to the local version

Then, you can continue working on the branch as before.

arrjon:
Thanks @LarsKue, looks fine!

bayesflow/approximators/helpers/conditions.py
```diff
@@ -79,26 +89,135 @@ def resolve(
...
+            The summary network.  If ``None``, ``child_summary_variables`` must also
+            be ``None`` and ``inference_conditions`` is returned as-is.
+        inference_conditions : Tensor or None, shape (flat_batch, ic_dim)
+            Already-expanded inference conditions from the merged adapter call.
+        child_summary_variables : Tensor or None, shape (n_datasets * n_children, sv_dim)
+            Un-expanded child summary variables to be summarized before expansion.
+        summary_outputs : Tensor or None
+            If already computed, the output of the summary network. If provided, this will be used instead of
+            computing summaries again from summary variables.
+        n_datasets : int
```
stefanradev93: Let's use consistently `num_stuff` instead of `n_stuff` throughout all methods.

stefanradev93:
Thanks to lots of work by @arrjon this PR is now almost ready to merge! One design choice I'd like to discuss here @daniel-habermann @paul-buerkner @LarsKue would be the following: we are currently adding lots of composition-specific functionality to existing classes (`BasicWorkflow`, `ContinuousApproximator`), which may go against the notion of lean design. Would it be better to create minimal dedicated classes (e.g., `CompositionalWorkflow`, `CompositionalApproximator`) which inherit from the existing ones and add the several necessary methods? Or should we accept the overhead to the existing classes? Please, chip in.

arrjon:
At the moment we have a distinct compositional diffusion model class. As you have to choose this before training, we can also add the compositional workflow and approximator classes, as all these choices are then in one place. However, it could make sense to remove the compositional classes entirely (even the one for the diffusion model), as during training this is essentially the standard workflow with the standard diffusion model. This would have the advantage that an already trained diffusion model can then be easily used in a compositional manner, which is not the case at the moment as it is a separate class.


If possible, a nice solution in my opinion would be to have separate classes to keep it better maintainable, but somehow allow an already trained diffusion model to be loaded into the compositional workflow, but I am not sure this is easily doable.

paul-buerkner:
Shall we discuss this next week in a call perhaps? I would need to see and understand the current structure in more detail to have an informed opinion.

Issue #559, opened 2025-08-20
[FEATURE] Recovery plot error bars for asymmetric posteriors


sjedhoff:
Currently, in the recovery diagnostic plot, the uncertainty of the posteriors (shown via error bars) is computed using the MAD through the `uncertainty_agg` argument. This can be misleading, particularly when dealing with asymmetric posterior distributions.

A more informative approach would be to allow error bars to represent credible intervals (e.g., 90% or 95%). This would capture asymmetry in the posteriors more accurately and provide a clearer picture of the underlying uncertainty.

vpratz:
Closed by #573

PR #573, opened 2025-09-11
Use credible intervals for uncertainties in recovery plots

paul-buerkner:
Fixes issue #559 by ensuring proper uncertainty bounds (via credible intervals) are used instead of symmetric error bounds.

@vpratz can you review this PR?

### DIFF
--- a/bayesflow/diagnostics/plots/recovery.py
+++ b/bayesflow/diagnostics/plots/recovery.py
@@ -3,9 +3,8 @@ from collections.abc import Sequence, Mapping
 import matplotlib.pyplot as plt
 import numpy as np
 
-from scipy.stats import median_abs_deviation
-
 from bayesflow.utils import prepare_plot_data, prettify_subplots, make_quadratic, add_titles_and_labels, add_metric
+from bayesflow.utils.numpy_utils import credible_interval
 
 
 def recovery(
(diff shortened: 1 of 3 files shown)


stefanradev93:
Nice! Should we hide the `prob` argument into optional `uncertainty_agg_kwargs` (along with optional `point_agg_kwargs`)?

paul-buerkner:
Sure, why not. Would you have time to make this change quickly?

vpratz:
Thanks for the PR, looking forward to use this!
I think it would be nice to make it explicit in the docstring which probability is used by default, as this is what users would write into their figure captions for this plot. Also, as this is an option that people might regularly want to modify, maybe we can make `point_agg_kwargs` and `uncertainty_agg_kwargs` proper keyword arguments, and document that options can be changed there (maybe even with `prob` as an example).
What do you think?

paul-buerkner:
I agree. Will make the changes tomorrow

paul-buerkner:
Made the change. @vpratz is that what you had in mind?

vpratz:
Yes, thanks for adding it, that was what I had in mind.

Playing around with the implementation, I think it is not general enough yet. In my opinion, we should support symmetric uncertainty measures (like `np.std` or `median_abs_deviation`, which was the previous default). We also have to document that for asymmetric uncertainties we expect the bounds, and not the deviations like `plt.errorbar` does.

vpratz:
I have implemented the changes, please take another look if you like it this way, @paul-buerkner

paul-buerkner:
Thank you for further improve the PR! 

I am fine with supporting symmetric uncertainty measures although I don't really see the use in them, since they tend to be misleading for anything that is not a symmetric posterior. That said, the current implementation looks good to me. 

@stefanradev93 if you agree, we can merge this one.

vpratz (approved):

PR #574, opened 2025-09-16
Fix SetTransformer kwargs handling

arrjon:
This pull request introduces improvements to keyword argument handling in `SetTransformer` (please check if this fine) and adds tests for the `ContinuousApproximator` sampling functionality. Before calling `workflow.sample` with additional `kwargs`, e.g. `method='euler'` for `FlowMatching`, was failing with the `SetTransformer`.

### DIFF
--- a/bayesflow/networks/transformers/mab.py
+++ b/bayesflow/networks/transformers/mab.py
@@ -3,7 +3,7 @@ from keras import layers
 
 from bayesflow.networks import MLP
 from bayesflow.types import Tensor
-from bayesflow.utils import layer_kwargs
+from bayesflow.utils import layer_kwargs, filter_kwargs
 from bayesflow.utils.decorators import sanitize_input_shape
 from bayesflow.utils.serialization import serializable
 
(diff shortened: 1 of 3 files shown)


stefanradev93 (commented):
I was surprised that the existing filter in `ContinuousApproximator` (L525-527) is not sufficient:

``
summary_outputs = self.summary_network(
    summary_variables, **filter_kwargs(kwargs, self.summary_network.call)
)
``

Perhaps removing the unnecessary `**kwargs` argument from all attention helpers (e.g., `MultiHeadAttentionBlock`) solves it cleanly?

Perhaps it is because the call method itself accepts `kwargs`, so the argument gets propagated down to the multihead attention block.

arrjon:
> Perhaps removing the unnecessary **kwargs argument from all attention helpers (e.g., MultiHeadAttentionBlock) solves it cleanly?

I can have a look.


> Perhaps it is because the call method itself accepts kwargs, so the argument gets propagated down to the multihead attention block.

That is indeed the problem.

arrjon:
`self.attention_blocks = keras.Sequential()` 
this part did not accept `kwargs` at all, so I removed them entirely

PR #579, opened 2025-09-23
Add empirical coverage diagnostic plots

### LINKED ISSUE #577: [FEATURE] Add empirical coverage plot to diagnostics
**Motivation**
The existing diagnostic plots, such as the rank histograms and ECDF plots are very useful to notice problems in the model (plus what kind of mismatch you see: if the model is overestimating, or being too certain, etc). But based on the ranks or their deviation from uniformity alone it is not clear how bad the consequences are for inference.

**Suggestion**

It would be nice to have equally easy access to the empirical coverage diagnostic which shows the coverage of the central posterior credible intervals on simulated datasets, which is for example also implemented in the `SCB` R package as explained in this [vignette](https://hyunjimoon.github.io/SBC/articles/rank_visualizations.html#plot_coverage-and-plot_coverage_diff---empirical-coverage).

I imagine something like this (from the `Linear_Regression_Starter.ipynb` example)

Image

lasse-meixner:
Resolves #577.

**Description**

Creates coverage plots showing empirical coverage of posterior credible intervals.
A well-calibrated model would have coverage exactly match interval width (i.e. 95% credible interval contains the true value 95% of the time) as shown by a diagonal line.

The coverage for the provided simulated datasets is accompanied by credible intervals for coverage proportion (in form of a gray ribbon). These are calculated via the standard (conjugate) Beta-Binomial model for binomial proportions with a uniform prior.

**Changes**
- adds module `bayesflow.diagnostics.plots.coverage` implementing coverage plot and coverage diff plot
- adds auxiliary function `compute_empirical_coverage` in `bayesflow.utils.plot_utils`

**Example**
Here an example of a model that is "too uncertain", or too conservative. It over-covers:

Screenshot 2025-09-23 at 13 35 47

### DIFF
--- a/bayesflow/diagnostics/plots/__init__.py
+++ b/bayesflow/diagnostics/plots/__init__.py
@@ -1,7 +1,7 @@
 from .calibration_ecdf import calibration_ecdf
 from .calibration_ecdf_from_quantiles import calibration_ecdf_from_quantiles
 from .calibration_histogram import calibration_histogram
-from .coverage import coverage, coverage_diff
+from .coverage import coverage
 from .loss import loss
 from .mc_calibration import mc_calibration
 from .mc_confusion_matrix import mc_confusion_matrix
--- a/tests/test_diagnostics/test_diagnostics_plots.py
+++ b/tests/test_diagnostics/test_diagnostics_plots.py
@@ -294,8 +294,8 @@ def test_coverage(random_estimates, random_targets):
 
 def test_coverage_diff(random_estimates, random_targets):
     # basic functionality: automatic variable names
-    out = bf.diagnostics.plots.coverage_diff(random_estimates, random_targets)
+    out = bf.diagnostics.plots.coverage(random_estimates, random_targets, difference=True)
     assert len(out.axes) == num_variables(random_estimates)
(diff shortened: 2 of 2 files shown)


stefanradev93:
Hi Lasse, I agree that this would be a very nice feature (which we implicitly care about when we compute the expected calibration error / ECE metric). And thank you for the PR, which I will review asap.

stefanradev93:
Hi Lasse, thanks for the PR and for following all contribution guidelines! I think the PR is almost merge-ready up to a small discussion on the following point: 
- Since the two new functions `coverage` and `coverage_diff` share a lot of code and logic, I wonder if it would make sense to merge them into a single function with an optional `difference` boolean flag, akin to what we have for the ECDF plots?

Also, do you happen to have a reference for the beta formulation of the coverage confidence bands which we can include in the docstring?

lasse-meixner:
@stefanradev93 sure! 
I agree with your first point, I merged the two as suggested. Regarding a source for the beta-binomial model I added a docstring reference to Chapter 2 of `Bayesian Data Analysis (2013, 3rd ed.) by Gelman A., et al. where this is covered.

PR #596, opened 2025-10-31
Adding z-score metric and fixing test_quantities problem in plot_quantity/pairs_quantity

sjedhoff:
I added the posterior z-score as a metric function, since it was already used in the `z_score_contraction` plot. This makes it available in `pairs_quantity` and` plot_quantity` as well.

I also fixed an issue that occurred when defining a custom test quantity and trying to visualize it with `pairs_quantity`/`plot_quantity`. Computing such a quantity sometimes requires attaching extra data to the estimates (e.g. the graph’s adjacency matrix). Previously this raised an error; this is now resolved.


paul-buerkner:
Thank you! Looks good. Merging now.

PR #599, opened 2025-12-21
Add conditional optimal transport for flow matching

arrjon:
This pull request introduces significant improvements to the optimal transport functionality, with a focus on supporting partial and conditional optimal transport in both the Sinkhorn and Log-Sinkhorn algorithms. 

* Added support for partial optimal transport and conditional optimal transport in both Sinkhorn and Log-Sinkhorn algorithms. Both can in principle learn a better transport map for conditional flow matching.
* Refactored the Sinkhorn and Log-Sinkhorn methods to improve computational efficiency (using the dual optimization problem). 
* Updated the `optimal_transport` API to accept and correctly propagate `conditions`, and to return resampled conditions and assignments for both balanced and unbalanced (partial/conditional) settings. 
* Changed default configuration values and documentation to reflect new options and recommended settings for conditional optimal transport.
* Refactored code in `flow_matching.py` to handle new outputs from the optimal transport functions and to properly resample conditions when provided.

### DIFF
--- a/bayesflow/networks/flow_matching/flow_matching.py
+++ b/bayesflow/networks/flow_matching/flow_matching.py
@@ -57,7 +57,7 @@ class FlowMatching(InferenceNetwork):
         "max_steps": 100,
         "atol": 1e-5,
         "partial_ot_factor": 1.0,  # no partial OT
-        "conditional_ot_ratio": 0.01,  # only used if conditions are provided
+        "condition_ratio": 0.01,  # only used if conditions are provided
     }
 
     INTEGRATE_DEFAULT_CONFIG = {


stefanradev93:
Btw, we can safely assume that the function will always be called on an embedding one or more conditions.

LarsKue:
> Btw, we can safely assume that the function will always be called on an embedding one or more conditions.

I would not make this assumption to try to make the OT functions as agnostic to the rest of the codebase as possible - unless we can see very significant simplifications in the required code here.

arrjon:
> @arrjon thank you for working on this. Is there a reason to absorb the transport of conditions into the function rather than returning assignments? I had initially implemented it that way so that it is up to the user what to do with the assignments (e.g. in case of multiple condition tensors).

When I rewrote the function to also use the condition, I decided to handle the assignments directly in the function similarly to the parameters. I thought this might be cleaner.

PR #603, opened 2025-12-24
add graphical_approximator module

daniel-habermann:
This adds support for graphical approximators. This depends on #487, so we can just rebase to dev later on. Below is a minimal working example:

```python
import bayesflow as bf
from bayesflow.adapters import Adapter
from bayesflow.experimental.graphical_approximator import GraphicalApproximator
from bayesflow.experimental.graphical_simulator.example_simulators import two_level_simulator 

# Adapter definition is much shorter now because inference variables, inference conditions and
# summary conditions can be learned from the graph.
def adapter():
    adapter = Adapter()
    adapter.to_array()
    adapter.convert_dtype("float64", "float32")

    return adapter

def summary_networks():
    return [
       bf.networks.DeepSet(summary_dim=10), # group-wise (=local) summaries
       bf.networks.DeepSet(summary_dim=10), # global summaries
    ]

def inference_networks():
    return [
        bf.networks.CouplingFlow(), # infers global parameters 
        bf.networks.CouplingFlow(), # infers local parameters
    ]

def approximator():
    # invert the `SimulationGraph' defined in the example simulator
    inverted_graph = two_level_simulator().graph.invert()

    # show which networks estimate which variables.
    # This is usually done before defining the summary and inference networks.
    print(inverted_graph.network_composition())

    # show which networks need which conditions:
    print(inverted_graph.network_conditions())

    # show a list of nodes that can be amortized over, i.e. where the joint posterior
    # can be factorized in such a way that a network is able to estimate variables
    # independently for each group
    print(inverted_graph.amortizable_node())
    
    approximator = GraphicalApproximator(
        inverted_graph, 
        adapter=adapter(), 
        inference_networks=inference_networks(),
        summary_networks=summary_networks()
    )

    return approximator
```

Open Todos:
- ~add support for serialization~ [x]
- fix backend inconsistencies
- make simulator output shape deterministic
- ~fix online training~ [x]
- ~fix amortization over variable number of nodes for offline training~ [x]
- improve type hints

### DIFF
--- a/bayesflow/experimental/graphical_approximator/graphical_approximator.py
+++ b/bayesflow/experimental/graphical_approximator/graphical_approximator.py
@@ -321,9 +321,6 @@ class GraphicalApproximator(Approximator):
                 output_signature=signature,
             ).repeat()
 
-            if not self.built:
-                self.build(data_shapes)
-
             return super(GraphicalApproximator, self).fit(*args, **kwargs, steps_per_epoch=num_batches)
         else:
             if "dataset" in kwargs:
--- a/bayesflow/experimental/graphical_approximator/utils.py
+++ b/bayesflow/experimental/graphical_approximator/utils.py
@@ -498,11 +498,7 @@ def concatenate_tf(tensors, batch_dims=1):
         Special case for tensorflow, because keras.ops.shape(t) returns a Python tuple.
         """
         s = keras.ops.shape(t)
-
-        if isinstance(s, tuple):
-            return keras.ops.convert_to_tensor(s, dtype="int32")
-
-        return keras.ops.cast(s, "int32")
+        return keras.ops.convert_to_tensor(s, dtype="int32")
 
     max_rank = max(len(t.shape) for t in tensors)
 
--- a/bayesflow/experimental/graphs/expanded_graph.py
+++ b/bayesflow/experimental/graphs/expanded_graph.py
@@ -14,7 +14,7 @@ SimulationNode: TypeAlias = str
 ExpandedNode: TypeAlias = str
 
 
-@serializable("bayesflow.experimental")
+@serializable("bayesflow.experimental")  # type: ignore[missing-argument]
 class ExpandedGraph(nx.DiGraph):
     """
     Directed graph with a similar structure as the graph defined in `SimulationGraph`,
--- a/bayesflow/experimental/graphs/inverted_graph.py
+++ b/bayesflow/experimental/graphs/inverted_graph.py
@@ -14,7 +14,7 @@ SimulationNode: TypeAlias = str
 ExpandedNode: TypeAlias = str
 
 
-@serializable("bayesflow.experimental")
+@serializable("bayesflow.experimental")  # type: ignore[missing-argument]
 class InvertedGraph(nx.DiGraph):
     """
     Directed graph representing the factorization of the joint posterior of a
--- a/bayesflow/experimental/graphs/simulation_graph.py
+++ b/bayesflow/experimental/graphs/simulation_graph.py
@@ -21,7 +21,7 @@ SimulationNode: TypeAlias = str
 ExpandedNode: TypeAlias = str
 
 
-@serializable("bayesflow.experimental")
+@serializable("bayesflow.experimental")  # type: ignore[missing-argument]
 class SimulationGraph(nx.DiGraph):
     """
     Directed acyclic graph defining a simulation composed of sampling nodes.


bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,292 @@
+from collections.abc import Mapping, Sequence
+
+import keras
+import numpy as np
+
+# TODO: add log_prob method to approximator
```
paul-buerkner: I assume you will add log_prob before merging? Or is this TODO outdated?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,292 @@
...
+
+    This approximator implements amortized Bayesian inference for simulation-based
+    models whose probabilistic structure can be expressed as a directed acyclic
+    graph (DAG). The dependency structure is encoded in an
+    :class:`bayesflow.experimental.graphs.InvertedGraph`, which specifies the order
+    in which variables are summarized, conditioned on, and inferred.
+
+    Parameters
+    ----------
+    graph : bayesflow.experimental.graphs.InvertedGraph
```
paul-buerkner: can the doc of `graph` be a little extended?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,292 @@
...
+    adapter : bayesflow.adapters.Adapter
+        Adapter for data processing. You can use :py:meth:`build_adapter`
+        to create it.
+    inference_networks : list[InferenceNetwork]
```
paul-buerkner: so we can only assign by order and not by name? Is assignment by name something we may want to have?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,292 @@
...
+        The inference networks used for posterior estimation.
+        Each network is responsible for a subset of variables, as
+        determined by the network composition defined by the supplied `graph`.
+        Call the :py:meth:`network_composition` method of the `InvertedGraph` for an
+        overview of how variables are assigned to inference networks.lihood approximation.
```
paul-buerkner: is "inference networks.lihood approximation" a typo. Or what is it meant to say exactly?

paul-buerkner (commented):
I think the PR looks great overall! I didn't review the code details but I know this has been done already separately. I appreciate the clear logic of the code and the extensive testing.

bayesflow/experimental/graphical_approximator/graphical_approximator.py
LarsKue: +1, this would definitely be nice to have. I can see it being confusing though if the interface does not fully communicate which variables can be jointly amortized over.

bayesflow/experimental/graphical_approximator/graphical_approximator.py
LarsKue: Likely a typo and everything after the period should be deleted.

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,292 @@
...
+    summary_networks : list[SummaryNetwork], optional
+        The summary networks used for data summarization (default is None).
+    standardize : str | Sequence[str] | None
+        The variables to standardize before passing to the networks. Can be either
+        "all" or any subset of the variables generated by the SimulationGraph contained
+        in the `simulation_graph` field of the supplied InvertedGraph (default is "all").
+    **kwargs : dict, optional
+        Additional arguments passed to the :py:class:`bayesflow.approximators.Approximator` class.
+    """
```
LarsKue, line 44: I'm a bit confused on why the docs for parameters to `__init__` are in the docstring for the class and not for the `__init__`. I think this is common practice in some libraries, but iirc we didn't do it this way in BayesFlow.

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,292 @@
...
+            A history object containing the training loss and metrics values.
+
+        Raises
+        ------
+        ValueError
+            If both `dataset` and `simulator` are provided or neither is provided.
+        """
+
+        if "dataset" in kwargs.keys():
+            if type(kwargs["dataset"]) is SimulationOutput:
```
LarsKue: Use `isinstance` instead

bayesflow/experimental/graphical_approximator/graphical_approximator.py
daniel-habermann: > so we can only assign by order and not by name? Is assignment by name something we may want to have?

Assignment by name is a bit tricky because multiple nodes can be estimated by the same network. We could have a `dict[list[Node], InferenceNetwork]`, but this would get messy quite fast. Or were you thinking of assigning inference networks by network idx? I.e. `{0: CouplingFlow(...), 1: CouplingFlow(...)}`

bayesflow/experimental/graphical_approximator/graphical_approximator.py
daniel-habermann: edit: yes, log_prob will be added next now that I'm mostly done with unit tests.

paul-buerkner: Very good!

bayesflow/experimental/graphical_approximator/graphical_approximator.py
paul-buerkner: I can see that default names would be tricky perhaps. But perhaps we can come up with something that does not blow up, say, the first node of the "block" with an "_others" added if there are more than one node in a block? Of course, we should continue to support lists to. So we can also add support for names later on since I would like lists to continue being supported anyway even once we support dicts too.

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,294 @@
...
+
+        Parameters
+        ----------
+        stage : str, optional
+            Current training stage (e.g., "training", "validation", "inference"). Controls
+            the behavior of standardization and some metric computations (default is "training").
+
+        Returns
+        -------
+        metrics : dict[str, Tensor]
```
elseml, line 203: Now returns a `total_loss, combined_metrics` tuple of dicts. I suppose it is necessary to deviate from the convention of returning a single dict here?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,294 @@
...
+            inference_metrics[i] = inference_network.compute_metrics(
+                inference_variables[i], conditions=inference_conditions[i], stage=stage
+            )
+
+        # combine metrics
+        total_loss = 0
+        combined_metrics = {}
+
+        for i, metric_type in enumerate([summary_metrics, inference_metrics]):
+            prefix = "summary_metrics" if i == 0 else "infrence_metrics"
```
elseml: Typo in `infrence_metrics`

bayesflow/experimental/graphical_approximator/utils.py
```diff
@@ -0,0 +1,536 @@
...
+    conditions = []
+    nodes_to_condition_on = set(network_conditions) - {data_node} - set(network_composition)
+
+    for node in sort_nodes_topologically(approximator.graph.simulation_graph, list(nodes_to_condition_on)):
+        for name in variable_names[node]:
+            var = data[name]
+
+            # standardize conditions if required
+            if name in approximator.standardize:
+                var = approximator.standardize_layers[name](var, staging="validation")
```
elseml: `stage` instead of `staging`?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,294 @@
...
+                target_shape = (batch_size, num_samples, *keras.ops.shape(v)[1:])
+                sample_dict[k] = keras.ops.convert_to_numpy(keras.ops.reshape(v, target_shape))
+
+        return sample_dict
+
+    def _batch_size_from_data(self, data):
+        """
+        Fetches the current batch size from an input dictionary.
+        """
+        data_shapes = self.data_shapes(data)
```
elseml: `self.data_shapes` is defined as an attribute during building, I guess you mean `_data_shapes` here?

bayesflow/experimental/graphical_approximator/utils.py
```diff
@@ -0,0 +1,536 @@
...
+    data_keys = approximator.graph.simulation_graph.variable_names()[data_node]
+
+    summary_input_shape = concatenate_shapes([data_shapes[k] for k in data_keys])
+    shape_order = approximator.graph.data_shape_order()
+
+    repetitions = {}
+
+    # looping in reverse for easier indexing
+    for i, variable in enumerate(shape_order):
+        repetitions[variable] = summary_input_shape[i - len(shape_order) - 1]
```
elseml: I'm not too familiar with this part of the code, but is this indexing (giving indices `-(len+1), -len, ... , -2`) intended?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,294 @@
...
+        Generates samples from the approximator given input conditions. The `conditions` dictionary is preprocessed
+        using the `adapter`. Samples are converted to NumPy arrays after inference.
+
+        Parameters
+        ----------
+        num_samples : int
+            Number of samples to generate.
+        conditions : dict[str, np.ndarray]
+            Dictionary of conditioning variables as NumPy arrays.
+        split : bool, default=False
```
elseml, line 349: Deprecated but `conditions` missing?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        # compute summary metrics
+        summary_inputs = summary_inputs_by_network(self, kwargs)
+        summary_metrics = {}
+
+        for i, summary_network in enumerate(self.summary_networks or []):
+            summary_metrics[i] = summary_network.compute_metrics(summary_inputs[i], stage=stage)
+            summary_metrics[i].pop("outputs")
+
+        # compute inference metrics
```
elseml, line 217: Is it intended that both `inference_conditions_by_network` and `inference_variables_by_network` ignore the `stage` argument passed to `compute_metrics`? Both have `stage="validation"` hardcoded in their respective prepare helper function

bayesflow/experimental/graphical_approximator/utils.py
```diff
@@ -0,0 +1,536 @@
...
+
+    return conditions
+
+
+def add_node_reps_to_conditions(conditions, repetitions: Mapping[str, int]):
+    """
+    Appends node repetition features to a conditions tensor.
+    """
+    rep_values = keras.ops.convert_to_tensor(list(repetitions.values()))
+    squared = keras.ops.sqrt(rep_values)
```
elseml, line 245: returns square root

bayesflow/experimental/graphical_approximator/graphical_approximator.py
daniel-habermann: this was intended, but is outdated now since changing the `split_network_output` function.

bayesflow/experimental/graphical_approximator/utils.py
daniel-habermann: yes, this is intended. Updated the indexing to be easier to follow

bayesflow/experimental/graphical_approximator/graphical_approximator.py
daniel-habermann: could you elaborate what you mean by this?

bayesflow/experimental/graphical_approximator/graphical_approximator.py
daniel-habermann: yes, this is intended since we are just collecting the required inputs.

bayesflow/experimental/graphical_approximator/utils.py
daniel-habermann: intended because we don't want to condition on node repetitions directly. Updated the docstring to make this clear.

bayesflow/experimental/graphical_approximator/graphical_approximator.py
daniel-habermann: I changed how the `compute_metrics` method is called and it now returns a single dict.

daniel-habermann:
> Great to see this huge PR moving to the codebase! I left some specific comments, with one additional (minor) general comment: It would be great if you could do a brief sweep for type hint completeness, especially concerning the returned objects (`->`; e.g., in `compute_metrics` and most utils functions).

Thank you for reviewing the PR! :) Will definitely add more type hints once and outstanding issue with the tensorflow backend is resolved.

bayesflow/experimental/graphical_approximator/graphical_approximator.py
elseml: sorry, overlooked the `conditions` arg in the docstring, so this only refers to the `split` argument being described in the docstring but seeming to be deprecated

PR #606, opened 2026-01-09
Remove assert statements to support Python -O / -OO

arunkumargururaj07-star:
This fixes test failures under Python’s optimization flags (-O / -OO).  

I found that test_copy_unequal was silently passing under -O because the helper assert_layers_equal used assert statements — which get stripped when assertions are disabled.  

This PR replaces all assert calls in tests/utils/assertions.py with explicit ValueError raises, so the tests behave consistently regardless of optimization level.  

Also cleaned up a few commented-out assert lines for clarity.

### DIFF
--- a/tests/utils/assertions.py
+++ b/tests/utils/assertions.py
@@ -1,4 +1,3 @@
-# Import Layer directly from keras.layers
 from keras.layers import Layer
 import keras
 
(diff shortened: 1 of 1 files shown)


LarsKue (changes requested):
@arunkumargururaj07-star Thank you for the PR! Unfortunately, I think the changes you made are not fully in the spirit of the corresponding issue. Since we use `pytest` to run the tests, the best practice is to actually write the plain `assert` statements in the tests. The idea of the issue was to add a git workflow that runs the tests with the `-O` or `-OO` flag. `pytest` actually supports running the tests with these flags, since it automatically rewrites `assert` statements in the test directory at import time (i.e., before they are stripped from the optimization). `assert` statements in the main codebase will still be stripped, as expected.

I'm sorry if this was not clear from the issue I made. Would you be able to change this PR or (probably cleaner) open a new one that does not change the tests themselves, but rather the git workflows?

arunkumargururaj07-star:
@LarsKue Thanks for the clarification. I’ve opened this new PR which only updates the CI workflow to run tests under -O / -OO, without modifying any test code.

PR #608, opened 2026-01-09
Run test suite under Python -O / -OO in CI

arunkumargururaj07-star:
This PR updates the CI workflow to run the test suite under Python optimization
flags (-O and -OO).

The goal is to ensure that tests behave correctly when assertions in the main
codebase are stripped, while relying on pytest’s assertion rewriting for test
files.

No test logic or helper functions were modified — only the GitHub Actions
workflow was updated to add optimized test runs.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,6 +26,7 @@ jobs:
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.10"]  # we usually only need to test the oldest python version
         backend: ["jax", "tensorflow", "torch"]
+        optimization: ["", "-O", "-OO"]
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -26,6 +26,7 @@ jobs:
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.10"]  # we usually only need to test the oldest python version
         backend: ["jax", "tensorflow", "torch"]
+        optimization: ["", "-O", "-OO"]
```
LarsKue, line 29: Running the optimizations in the matrix is too expensive. Can we add a separate manual-trigger only workflow to handle this instead?

.github/workflows/tests.yaml
```diff
@@ -70,7 +71,7 @@ jobs:
 
       - name: Run Tests
         run: |
-          pytest -x -m "not slow"
+          python ${{ matrix.optimization }} -m pytest tests/ --tb=short -v
```
LarsKue, line 74: Can you go into some detail why you chose to make this verbose but shorten the traceback? I'm not entirely sure of the implications for the output yet.

tests/utils/assertions.py
LarsKue, line 1: Please keep this file unchanged with respect to the dev branch.

LarsKue:
Closing this due to inactivity

PR #609, opened 2026-01-12
Add batch sampling to workflow

arrjon:
This pull request introduces support for batched sampling in the `BasicWorkflow` class, allowing users to efficiently generate samples in manageable batches to address memory constraints. I often use a wrapper around the workflow to do this, so I thought it might be useful for more people.


stefanradev93:
This is an important feature! Should we also add a progress bar to the loop?

arrjon:
I added the progress bar and made sure it works with PR #611

stefanradev93:
Thanks, Jonas! Made a few changes to simplify the code and use existing utilities (e.g., `tree_concatenate`). Also switched to `tqdm.auto` for properly displaying the progress bar in a notebook` and added the `batch_size` option to `estimate`. See if you are fine with my changes.

EDIT: Added recursive checks for nested dicts of tensors (e.g., as required by fusion network).

arrjon:
Looks much nicer now!

PR #610, opened 2026-01-13
fix output projector initialization in single coupling

arrjon:
This changes how the `output_projector` layer is initialized in the `SingleCoupling` class. The main change is that the layer is no longer created during object instantiation, but is instead initialized in the `build` method where the necessary shape information is available. With keras 3.13.0, I could not use pass `None` to the `Dense` layer.


stefanradev93:
Good catch. This explains the recently reported error in the forums: https://discuss.bayesflow.org/t/units-none-set-in-equivariantlayer/220/8

I went ahead and fixed it across all layers using `units=None`.

arrjon:
I fixed the ones missing and made the output projector use standard initialization with zero bias for diffusion type models, as targets are something like a normal distribution usually.

PR #617, opened 2026-01-19
Fix density eval adaptive solvers

arrjon:
This pull request updates the integration method used in test configurations for diffusion models and fixed the adaptive step size calculation in the ODE solver to work with additional trace values for density calculation.

### DIFF
--- a/bayesflow/networks/diffusion_model/diffusion_model.py
+++ b/bayesflow/networks/diffusion_model/diffusion_model.py
@@ -369,9 +369,9 @@ class DiffusionModel(InferenceNetwork):
 
         if integrate_kwargs["method"] in STOCHASTIC_METHODS:
             logging.warning(
-                "Stochastic methods are not supported for forward integration."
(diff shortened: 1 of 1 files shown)


arrjon:
The function `from bayesflow.utils import jacobian` currently does not support adaptive samplers with the `jax` backend. I therefore, changed the test to use the fixed step size for the density calculation. However, density computation now works also with the adaptive solvers for all backends.

stefanradev93:
Idea: Instead of erroring on SDE solver for density estimation (not supported by design, but faster for sampling -> hence default), fallback on best possible ODE solver with a warning and a hint that the warning can be suppressed by explicitly passing a method contained in `DETERMINISTIC_SOLVERS`.

paul-buerkner:
I agree

Am Mo., 19. Jan. 2026 um 15:16 Uhr schrieb Stefan Radev <
***@***.***>:

PR #618, opened 2026-01-19
Enable diffusion guidance

arrjon:
This pull request adds support for custom guidance during diffusion model sampling, including constraint-based and function-based guidance. The main changes are the introduction of the `constraint_guidance_term` method, updates to the `score`, `velocity`, and related methods to accept and apply custom guidance, and new tests for guidance features.

Adding guidance is now very easy: 
```
    def constraint(z):
        params = workflow.approximator.standardize_layers["inference_variables"](z, forward=False)
        a1 = params[..., 0]
        return a1  # a1 < 0, pick one moon

    samples_guided = workflow.sample(
        num_samples=10, conditions=test_conditions, constraint_guidance=dict(constraints=constraint) # this adds a constrain
    )

    def guidance_function(x, time):
        return x * 0  # dummy function, could do anything you want, is then added directly to the score

    samples_guided_func = workflow.sample(
        num_samples=10, conditions=test_conditions, guidance_function=guidance_function
    )
```

### DIFF
--- a/tests/test_networks/test_diffusion_model/test_diffusion_model.py
+++ b/tests/test_networks/test_diffusion_model/test_diffusion_model.py
@@ -47,7 +47,7 @@ def test_diffusion_guidance():
         return a1  # a1 < 0, pick one moon
 
     samples_guided = workflow.sample(
-        num_samples=10, conditions=test_conditions, guidance_constraints=dict(constraints=constraint)
+        num_samples=2, conditions=test_conditions, guidance_constraints=dict(constraints=constraint)
     )["parameters"]
     assert samples_guided.shape == samples.shape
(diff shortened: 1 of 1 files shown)


stefanradev93:
Excellent! Made the naming more consistent and used the opportunity to remove redundant args. Should be good to merge once the tests have passed.

arrjon:
In the tests for inference networks: We should remove the specification of the sampler in the model definition, to always test the default samplers, now that samplers switch automatically for density computation. What do you think?

stefanradev93:
> In the tests for inference networks: We should remove the specification of the sampler in the model definition, to always test the default samplers, now that samplers switch automatically for density computation. What do you think?

Agreed.

PR #619, opened 2026-01-20
Use batch_size argument in simulators instead of batch_shape

daniel-habermann:
Consider a simulator like:

```
import bayesflow as bf
import numpy as np

def prior():
    theta = np.random.uniform(-1, 1, 2)
    return {"theta": theta}


def forward_model(theta):
    theta = np.random.uniform(-1, 1, 2)
    alpha = np.random.uniform(-np.pi / 2, np.pi / 2)
    r = np.random.normal(0.1, 0.01)
    x1 = -np.abs(theta[0] + theta[1]) / np.sqrt(2) + r * np.cos(alpha) + 0.25
    x2 = (-theta[0] + theta[1]) / np.sqrt(2) + r * np.sin(alpha)

    return dict(x=np.array([x1, x2]))

simulator = bf.make_simulator([prior, forward_model])
```

This PR changes the signature to the sample call of all simulator classes to use a single batch size, i.e. `simulator.sample(10)["theta"].shape == (10, 2)`. To support arbitrary additional batch dimensions, the sample method now has an additional `sample_shape` argument, i.e. `simulator.sample(10, sample_shape=(5, 7))["theta"].shape == (10, 5, 7, 2)`.

The purpose of this PR is mainly to stay consistent with #616 and to avoid the need for decorators.

### DIFF
--- a/tests/utils/normal_simulator.py
+++ b/tests/utils/normal_simulator.py
@@ -1,8 +1,7 @@
 import numpy as np
 
 from bayesflow.simulators import Simulator
-from bayesflow.types import Shape, Tensor
-from bayesflow.utils.decorators import allow_batch_size
+from bayesflow.types import Tensor
 
 
 class NormalSimulator(Simulator):


stefanradev93:
Thanks, Daniel, that's exactly what I had in mind! Can you please fix the linter's complaints regarding unused imports and too long lines (https://github.com/bayesflow-org/bayesflow/actions/runs/21158504580/job/60848059851?pr=619) before merging?

stefanradev93:
Update: Fixed the linter issues, docs, and simplified parts of the code. I guess this is good enough now that we will turn our attention to the graphical simulator.

paul-buerkner:
Sounds Great!

Stefan Radev ***@***.***> schrieb am Di. 20. Jan. 2026 um
21:52:

PR #622, opened 2026-01-25
Add Neural Ratio Approximator

LarsKue:
This is still work in progress. Specifically, currently still missing are:

- [x] Exhaustive Docs
- [x] Tests up to our coverage standards
- [x] An example notebook

I added an example notebook that sufficiently tests the existing functionality. We should still add actual tests to verify the code properly and improve our coverage.

I opted to disable NRE-B for now, since it introduces numerical issues in the generalization due to relying on `gamma = float("inf")`. We should be able to re-enable it either using a special case treatment or some additional tricks (preferred).

I think this is ready for review as-is, since we want to push the feature quickly. However, I am also fine with being more thorough regarding the remaining open points above, if we have time to address them.


stefanradev93:
I have added the following updates to the class (@paul-buerkner, @LarsKue):
- Removed NRE-A / NRE-B, as these are clearly suboptimal according to the paper and empirics (if needed, NRE-A can still be created with `gamma=1` and `K=5`). Removed the factory methods, the approximator is now created in a similar way to other methods.
- Harmonized approximator with the rest of the lib (optional summary network, correct passing of stage, consistent naming of targets and conditions...)
- Changed loss computation to be consistent with the paper. The implementation was missing the bootstrapping of `K` "contrastive batches" and was essentially implementing NRE-A all the way with the weighting of NRE-C. Everything in the new loss computation remains vectorized. Pls, double check.
- See updated notebook. I have also kept `gamma = 1` as the default, since the paper suggests values of approximately 1, but the paper is itself a bit fuzzy on the value. We would have to do some experiments ourselves.

TODO: Actually plug the NRE into a PyMC loop and check if the correct posterior is recovered.

paul-buerkner:
Thank you @stefanradev93! @LarsKue could you take care of the tests coverage today? I will check the notebook, add some docs, and verify the correctness on some example. I also noticed that the adapter is not support in a way as is done in other approximators. I will add this too.

LarsKue:
@stefanradev93 thank you for the review and extensive updates!

In the NRE-C paper, the authors mention just before showing Eq. 9:

> We point out that the final term is symmetric up to permutation of Θ, enabling the replacement of the sum by multiplication with K.

They go on to show a purely weighted version of the loss in Eq. 8 and 9. We now have a version that actually computes K contrastive values instead of weighting. Do you think this works better than the weighted version the authors suggest here?

paul-buerkner:
I have added proper support for an adapter and some docs. 

I also extended the example notebook to contain both direct verification by comparing with the true log likelihood and using pymc for MCMC with the estimated likelihood to evidence ratio. Both checks seem to indicate something is still wrong. log lik - log ratio should be constant since they differ only in the log p(x) term. Also the posterior estimated by pymc shows less than expected posterior shrinkage. It was very well possible of course that I made a mistake in both checks and the results are actually correct. 

@stefanradev93, @LarsKue could you take a look at the updated example notebook? Before merging, I will simplify the notebook again and remove in particular all the pymc related code (unless we really want to have it there)

LarsKue:
Once tests pass, this is ready to merge now :rocket:

PR #624, opened 2026-01-28
Masking diffusion

arrjon:
This pull request introduces new functionality for conditional masking and partial target conditioning in the `flow_matching`, `consistency_model`, `stable_consistency_model`, and `diffusion_model` classes. The main additions are the `drop_cond_prob` parameter (which allows probabilistic masking of conditioning inputs during training, enabling simultaneous training of conditional and unconditional models) and, for diffusion models, the `drop_target_prob` parameter (enabling training and inference with partial targets). The code ensures proper handling of these features during inference and metrics computation.

Diffusion model:
- During inference, pass a `target_mask` and `targets_fixed` to specify which parts of the target should be kept fixed to the values in `targets_fixed`.

All diffusion type models:
- During inference, set `unconditional_mode=True`, to use mask conditions during sampling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ examples/checkpoints/
 build
 docs/
 
+# uv
+uv.lock
 
 # mypy
 .mypy_cache
--- a/tests/test_networks/conftest.py
+++ b/tests/test_networks/conftest.py
@@ -226,6 +226,37 @@ def generative_inference_network(request):
     return request.getfixturevalue(request.param)
 
 
+@pytest.fixture(
+    params=[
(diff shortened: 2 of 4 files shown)


arrjon:
This PR is done, and should not interfere with #633, so can be merged from my side

LarsKue:
@arrjon Is there an easy way we can hit the patch coverage target? If it's too much work, we can merge without for now and put better coverage on our todo list.

arrjon:
I expanded the tests, @LarsKue is this good enough now?

LarsKue:
Great, thanks!

Issue #626, opened 2026-01-28
[FEATURE] Support AtttentionMask in SetTransformer


arrjon:
**Is your feature request related to a problem? Please describe.**
During training, the summary network is invoked as
`summary_metrics, summary_outputs = self._compute_summary_metrics(summary_variables, stage=stage)`
so no `**kwargs` provided to `fit()` are forwarded to the summary network call function. As a result, it’s not possible (or at least not obvious to me) to pass an attention mask during training.

Additionally, in `SetTransformer`, the mask is not propagated through the attention stack; it is only passed to the pooling operation:

```
summary = self.attention_blocks(input_set, training=training)
summary = self.pooling_by_attention(summary, training=training, **kwargs)
```

**Describe the solution you'd like**

1. Add first-class support for an attention mask throughout the summary network during training and inference.
2. Fix `SetTransformer` kwargs propagation.

**Additional context**

This is part of the issue #625, where a minimal example is provided (the compositional part can be ignored for that).

stefanradev93:
I can take care of it!

PR #627, opened 2026-01-29
Attention Refactor (WIP)

stefanradev93:
This PR refactors the transformers module for internal consistency and directly exposes `attention_mask` and `use_causal_mask` in the relevant transformers. The following changes were made:

- Transformer building blocks were moved into a dedicated `attention` module.
- Abstract base class `Transformer` was added to easily tell apart transformer summaries from other summary nets.
- Files were renamed to reflect semantic names.
- `TimeSeriesTransformer` can now act as a many-to-many network (e.g., for modeling time-varying targets)

It also prepares to address #https://github.com/bayesflow-org/bayesflow/issues/626.

@arrjon @paul-buerkner It remains to decide how we want the attention mask passed? Should we search for it in the simulator outputs (as we do for other special args like `sample_weights`)? This would have the advantage that the mask can be constructed very flexibly.

### DIFF
--- a/bayesflow/approximators/continuous_approximator.py
+++ b/bayesflow/approximators/continuous_approximator.py
@@ -281,6 +281,47 @@ class ContinuousApproximator(Approximator):
 
         return inference_variables
 
+    def _prepare_data(
+        self, data: Mapping[str, np.ndarray], log_det_jac: bool = False, **kwargs
+    ) -> dict[str, Tensor] | tuple[dict[str, Tensor], dict[str, Tensor]]:
+        """
+        Adapts, (optionally) standardizes, and converts data to tensors for inference.
+
+        Handles inputs containing only conditions, only inference_variables, or both.
+        Optionally tracks log-determinant Jacobian (ldj) of transformations.
+        """
+        adapted = self.adapter(data, strict=False, log_det_jac=log_det_jac, **kwargs)
+
+        if log_det_jac:
+            data, ldj = adapted
+            ldj_inference = ldj.get("inference_variables", 0.0)
+        else:
+            data = adapted
+            ldj_inference = None
+
+        # Standardize conditions
+        for key in self.CONDITION_KEYS:
+            if key in self.standardize and key in data:
+                data[key] = self.standardize_layers[key](data[key])
+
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Nice! Yes, I would love to search for `attention_mask` in the simulator output! In fact, I cannot really think of an alternative that would be nearly as nice. Is there any disadvantage to doing so?

stefanradev93:
Not any I could think of, except some overhead on the user's side which may be unavoidable. @arrjon if you concur, I will proceed with adding the search for attention mask among simulation outputs.

arrjon:
Sounds good to me as it gives maximal flexibility to the user!

stefanradev93:
I now allowed for `mask` and `attention_mask` keyword arguments in `compute_metrics` of the `approximator`, which are expected to be part of the output (or `None` - default) and be propagated both to the inference and summary network, depending on the signature of each network. 

The case for sampling is tricky tho, as now the user will have to provide the `attention_mask` or `mask` arguments as `**kwargs` instead of passing them as part of the `conditions` dict. This may pose some confusion, but expecting the masks to be part of the conditions invalidates our internal logic...

paul-buerkner:
Can you elaborate why we cannot use `conditions` for this purpose?

stefanradev93:
I can solve it with a bunch of checks for now.

paul-buerkner:
I don’t fully understand what you mean by this. Solve what? Making it part
of conditions?

Stefan Radev ***@***.***> schrieb am Di. 3. Feb. 2026 um
14:56:

stefanradev93:
@arrjon Can you please check if my latest commit enables the functionality you needed?

arrjon:
Thanks @stefanradev93, now the masking seems to work!

stefanradev93:
@LarsKue @arrjon  I would actually like to discard that PR in favor of the approximator refactor https://github.com/bayesflow-org/bayesflow/pull/633, which would make mask handling much easier. Thus, I would only keep the attention module refactor from this branch/PR.

LarsKue:
@stefanradev93 so this can be closed?

Issue #623, opened 2026-01-27
PointApproximator and PointInfernceNetwork are suboptimally named


han-ol:
`PointApproximator` and `PointInfernceNetwork` are really not limited to estimating point summaries of continuous quantities.

1. Approximation by a parametric distribution is not generally understood as point estimation.
2. Classification is not usually understood as point estimation.

Yet, both can be done with `PointInfernceNetwork` and the name is the only thing that makes this awkward.
The model comparison classifiers should be `PointInferenceNetwork`s but currently they are independent reimplementations. Especially, when we implement more losses for model comparison it would be prudent to include them in `bayesflow.scores`, since they too are (typically strictly proper) scoring rules.

**A more descriptive name might be `ScoringRuleApproximator`.**

I ruled out `ScoreApproximator` because score is already sufficiently overloaded in both statistics and ML.
I ruled out `ScoringRuleFeedForwardApproximator` for being too long.

Please do comment any name preferences you might have.

stefanradev93:
I agree with the naming `ScoringRuleApproximator`. However, it will imply a deeper redesign of how we currently handle point estimators / model comparison. To reiterate your idea:
1. Add all scoring rules and upcoming model comparison losses to the `scores` module.
2. Additionally, complete revamp the approximator.

I would like to hear your ideas regarding 2., as this will be the crucial entry point for the new changes,

han-ol:
I discussed with @paul-buerkner about this and we thought that while it would be sensible to do the rename, we can add a (sugar) `PointInferenceNetwork` for easy access to the the point estimation aspect of `ScoringRuleInferenceNetwork` and that would have an interface like this:
```python
inference_network = PointInferenceNetwork(["mean", "quantiles"], quantile_levels=[0.1,0.5,0.9])
```
which would be equivalent to
```python
inference_network = ScoringRuleInferenceNetwork({"mean": bf.scores.MeanScore(), "quantiles": bf.scores.QuantileScore(q=[0.1,0.5,0.9])})
```

This way we do not loose the intuitive aspect for users. It can be weird to have to introduce how empirical risk minimization with scoring rules leads to point estimates, just to get the simplest mean posterior approximation.
At the same time, the wider applicability of `ScoringRuleInferenceNetwork` isn't an easter egg anymore and we can use it for model comparison and the like.

[...]

PR #628, opened 2026-02-02
Rename PointXYZ to ScoringRuleXYZ

han-ol:
Addresses #623

### DIFF
--- a/bayesflow/experimental/graphical_approximator/graphical_approximator.py
+++ b/bayesflow/experimental/graphical_approximator/graphical_approximator.py
@@ -295,7 +295,7 @@ class GraphicalApproximator(Approximator):
             import tensorflow as tf
 
             if "simulator" in kwargs:
-                ds = GraphicalApproximator.build_dataset(**filter_kwargs(kwargs, GraphicalApproximator.build_dataset))
+                ds = self.build_dataset(**filter_kwargs(kwargs, self.build_dataset))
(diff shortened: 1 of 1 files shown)


arrjon:
Thanks @han-ol! Since you are working on the scoring rules anyway, I think it would be good to rename the module `scores` to something else like `scoring_rules`. People working on diffusion type models might assume they find there something related to the score of the diffusion model.

han-ol:
@arrjon that makes sense and I can get behind that.
The original thinking was to avoid an underscore and lengthy name in bayesflow modules, which we so far were able to avoid.
Tagging @paul-buerkner too.

arrjon:
Maybe `scoring` then instead of `scores`?

han-ol:
I went with scoring_rules after discussing with Paul and changed the names of ALL score classes too.
Instead of `MeanScore`, it is now `MeanScoringRule`. Also instead of `MultivariateNormalScore` it is `MvNormalScoringRule`.

han-ol:
Update: Keeping the pattern for individual score classes `<EstimateType>Score` instead of `<EstimateType>ScoringRule`. 

Importing from module `scoring_rules` and passing the score objects to `ScoringRuleNetwork` already makes it abundantly clear that these are scoring rules and calling them scores is more in line with the literature, where a concrete scoring rule is often called score, e.g. interval score. No need to add "Rule" to all class names.

PR #656, opened 2026-03-17
[Feature] Add Latent Diffusion and generic latent inference wrapper for all inference networks.

bhanuprasanna2001:
- Included the use of any inference network as part of the latent diffusion model.
- Only remaining task is to update the model's name to a more suitable one.
- Developed a notebook experiment with two moons to test all inference networks and observe their behavior in the latent space.
- The experiment is functioning well.


bhanuprasanna2001 (commented):
- Added Latent Inference Network, which can be used with any Inference Network to operate in the latent space.
- Added Latent Diffusion Model as a thin wrapper for the Latent Inference Network.

LarsKue:
@stefanradev93 I refactored this PR, including more rigorous testing for the network. Please review.
@paul-buerkner I also improved the user-interface a bit, the current method to instantiate the network is shown below. Please give feedback and ideas for improvement if you have any.

```py3
import bayesflow as bf

latent_diffusion = bf.experimental.LatentIN(
    autoencoder=bf.experimental.VariationalAutoEncoder(
        latent_dim=64, encoder_network=bf.networks.MLP(), decoder_network=bf.networks.MLP()
    ),
    inference_network=bf.networks.DiffusionModel(subnet=bf.networks.TimeMLP()),
)
```

Another smaller but in my opinion very good change is that I adjusted the way we show nested metrics in the progress bar. Instead of adding a plain "inference" to the metric name like so:
```py3
inference/inference_loss=0.10067
```
We now use specific network class names to show where the loss is coming from:
```py3
LatentIN/VariationalAutoEncoder/loss=0.10067
```

Some networks may still be doing the former, so there may be some additional effort required to migrate all of the networks to the latter case.

paul-buerkner:
Thank you! I will review it in more detail next week. One quick comment already: Please rename `LatentIN` to `LatentInferenceNetwork` to match our general naming conventions.

paul-buerkner (changes requested):
Just checked superficial aspects of this PR right now.

Two comments:
- Please rename `IN` to `InferenceNetwork` and also rename files accordingly: `latent_in` to `latent_inference_network` 
- Even for the experimental module, the PR should have minimal amount of docs. Right now I don't see any, I think.

stefanradev93:
Did the rename and made some simple changes.

The VAE currently is the basic non-information preserving model. Shouldn't we follow the insights from:

https://arxiv.org/abs/1706.02262

And implement a more general InfoVAE?

PS. There are some numerical issues with the tests.

LarsKue:
@stefanradev93 I agree that we can use other VAE implementations on top of the current one. However, this could be done in a separate PR specifically for the `InfoVAE`. I will add a special-case for the numerical problems in the tests.

@paul-buerkner I will add docs before we merge.

stefanradev93:
I have added the `InfoVAE` class, which I think can safely replace the current vanilla VAE given that it's a superset of VAEs. Please, check it out.

stefanradev93:
I believe this PR is mergeable upon fixing the numerical issues flagged by the test. I've also noticed that the following deprecation warning is incorrect:

`DeprecationWarning: vjp is deprecated; we are working on moving these utilities upstream or into their own module with improved signatures.`

The utilities have been moved but not yet integrated throughout the codebase.

LarsKue:
@stefanradev93 In my tests the deprecation warnings only appeared when the functions are called directly by the user, not if called by the library. Did this work differently for you?

I will push my fix for the numerical test asap.

stefanradev93 (approved):
The PR is ready to be merged.

paul-buerkner (approved):

PR #661, opened 2026-03-24
fix unconditional sampling

arrjon:
These minor changes allow for unconditional sampling (i.e., sampling without providing conditioning variables). The changes make the `conditions` argument optional in relevant methods and update the code to handle cases where `conditions` is `None`. Additionally, a new test verifies that unconditional sampling works as expected.


stefanradev93:
I shifted the responsibility to `_prepare_conditions` in the base class.

PR #663, opened 2026-03-25
Improve docstrings in workflows and order of methods in BasicWorkflow

han-ol:
`EnsembleWorkflow` did not have its own docstring which led to misleading docs.

`BasicWorkflow` had suboptimal order of methods which led to risking disorienting users.

workflows module had no example.

These issues are addressed by this PR.


stefanradev93:
Excellent additions, Hans!

Issue #666, opened 2026-04-08
[FEATURE] Control randomness of sample methods (optional seed)


han-ol:
**Is your feature request related to a problem? Please describe.**
I want to persist a dictionary of samples to disk in an analysis pipeline for later use with reproducible randomness (fixed seed).
The main benefit I would get from this is that I can later check whether a deserialized approximator has changed: If samples are identical to the what was saved to disk, the approximator has not changed. If sample is stochastic, you need to use statistical tests to check if distributions are identical.

**Describe the solution you'd like**
Provide a public API for specifying a seed.
```python
workflow/approximator/inference_network/scoring_rule.sample(..., seed: keras.random.SeedGenerator | int | None = None)
```
These get passed on to internal stochastic components (like `DiagonalNormal`, used as `inference_network.base_distribution` AND a stochastic `_inverse` method). If seed=None (default) we recover the current behavior: components rely on their own `self.seed_generator` (https://keras.io/api/random/seed_generator/). Passing an integer seed to the sample method should create a locally scoped new seed generator that is shared by all of the components during the sample call.

**Describe alternatives you've considered**
If we do not provide public API, users can check the source code and rely on internals to get reproducible samples e.g. kind of like this 

```python
def _reset_model_seed_generators(model, seed: int) -> None:                                                                                                                 
    rng = np.random.default_rng(seed)                                                                                                                                       
    for layer in model._flatten_layers(include_self=True):                                                                                                                  
        for attr in vars(layer).values():                                                                                                                                   
            if isinstance(attr, keras.random.SeedGenerator):                                                                                                                
                derived = int(rng.integers(0, 2**31))                                                                                                                       
                attr.state.assign([derived, 0])
```

han-ol:
The issue is resolved by #668.

PR #668, opened 2026-04-09
Sample method determinism

han-ol:
This PR addresses issue #666.

Essentially, it extends the seed argument that already exists in all `keras.random` functions (like https://github.com/keras-team/keras/blob/v3.14.0/keras/src/random/random.py#L5) all the way up to the `workflow.sample` method in a way that makes sure that randomness in all different components of high-level objects (like workflow) stays uncorrelated.

Moreover: ports `distribution.Mixture` randomness from numpy to keras, adds a `keras_utils.keras_multinomial` convenience function and fixes a bug regarding correct handling of `batch_shape` between `ScoringRuleApproximator` and `MixtureScore`.

To confirm that samples for the same seed are identical, corresponding integration tests with 

* an `EnsembleApproximator` with a CouplingFlow and a DiffusionModel approximator and 
* a `ScoringRuleApproximator` with both a standalone `MvNormalScore` and a `MixtureScore` of two `MvNormalScore`s 

are passing.

to do:

* [x] The tests included in the initial commits are still "work in progress" and should be improved to run faster (especially `DiffusionModel` can be wasteful to sample) and integrate well with the existing fixtures. (The `DiffusionModel` tests were only performed locally at this point)
* [x] `workflow.sample`

### DIFF
--- a/bayesflow/approximators/continuous_approximator.py
+++ b/bayesflow/approximators/continuous_approximator.py
@@ -234,7 +234,7 @@ class ContinuousApproximator(Approximator):
         seed : int, keras.random.SeedGenerator, or None, optional
             Seed for reproducible sampling. An integer is converted to a ``keras.random.SeedGenerator``
             and shared across all stochastic operations in the call. A ``SeedGenerator`` is passed through
-            as-is. If ``None`` (default), each component uses its own instance seed generator or a global one.
+            as-is. If ``None`` (default), each component uses its own instance seed generator.
         **kwargs : dict
             Additional keyword arguments for the sampling process.
 
--- a/bayesflow/approximators/continuous_approximator.py
+++ b/bayesflow/approximators/continuous_approximator.py
@@ -9,6 +9,7 @@ from bayesflow.adapters import Adapter
 from bayesflow.networks import InferenceNetwork, SummaryNetwork
 from bayesflow.types import Tensor
 from bayesflow.utils import split_arrays
+from bayesflow.utils.keras_utils import resolve_seed
 from bayesflow.utils.serialization import serialize, serializable
 
 from .approximator import Approximator
@@ -200,6 +201,7 @@ class ContinuousApproximator(Approximator):
         batch_size: int | None = None,
         sample_shape: Literal["infer"] | Tuple[int] | int = "infer",
         return_summaries: bool = False,
+        seed: int | keras.random.SeedGenerator | None = None,
         **kwargs,
     ) -> dict[str, np.ndarray]:
         """
@@ -229,6 +231,10 @@ class ContinuousApproximator(Approximator):
         return_summaries: bool, optional
             If set to True and a summary network is present, will return the learned summary statistics for
             the provided conditions.
+        seed : int, keras.random.SeedGenerator, or None, optional
+            Seed for reproducible sampling. An integer is converted to a ``keras.random.SeedGenerator``
+            and shared across all stochastic operations in the call. A ``SeedGenerator`` is passed through
+            as-is. If ``None`` (default), each component uses its own instance seed generator or a global one.
         **kwargs : dict
             Additional keyword arguments for the sampling process.
 
@@ -248,6 +254,7 @@ class ContinuousApproximator(Approximator):
             conditions=resolved_conditions,
             batch_size=batch_size,
             sample_shape=sample_shape,
+            seed=resolve_seed(seed),
             **inference_kwargs,
         )
 
--- a/bayesflow/approximators/ensemble_approximator.py
+++ b/bayesflow/approximators/ensemble_approximator.py
@@ -10,6 +10,7 @@ from bayesflow.adapters import Adapter
 from bayesflow.simulators import Simulator
 from bayesflow.types import Tensor
 from bayesflow.utils import logging, filter_kwargs
+from bayesflow.utils.keras_utils import resolve_seed, keras_multinomial
 from bayesflow.utils.serialization import serializable, serialize
 from bayesflow.datasets import EnsembleDataset
 
@@ -230,6 +231,7 @@ class EnsembleApproximator(Approximator):
         split: bool = False,
         member_weights: Mapping[str, float] | None = None,
         merge_members: bool = True,
+        seed: int | keras.random.SeedGenerator | None = None,
         **kwargs,
     ) -> dict[str, np.ndarray]:
         """
@@ -268,23 +270,27 @@ class EnsembleApproximator(Approximator):
                 fn=lambda name, a: a.sample(num_samples=num_samples, conditions=conditions, split=split, **kwargs),
             )
 
+        seed_generator = resolve_seed(seed)
+
         weights = self._resolve_member_weights(member_weights)
         names = tuple(weights.keys())
-        probs = np.fromiter(weights.values(), dtype=float, count=len(weights))
+        probs = np.array(list(weights.values()))
 
-        counts = np.random.multinomial(num_samples, probs)
+        counts = keras_multinomial(num_samples, probs, seed=seed_generator)
         alloc = {name: int(count) for name, count in zip(names, counts) if count > 0}
 
         per_member = self._map_members(
             list(alloc.keys()),
             capability="distribution",
-            fn=lambda name, a: a.sample(num_samples=alloc[name], conditions=conditions, split=split, **kwargs),
+            fn=lambda name, a: a.sample(
+                num_samples=alloc[name], conditions=conditions, split=split, seed=seed_generator, **kwargs
+            ),
         )
 
         merged = keras.tree.map_structure(lambda *xs: np.concatenate(xs, axis=1), *list(per_member.values()))
-        idx = np.random.permutation(num_samples)
+        shuffle_idx = keras.random.shuffle(keras.ops.arange(num_samples), seed=seed_generator)
 
-        return keras.tree.map_structure(lambda a: np.take(a, idx, axis=1), merged)
+        return keras.tree.map_structure(lambda a: np.take(a, shuffle_idx, axis=1), merged)
 
     def log_prob(
         self,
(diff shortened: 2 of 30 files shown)


han-ol:
Now, includes fast tests for all inference networks. Ready for review! 

I used a kwargs.pop pattern in _inverse methods. We can also explicitly add seed to the arguments, if you prefer.

bayesflow/approximators/continuous_approximator.py
```diff
@@ -229,6 +231,10 @@ def sample(
         return_summaries: bool, optional
             If set to True and a summary network is present, will return the learned summary statistics for
             the provided conditions.
+        seed : int, keras.random.SeedGenerator, or None, optional
+            Seed for reproducible sampling. An integer is converted to a ``keras.random.SeedGenerator``
+            and shared across all stochastic operations in the call. A ``SeedGenerator`` is passed through
+            as-is. If ``None`` (default), each component uses its own instance seed generator or a global one.
```
LarsKue: The doc is unspecific here, is a single global seed generator used or are multiple used? By default, I believe keras uses a single global one.

bayesflow/approximators/scoring_rule_approximator.py
```diff
@@ -62,8 +63,8 @@ def __init__(
 
         self.distribution_keys = []
         for score_key, score in self.inference_network.scoring_rules.items():
-            has_sample = callable(getattr(score, "sample", None))
-            has_log_prob = callable(getattr(score, "log_prob", None))
+            has_sample = getattr(score, "sample", None)
+            has_log_prob = getattr(score, "log_prob", None)
```
LarsKue: Why should these `has_*` variables not be boolean?

bayesflow/utils/keras_utils.py
```diff
@@ -1,9 +1,29 @@
...
+def resolve_seed(seed):
+    """Convert an integer seed to a SeedGenerator; pass a SeedGenerator or None through unchanged."""
+    if isinstance(seed, int):
+        return keras.random.SeedGenerator(seed)
+    return seed
+
+
+def keras_multinomial(
+    num_samples: int, probs: Sequence[float], *, seed: int | keras.random.SeedGenerator | None = None
+):
```
LarsKue: needs doc

bayesflow/approximators/ensemble_approximator.py
```diff
@@ -268,23 +270,27 @@ def sample(
...
 
+        seed_generator = resolve_seed(seed)
+
         weights = self._resolve_member_weights(member_weights)
         names = tuple(weights.keys())
-        probs = np.fromiter(weights.values(), dtype=float, count=len(weights))
+        probs = np.array(list(weights.values()))
 
-        counts = np.random.multinomial(num_samples, probs)
+        counts = keras_multinomial(num_samples, probs, seed=seed_generator)
```
LarsKue: This is called only in two places, and I don't see the need for the utility function here yet. I think the code would be more explicit, if we remove the utility function and repeat ourselves a little bit instead (notably not following DRY).

bayesflow/distributions/diagonal_student_t.py
```diff
@@ -111,18 +112,19 @@ def log_prob(self, samples: Tensor, *, normalize: bool = True) -> Tensor:
         return result
 
     @allow_batch_size
-    def sample(self, batch_shape: Shape) -> Tensor:
+    def sample(self, batch_shape: Shape, seed: int | keras.random.SeedGenerator | None = None) -> Tensor:
```
LarsKue, line 115: Needs doc, particularly for the seed override. If doc is inherited from the parent class (which I am not sure of right now), then we can get away with a reference to such here, but there should be some reference to the actual doc in the code.

bayesflow/networks/inference/scoring/scoring_rules/mixture_score.py
```diff
@@ -201,32 +206,30 @@ def sample(self, batch_shape: Shape, **estimates: Tensor) -> Tensor:
         Tensor
             Samples with shape (batch_size, num_samples, ...).
         """
-        if len(batch_shape) != 2:
-            raise ValueError("`batch_shape` must be a tuple (batch_size, num_samples).")
+        assert len(batch_shape) == 1  # approximator.helpers.Sampler makes sure that batch_shape is flat
```
LarsKue: Prefer raising exceptions with error messages over plain `assert`

bayesflow/workflows/basic_workflow.py
```diff
@@ -213,446 +213,141 @@ def default_adapter(
...
-            Additional keyword arguments passed to the underlying `_fit` method.
+            If set to `"infer"` (default), the structural dimensions are inferred from the `inference_conditions`.
+            In that case, all non-vector dimensions except the last (channel) dimension are treated as structural
+            dimensions. For example, if the final `inference_conditions` have shape `(batch_size, time, channels)`,
+            then `sample_shape` is inferred as `(time,)`, and the generated samples will have shape
+            `(num_conditions, num_samples, time, target_dim)`.
+        seed : int, keras.random.SeedGenerator, or None, optional
+            Seed for reproducible sampling. An integer is converted to a ``keras.random.SeedGenerator``
+            and shared across all stochastic operations in the call. A ``SeedGenerator`` is passed through
+            as-is. If ``None`` (default), each component uses its own instance seed generator or a global one.
```
LarsKue: Unspecific, see my other comment

tests/test_approximators/test_ensemble_approximator/test_sample.py
```diff
@@ -20,3 +20,16 @@ def test_approximator_sample(ensemble_approximator, simulator, batch_size, adapt
 
     for samples_value in samples.values():
         assert isinstance(samples_value, np.ndarray)
+
+    samples_seed42_1 = ensemble_approximator.sample(num_samples=2, conditions=data, seed=42)
+    samples_seed42_2 = ensemble_approximator.sample(num_samples=2, conditions=data, seed=42)
+
+    for key in samples.keys():
+        assert np.allclose(
```
LarsKue: use `np.testing.assert_allclose` for better diagnostics

tests/test_approximators/test_ensemble_approximator/test_sample.py
```diff
@@ -20,3 +20,16 @@ def test_approximator_sample(ensemble_approximator, simulator, batch_size, adapt
...
+            samples_seed42_1[key],
+            samples_seed42_2[key],
+        ), "samples differ for identical seed"
+        assert not np.allclose(
```
LarsKue: If we also want to use `np.testing.assert_allclose` here, we could do something like:

```py3
with pytest.raises(AssertionError):
    np.testing.assert_allclose(x, y)
```

tests/test_approximators/test_scoring_rule_approximator/test_sample.py
```diff
@@ -47,3 +47,16 @@ def test_sample(scoring_rule_approximator_any, simulator, batch_size, num_sample
...
+
+    for key in samples_merged.keys():
+        assert np.allclose(
+            samples_seed42_1[key],
+            samples_seed42_2[key],
+        ), f"{key}: samples differ for identical seed"
+        assert not np.allclose(
+            samples_seed42_1[key],
+            samples_merged[key],
+        ), f"{key}: samples do not differ in unseeded case"
```
LarsKue: see my other comment

tests/test_networks/test_sample.py
```diff
@@ -0,0 +1,46 @@
...
+    def leaves(x):
+        return keras.tree.flatten(x)
+
+    for s1, s2, s_unseeded in zip(leaves(samples_seed42_1), leaves(samples_seed42_2), leaves(samples_no_seed)):
+        arr1 = keras.ops.convert_to_numpy(s1)
+        arr2 = keras.ops.convert_to_numpy(s2)
+        arr_unseeded = keras.ops.convert_to_numpy(s_unseeded)
+
+        assert np.allclose(arr1, arr2), f"samples differ for identical seed ({inference_network})"
+        assert not np.allclose(arr1, arr_unseeded), f"seeded and unseeded samples are identical ({inference_network})"
```
LarsKue: see my other comment

LarsKue (changes requested):
Thank you for these comprehensive and progressive changes! I noted a few things I would like to see improved, but overall this is looking great already.

bayesflow/distributions/diagonal_student_t.py
han-ol: Can confirm that docstrings are inherited, both for help text and on sphinx docs (e.g. here: https://bayesflow.org/v2.0.10/api/bayesflow.distributions.DiagonalNormal.html#bayesflow.distributions.DiagonalNormal.get_config).

In this case do you think a reference (e.g. by comment) is necessary? If so, how would it look like?

bayesflow/networks/inference/diffusion/diffusion_model.py
```diff
@@ -700,7 +702,7 @@ def score_fn(time, xz):
                 score_fn=score_fn,
                 noise_schedule=self.noise_schedule,
                 state=state,
-                seed=self.seed_generator,
+                seed=seed or self.seed_generator,
```
han-ol: will move the or case to definition of seed (so pattern is exactly as in other networks)

bayesflow/approximators/scoring_rule_approximator.py
han-ol: no reason. Will revert removing callable.

bayesflow/approximators/continuous_approximator.py
han-ol: This just depends on what the components end up doing. Would you be in favor of removing the "or a global one"? Then we are making a normative statement to always use instance seed generators. Currently, it is descriptive, as in describing what can happen here.

I am going with the normative proposal (removing "or a global one") now, let me know if you would prefer it differently. Else, feel free to resolve.

bayesflow/workflows/basic_workflow.py
han-ol: see answer in other comment

han-ol:
Thank you for your review! If you agree with how I addressed the comments and the tests are still passing this is ready for merging.

bayesflow/distributions/diagonal_student_t.py
LarsKue: Something like `See :py:class:...`, super simple

LarsKue (approved):
LGTM!

PR #670, opened 2026-04-17
Lre connect hssm v2

AlexanderFengler:
This pull request reintroduces the PyMC-HSSM-BayesFlow bridge.  Starting with the three notebooks, @stefanradev93 and I will work on creating a set of suitable wrappers to allow easy initial integration.


stefanradev93:
@AlexanderFengler I have created the `wrappers.pymc` module. The intended use is for both exchangeable likelihoods (`exchangeable=True`) where the parameters are broadcasted to n replicas and arbitrary `(param, data)` pairs. Any feedback is greatly appreciated.

stefanradev93:
@AlexanderFengler At this point we have a general NLE / NRE wrapper for PyMC. Please, take a detailed look at my wrappers as they may still be missing some nitty-gritty details regarding PyMC internals. Feel free to adapt the notebook as you see fit as well. I've kept the _af notebook for a quick reference before merging.

@paul-buerkner At this point, I would also like to get your feedback on the state of `Ratio_Estimation.ipynb` and the new interface of the wrappers / PyMC bridge. Turns out, it was pretty easy to unify the design and be agnostic of whether a Ratio or a Continuous approximator becomes a PyMC distro.

paul-buerkner:
Feedback given in a call. Looks very cool!

AlexAndorra:
Just chiming in to say this is 🔥 !!

AlexanderFengler:
> @AlexanderFengler At this point we have a general NLE / NRE wrapper for PyMC. Please, take a detailed look at my wrappers as they may still be missing some nitty-gritty details regarding PyMC internals. Feel free to adapt the notebook as you see fit as well. I've kept the _af notebook for a quick reference before merging.
> 
> @paul-buerkner At this point, I would also like to get your feedback on the state of `Ratio_Estimation.ipynb` and the new interface of the wrappers / PyMC bridge. Turns out, it was pretty easy to unify the design and be agnostic of whether a Ratio or a Continuous approximator becomes a PyMC distro.

@stefanradev93 sorry for the slight delay here. I'll be back from a wedding tomorrow and look at this properly before EOD.

AlexanderFengler (commented):
The patterns for random variable construction are close to what is happening in HSSM, the exchangeable notion makes sense and I couldn't see any obvious problems seeing through the code. I'll look through the notebooks and leave another message.

AlexanderFengler:
@stefanradev93 @LarsKue this is ready afaic.

stefanradev93:
Ok, great! The only remaining question for me would be which backend and PyMC config do we want to show in the NLE/NRE notebooks. For example, I picked this relatively vanilla setup:

```
trace_ddm = pm.sample(
    draws=1000, 
    tune=1000,
    nuts_sampler="pymc",
    chains=4, 
    cores=1,  # use cores=4 in a "spawn context" (will fork on Linux and die)
    random_seed=SEED,
    initvals={"v": 0.0, "a": 1.0, "z": 0.5, "t": 0.3}
)
```

to avoid all fork Shenanigans on Linux with `cores>1`, but we may want to use a more robust config for Mac, Windows, and Linux which uses something close to "recommended defaults for best performance". Perhaps @AlexAndorra can chip in as well.

AlexanderFengler:
@stefanradev93 one note on the gaussian mixture example you use to illustrate MCMC btw. 

If you look closely (correct me if I saw it wrong), you have four chains and it turns out that two sample from one mode the other two from the other mode. 

If the example is not about 'problems with MCMC' (and I think that's an interesting point in it's own right ofc, that, if a posterior is multimodal, posterior amortization has some nice advantages, because it's not vulnerable to the same geometric constraints that MCMC structurally is), I would actually choose one with a simpler posterior.

It doesn't directly hurt anyone as it is currently, stuff is working 'as expected', but maybe not the most didactically helpful example in this particular context.

---

Re best practice on the PyMC front, I would come down on leaning on defaults as much as possible to simplify on the BayesFlow side. Possibly just a link to PyMC resources on backends (nutpie / numpyro samplers, blackjax, new numba backend for pymc sampler etc.). Afaic there isn't a single best way for all cases, not worth worrying about here.

Can prob get away with only setting the spawn context explicitly, because of the known failure, and possibly the initvals.

stefanradev93:
Agreed about the example. I changed it completely to use a correlated g-and-k model with N observations, slightly more interesting than the GMM. What do you think? (@paul-buerkner ?)

paul-buerkner:
Examples and user-interface look very good! Just one minor note: in the likelihood case study we use base student-t distribution for the g-and-k family but with df=30, which is bascially normal. If you want/need to use student-t why not lower df, or why not change to normal (default) if we use df=30 anyway?

stefanradev93:
@paul-buerkner I only left a hint that it can be used; I previously had a more extreme parameterization of the distribution which stabilized training with student-t, but it seems no longer mandatory for the milder parameterization. This is ready to be merged.

PR #671, opened 2026-04-19
Sanitize Couplings and MLPs

stefanradev93:
This PR adds proper build logic to coupling flows that does not rely on calling the layers internally. It also makes the following changes:
- Removes spectral normalization (was broken since 2.0.8 and leads to instabilities)
- Adds default settings for coupling flows
- Adds orthogonal initialization to default MLPs 
- Breaks the dependency of MLPs on custom `Sequential` and `Residual` layers in preparation of removing boilerplate classes. This also removes redundant projector layers between blocks.
- Encapsulates common layer logic in `blocks` and aligns use for `MLP` and `TimeMLP`. 
- Improves docs in a few places and adds a few new tests for couplings


bayesflow/networks/subnets/mlp/mlp.py
```diff
@@ -1,48 +1,38 @@
...
-    Implements a flexible multi-layer perceptron (MLP) with optional residual connections, dropout, and
-    spectral normalization.
-
-    This MLP can be used as a general-purpose feature extractor or function approximator, supporting configurable
-    depth, width, activation functions, and weight initializations. If used in conjunction with a coupling net,
-    a diffusion model, or a flow matching model, it assumes that the input and conditions are already concatenated.
-
-    If `residual` is enabled, each layer includes a skip connection for improved gradient flow [1]. The model also
-    supports dropout for regularization and spectral normalization for stability in learning smooth functions.
+class MLP(keras.Layer):
```
LarsKue, line 13: Do these changes break serialization? It might be worth it to implement some backwards compatibility with a migration.

tests/test_networks/test_coupling_flow/test_coupling_flow.py
```diff
@@ -0,0 +1,154 @@
+import keras
+import numpy as np
+
+from bayesflow.utils.serialization import serialize, deserialize
+from tests.utils import assert_layers_equal, assert_allclose
+
+
+# ---- Build -----------------------------------------------------------------
```
LarsKue, line 8: Nit: I'm not usually a fan of these separating comments

tests/test_networks/test_coupling_flow/test_coupling_flow.py
```diff
@@ -0,0 +1,154 @@
...
+    assert keras.ops.shape(x) == keras.ops.shape(random_samples)
+
+
+def test_inverse_density_output_shape(coupling_flow, random_samples, random_conditions):
+    xz_shape = keras.ops.shape(random_samples)
+    cond_shape = keras.ops.shape(random_conditions) if random_conditions is not None else None
+    coupling_flow.build(xz_shape, conditions_shape=cond_shape)
+
+    z = keras.random.normal(keras.ops.shape(random_samples))
+    x, log_density = coupling_flow(z, conditions=random_conditions, inverse=True, density=True)
```
LarsKue, line 59: Side note: Do we even want to support inverse density? I think there is no use case for it, and it might even be confusing for users (even if it is implemented correctly)

LarsKue (approved):
LGTM! Thank you for modernizing this module. I left some minor comments, please address before merging.

tests/test_networks/test_coupling_flow/test_coupling_flow.py
stefanradev93: Agreed, can you get rid of them? Claude loves them.

tests/test_networks/test_coupling_flow/test_coupling_flow.py
stefanradev93: It could be a bit confusing looking at it internally, but I like the symmetry and users will never experiences it. It could be useful for checking consistency tho.

bayesflow/networks/subnets/mlp/mlp.py
stefanradev93: Yes, but we can only guarantee backward compatibility starting with `2.1.x`.

PR #673, opened 2026-04-22
Enable sample passing to diagnostic methods in workflow

stefanradev93:
Calling the below very common pattern is wasteful as it samples **twice** internally, but we are tacitly recommending it by using it across tutorials:

```
figs = workflow.plot_default_diagnositcs(...)
metrics = workflow.compute_default_diagnostics(...)
```

This PR introduces an option to directly pass samples to the interfaces like so:

```
samples = workflow.sample(test_data)
figs = workflow.plot_default_diagnositcs(test_data, samples=samples)
metrics = workflow.compute_default_diagnostics(test_data, samples=samples)
```

This provides a more efficient way to deal with multiple diagnostic calls AND has the advantage of keeping the samples around. If approved, I would also modify the tutorials to use this pattern as it should be default. The old functionality is kept. There is a guardrail if the user provides samples but no origin test data:

```
figs = workflow.plot_default_diagnositcs(test_data=300, samples=samples) # throws ValueError
```


bayesflow/workflows/basic_workflow.py
```diff
@@ -1102,18 +1115,24 @@ def compute_custom_diagnostics(
...
-        return self.compute_default_diagnostics(**kwargs)
+    def _prepare_for_diagnostics(
+        self,
+        test_data: Mapping[str, np.ndarray] | int,
+        num_samples: int = 1000,
+        samples: Mapping[str, np.ndarray] = None,
+        **kwargs,
+    ):
+        if samples is not None:
+            if isinstance(test_data, int) and samples:
```
paul-buerkner: why whck for "samples" here again after already verifying that samples is not None?

paul-buerkner (approved):
Looks good. Just one minor change suggestions. You can merge after the change.

bayesflow/workflows/basic_workflow.py
stefanradev93: Good catch, that was a leftover from the its previous home!

PR #675, opened 2026-04-24
[Feature] Add backend-agnostic autodiff utilities

LarsKue:
# Features
This PR:
- [x] adds a new module: `bayesflow._backend`;
- [x] adds functional autodiff utilities:
  - [x] `grad`,
  - [x] `value_and_grad`,
  - [x] `jvp`,
  - [x] `vjp`,
  - [x] `jacfwd`,
  - [x] `jacrev`;
- [x] adds a backend-agnostic `jit`;
- [ ] adds a backend-agnostic, functional `vmap`;
- [x] adds extensive tests for all of the above;
- [x] cleans up remaining autodiff utilities from the `utils` module;
- [x] cleans up internals of `FlowMatching` that relied on old autodiff code.

# Known Issues
- Under `torch` backend, `jacrev` and `vjp` are not compatible with `jit` compilation. This likely results from a bad interaction between `keras` and `torch`. The resulting error is `RuntimeError: Cannot access data pointer of Tensor that doesn't have storage`.

# Additional Points of Discussion
- I opted to make the new module protected since we really want to have these utilities in `keras` instead of `bayesflow`. Nevertheless, I think it makes sense to immediately start supporting them even before changes are proposed or merged in `keras`.
- `jit` currently does not support any arguments other than the function. We could try to support additional arguments, but the signatures differ strongly between backends, so we would give up being backend-agnostic if we do.


stefanradev93:
Ok, I've fixed the tf input-output inconsistencies and added additional tests. Our inference network also no longer require any nested logic tested for backends. I've kept the `jacobian` utils as a thin wrapper around the `_backend` for now because of some specific input - output massaging. However, I am happy to also delegate this fully to the backend ops, in which case they would need a bit more controllability. Please, check it out.

PR #688, opened 2026-05-04
Guidance, masking and GLASS sampler

arrjon:
This pull request focuses on more flexible guidance handling, enhanced masking logic for sampling with diffusion type models, and adds the GLASS-flow sampler. 

* Refactored the guidance mechanism in `DiffusionModel` by introducing a new `guidance_function` method, replacing the previous use of `guidance_constraints` and `guidance_function` arguments. This enables easier customization and extension of guidance logic. The main `score` and `compositional_score` methods now use `guidance_kwargs` and a `check_for_guidance` flag for more flexible and consistent guidance application.
* Improved the masking logic in the sampler helper by introducing a `masking_names` parameter, ensuring that only the appropriate tensors are repeated and flattened during sampling, and feature-level masks (like `target_mask`) are handled correctly. Now enables batched masking for unconditional and conditional diffusion like models.
* Added support for the `glass` sampling method (https://arxiv.org/html/2509.25170v1), allowing to use stochastic integration also with a flow matching model. This sampler can be used to switch between deterministic and stochastic sampling easily, and might be especially helpful when applied in the context of guidance of the flow machting or diffusion model.

### DIFF
--- a/bayesflow/approximators/helpers/samplers.py
+++ b/bayesflow/approximators/helpers/samplers.py
@@ -76,6 +76,7 @@ class Sampler:
         batch_size: int | None = None,
         sample_shape: Literal["infer"] | Sequence[int] | int = "infer",
         seed: int | keras.random.SeedGenerator | None = None,
+        masking_names: Sequence[str] = ("target_mask", "targets_fixed"),
         **kwargs,
     ):
         if conditions is None:
@@ -85,6 +86,7 @@ class Sampler:
                 conditions=None,
                 sample_shape=sample_shape,
                 seed=seed,
+                masking_names=masking_names,
                 **kwargs,
             )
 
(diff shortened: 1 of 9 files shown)


stefanradev93:
I've reviewed and approve of most changes. The new sampler is a great addition! 

One note: The `guidance_function` method implies arbitrary `kwargs` but these can actually be only the kwargs accepted by the `guidance_constraint_term` method. This is confusing and I would rather specify the arguments. In fact, the `guidance_function` method is such a thin wrapper that `guidance_constraint_term` itself can check if any `constraints` are provided (with a `None`) option and return 0 if `None`. This also eliminates the need for a `check_for_guidance` flag (similiar to the `keras.layers.Identity` or `maybe_do_stuff` design patterns).

arrjon:
Agreed, I merged the two guidance functions now.

PR #689, opened 2026-05-05
Modernize transformer attention stack with GLU FFN, Pre-LN, and composition

stefanradev93:
This PR replaces the legacy MLP-based feedforward sublayers and ad-hoc normalization patterns across our transformer modules with a cohesive set of SOTA practices used in modern large-scale transformers. I did some preliminary benchmarking and the changes are notable. Will post some numbers tomorrow when I set up some more.

---

### Gated feedforward network

Adds `bayesflow/networks/summary/transformers/attention/feedforward_net.py` implementing a GLU-style feedforward layer with four supported variants. The intermediate dimension is scaled by 2/3 * expansion_factor` so the total parameter count matches a standard non-gated FFN at the same `expansion_factor`.

---

### Major rewrite of `MultiHeadAttention` (MAB)

- **Pre-LN via RMSNorm**: normalization applied *before* each sublayer; residual path stays clean with no post-norm
- **KV normalization**: separate `ln_kv` applied to key/value inputs in cross-attention / pooling-by-attention
- **GLU FFN**: replaces the old MLP sublayer with `FFN`
- **Single `dropout` parameter**: unifies attention and FFN dropout (previously split into `dropout` + `ffn_dropout`)
- **`use_bias=False` default**: bias-free projections throughout, matching SOTA practice

### Use composition for `SetAttention` (SAB)

Refactored from `class SetAttention(MultiHeadAttention)` to `class SetAttention(keras.Layer)` with `self.mab = MultiHeadAttention(...)`. This:
- Gives `call()` a clean one-input signature, making it impossible to accidentally invoke as cross-attention
- Avoids the `super().call(x, x, ...)` anti-pattern

### `PoolingByMultiHeadAttention` (PMA)

- Replaced the pre-attention MLP transform with `FFN`
- Updated parameter signature to match the rest of the stack (`expansion_factor`, `glu_variant`, single `dropout`)

### `SetTransformer`, `TimeSeriesTransformer`, `FusionTransformer`

- Removed `mlp_depths`, `mlp_widths`, `mlp_activation` from all public signatures
- Added `expansion_factor`, `glu_variant` (both forwarded via `shared_kwargs`)
- `dropout` parameter propagated consistently to all sublayers
- Fixed bidirectional branch construction in `FusionTransformer`

### References
- https://www.youtube.com/watch?v=lVynu4bo1rY
- Shazeer (2020). *GLU Variants Improve Transformer.* [arXiv:2002.05202](https://arxiv.org/abs/2002.05202)
- Xiong et al. (2020). *On Layer Normalization in the Transformer Architecture.* ICML.

The changes are breaking but too important to pass on, I think. @LarsKue @paul-buerkner We could allow for a `use_legacy` flag in the transformers and throw a warning that a better version exist, but that would involve a hack to serialization. Alternatively, we can create a V2 close-clone module and warn users to use these (e.g., `SetTransformerV2`, ...). The problem is that this will then break serialization of `...V2` when it stops being v2.


paul-buerkner (commented):
Very cool! I can only comment on backwards-compatibility / serialization breaking aspect: I would prefer just breaking backwards compatibility / serialization here because, as you say, the changes are too important. And a legacy option is too cumbersome to maintain I think. If users want to make sure there code works 100% as expected (and that they can reload their NNs), they will have to freeze all their library versions anyway.

niels-leif-bracher (approved):
Cool, timely changes, and I'm curious about the improvements. All in all, looks good to me 💯

PR #690, opened 2026-05-05
Pin GitHub Actions to commit SHAs for supply-chain security

andife:
Replace mutable version tags with pinned commit SHAs across all GitHub Actions workflows, retaining the version as an inline comment for readability.

**Versions were updated to match what Dependabot already merged into `main`** before adding SHA pins:

| Action | Tag | SHA |
|--------|-----|-----|
| `actions/checkout` | `v6` | `de0fac2e` |
| `actions/setup-python` | `v6` | `a309ff8b` |
| `actions/upload-artifact` | `v7` | `043fb46d` |
| `actions/download-artifact` | `v8` | `3e5f45b2` |
| `codecov/codecov-action` | `v6` | `57e3a136` |
| `pypa/gh-action-pypi-publish` | `release/v1` | `cef22109` |

## Why

Pinning to SHAs prevents a compromised or force-pushed tag from silently injecting malicious code into CI. This is a supply-chain hardening measure flagged by [zizmor](https://github.com/woodruffw/zizmor), which is already part of this repo's CI.

## Maintenance

Dependabot is configured (`dependabot.yml`) to open PRs for version updates. When those land, the SHA pins here should be updated to match the new tags, keeping the inline `# vN` comments in sync.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -10,11 +10,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       with:
         persist-credentials: false
     - name: Set up Python
-      uses: actions/setup-python@v6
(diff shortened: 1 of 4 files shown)


LarsKue:
This is now covered by #691

PR #692, opened 2026-05-08
Fix `num_models` propagation issue in `ModelComparisonApproximator.fit()`

jerrymhuang:
This PR addresses a bug in `ModelComparisonApproximator.fit()`.

**Bug description**

Currently, in `ModelComparisonApproximator.fit()`, when the adapter is "auto", it calls:
```python
adapter = self.build_adapter(num_models=self.num_models, ...)
```
But `build_adapter` is only defined on the base `Approximator` class, which has no `num_models` parameter — so this raises a `TypeError`.

**Proposed fix**
Added a `build_adapter` override in `ModelComparisonApproximator` that accepts `num_models` and defaults `inference_variables` to `model_indices` (the key that `ModelComparisonSimulator` always emits) and delegates everything else to `super().build_adapter()`.

This should allow `num_models` to pass through `build_adapter()` without issue. `num_models`  is now not consumed directly in the adapter because `ModelComparisonSimulator` already outputs one-hot encoded arrays under `model_indices`, so no additional encoding step is needed.

### DIFF
--- a/bayesflow/approximators/model_comparison_approximator.py
+++ b/bayesflow/approximators/model_comparison_approximator.py
@@ -74,58 +74,6 @@ class ModelComparisonApproximator(Approximator):
         self.condition_builder = ConditionBuilder()
         self.standardizer = Standardization(standardize)
 
-    @classmethod
-    def build_adapter(
-        cls,
-        num_models: int = None,
-        inference_variables: str | Sequence[str] = "model_indices",
-        inference_conditions: str | Sequence[str] = None,
-        summary_variables: str | Sequence[str] = None,
-        sample_weight: str = None,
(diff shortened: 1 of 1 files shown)


stefanradev93:
Good catch! I don't quite get the fix? Why not simply not pass `num_models`, since it is not needed instead of doing a whole method override that does nothing but ignore an arg?

jerrymhuang:
> Good catch! I don't quite get the fix? Why not simply not pass `num_models`, since it is not needed instead of doing a whole method override that does nothing but ignore an arg?

@stefanradev93 My thought was that the user might want to call `build_adapter()` here directly, so a method override would provide this flexibility. But I think the simpler solution is better. Pivoting to that now.

PR #694, opened 2026-05-11
Add TARP diagnostic function

arrjon:
This pull request introduces the new TARP (Test of Accuracy with Random Points) diagnostic to the `bayesflow.diagnostics.metrics` module, providing a global calibration check for estimated posteriors. 

Do we to add this to our library and do we also want to add plotting capabilities for TARP?


stefanradev93:
I think this is ready to be merged. Let's consider a better name than tarp (harmonizing with the lack of abbreviations in the other diagnostics). For instance, `accuracy_random_points`? `test_accuracy_random_points`?

arrjon:
I renamed it to `accuracy_random_points`. Thought this aligns best with the current naming scheme.

PR #698, opened 2026-05-25
`ModelComparisonWorkflow`

jerrymhuang:
This PR introduces `ModelComparisonWorkflow`, a high-level entry point for amortized Bayesian model comparison, along with a family of new model-comparison scoring rules, three new diagnostic plots, and an accompanying deep-dive notebook.

## New scoring rules
Four new `ScoringRule` subclasses are introduced as generalized to K-model comparison from [Jeffrey and Wandelt (2024)](https://iopscience.iop.org/article/10.1088/2632-2153/ad1a4d/meta):
* `BrierScore`: Minimizes the squared error between softmax PMPs and the one-hot true model label.
* `PolynomialScore`: Tsallis proper scoring rule on the probability simplex; generalizes `BrierScore` via an exponent `alpha > 1`.
* `ExponentialScore`: Proper scoring rule for direct log Bayes factor estimation, using pairwise exponential penalties against the true model. This score is also generalized with a scale parameter `alpha`, and extended to incorporate a leaky parity-odd power (l-POP) nonlinear head to improve gradient flow at extreme log Bayes factor values.
* `LogisticScore`: Proper scoring rule for log Bayes factors using softplus (log-sum-exp) pairwise penalties instead of exponentials. This score is also generalized by raising the logistic term to a power `alpha`.

## New diagnostics for model comparison
* `bayes_factor_recovery` — scatter plot comparing estimated vs. true log Bayes factors.
* `pairwise_bayes_factors` — matrix of pairwise log Bayes factors across all model pairs.

In addition, a layout-related issue is addressed for the diagnostic plots. Once this PR is merged, more substantial structural changes to the `diagnostics` module will be rolled out. Stay tuned! 

## Associated changes
* `ModelComparisonApproximator` refactored to support arbitrary scoring rules (not just cross-entropy).
* `ModelComparisonSimulator` minor interface cleanup.
* **Comprehensive test coverage**: workflow, scoring rules, new diagnostic plots, and simulators.

## New notebook
New notebook demonstrating the full workflow end-to-end using two examples: 
* **Normal-normal model**: showing Bayes factor estimation workflow with analytic marginals
* **Computational finance**: showing the PMP-based workflow with the Generalized Autoregressive Conditional Heteroscedasticity (GARCH) model family for stochastic volatility modeling. 

## As a whole...
The `ModelComparisonWorkflow` adds a powerful component that can further enrich the `Workflow` ecosystem. Lots of changes are made as a result, so any feedback would be welcome.


stefanradev93:
Awesome work, Jerry! Before diving in, woudn't it be reasonable to just keep the generalized versions of the scores instead of a special case + generalized?

stefanradev93:
A few high-level comments before a thorough review:
- Perhaps it may make sense to switch to `dict` outputs and and `estimate` method for the model comp approximator / workflow in favor of a `predict` method that returns arrays. This will align it with the rest of the lib, disambiguate the outputs and allow us to also return (optionally) summaries and do all necessary transforms internally (e.g., logits to probs, log bayes factors to matrices, etc).
- We should only include the generalized scoring rules with suitable defaults (unless @paul-buerkner sees a benefit in having, e.g., `BrierScore` as a special case of `PolynomialScore`, etc).
- We should probably include a finder pattern with a dispatch for model comp approximator to allow building with quick string arguments in addition to fine-grained control with score networks.
- The new diagnostic plots are a great addition!
- The new notebook should eventually replace the simple t-test notebook from before.

jerrymhuang:
> We should only include the generalized scoring rules with suitable defaults (unless @paul-buerkner sees a benefit in having, e.g., BrierScore as a special case of PolynomialScore, etc).

@stefanradev93 I agree. My original thought was also to combine as many of the scoring rules into one generalized version as possible. I decided to separate them because I saw that currently, `MeanScore` is a special case of `NormalDifferenceScore`, and thought that maybe the same pattern should be followed if the user wants to access a special case directly. 

But I can see a scenario when, for example, `ExponentialScore`, `ScaledExponentialScore`, and `LeakyExponentialScore` grouped into one `ExponentialScore`, and `LogisticScore` and `PowerLogisticScore` grouped into one `LogisticScore`. I do think that `BrierScore` should stay in the similar spirit of the current `MeanScore`.

jerrymhuang:
Ok! @stefanradev93 the scoring rules are consolidated. I also replaced `predict()` with `estimate()` for `dict` outputs and added the finder dispatch. This can be reviewed more thoroughly now.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -56,18 +66,21 @@ def __init__(
         *,
         num_models: int,
         classifier_network: keras.Layer,
+        scoring_rule: ScoringRule = None,
```
han-ol: adjust type hint to ScoringRule | None

han-ol: applies elsewhere too

han-ol:
I wrote a little spec for the docstring summary line (the first line; PEP 257) in scoring rule classes. The goal here includes that it renders a nice overview on https://bayesflow.org/v2.0.12/api/bayesflow.scoring_rules.html

Can you follow this for the new ones and update the categorical cross entropy too? Let me know if the spec is insufficient / needs changes for the new ones.

Point estimate (one predicted value, one target):

    :math:`S(\hat \theta, \theta) = \lvert \hat \theta - \theta \rvert^2`

    Parametric distribution, generic:
    :math:`S(\hat p_\phi, \theta) = -\log \hat p_\phi(\theta)`

    Parametric distribution, named parameterization:
    :math:`S(\hat p_{\mu, \Sigma}, \theta) = -\log \mathcal{N}(\theta; \mu, \Sigma)`

    Indexed predictions (quantiles, class probabilities):
    :math:`S(\hat p_{1\ldots C}, \theta) = -\sum_{c=1}^C \theta_c \log \hat p_c`

    Sample-based:
    :math:`S(\hat \theta_{1\ldots K}, \theta) = \frac{1}{K}\sum_k \lVert \hat \theta_k - \theta \rVert - \frac{1}{2K^2}\sum_k\sum_l \lVert \hat \theta_k - \hat \theta_l
    \rVert`

    Rules

    - Use a raw string (r"""...""") so backslashes are not interpreted by Python.
    - The formula line must be the first and only content on the opening line.
    - Keep the formula self-contained: a reader should be able to identify inputs and
    outputs without reading further.
    - Subscript ranges like 1\ldots C only when the indexing is essential to the
    meaning (e.g. cross-entropy needs C classes explicit). For a scalar score like
    MeanScore, just \hat \theta and \theta suffice.
    - Scores are negatively oriented by convention.

bayesflow/networks/inference/scoring/scoring_rules/scoring_rule.py
```diff
@@ -27,6 +27,11 @@ class ScoringRule:
     """
 
     NOT_TRANSFORMING_LIKE_VECTOR_WARNING: tuple[str] = tuple()
+
+    # Kernel initializer for the output Dense head.  Subclasses that produce
+    # unbounded outputs fed directly into exp() override this to a near-zero
+    # initializer so that initial log-odds are small and exp() cannot overflow.
+    _head_kernel_initializer = "glorot_uniform"
```
han-ol, line 34: Interesting, this might be reasonable, have you tried solving the problem with a link function instead?
I can imagine you have and this was necessary additionally to keep activations sane in the beginning, then I approve.

bayesflow/utils/dispatch/find_scoring_rule.py
```diff
@@ -0,0 +1,56 @@
...
+    from bayesflow.networks.inference.scoring.scoring_rules import ScoringRule
+
+    if isinstance(arg, ScoringRule):
+        return arg
+    raise TypeError(f"Cannot infer scoring rule from {arg!r}.")
+
+
+@find_scoring_rule.register
+def _(name: str, *args, **kwargs):
+    match name.lower():
```
han-ol, line 15: I appreciate that this might change when ModelComparisonApproximator becomes a ScoringRuleApproximator anyway, but if such a finder will exist it should either be called find_comparison_scoring_rule or include the other ones.

bayesflow/networks/inference/scoring/scoring_rules/__init__.py
```diff
@@ -6,12 +6,29 @@
...
 >>> # A network to estimate both point estimates and parameters of a multivariate normal distribution.
 >>> from bayesflow.scoring_rules import MeanScore, QuantileScore, MvNormalScore
+>>> import bayesflow as bf
 >>> inference_network = bf.networks.ScoringRuleNetwork(
 ...     mean=MeanScore(),
 ...     quantiles=QuantileScore(),
 ...     mvn=MvNormalScore(),
 ... )
 
+>>> # A network trained with the Brier scoring rule.
```
han-ol: As the model comparison workflow will allow for multiple scoring rules too, the examples here should also show all categorical scores in one ScoringRuleNetwork, analogously to the continuous scores above.

bayesflow/networks/inference/scoring/scoring_rules/brier_score.py
```diff
@@ -0,0 +1,60 @@
...
+from .scoring_rule import ScoringRule
+
+
+@serializable("bayesflow.scoring_rules", disable_module_check=True)
+class BrierScore(ScoringRule):
+    r""":math:`S(\hat p_{1\ldots C}, y) = \sum_{c=1}^C (\hat p_c - y_c)^2`
+
+    Scores predicted logits against one-hot encoded target labels:
+
+    :math:`S(\hat y, y) = \sum_k (\mathrm{softmax}(\hat y)_k - y_k)^2`
```
han-ol: redundant with first docstring line

han-ol: issue applies to the other categorical scores too.

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
```diff
@@ -0,0 +1,134 @@
...
+    f_m = keras.ops.take_along_axis(f_full, m_idx, axis=-1)  # (..., 1)
+    return f_full - f_m  # (..., M), broadcast
+
+
+class _LeakyLink(keras.Layer):
+    """Applies the leaky parity-odd power transform J_alpha(x) = x(1 + |x|^{alpha-1}) element-wise."""
+
+    def __init__(self, alpha: float, eps: float = 1e-8, **kwargs):
+        super().__init__(**kwargs)
+        self.alpha = alpha
```
han-ol: Is there a good reason to call the exponent attribute alpha here? Elsewhere it is called lambda (exponential score docstring) and alpha is already a different attribute of that score as well.

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
```diff
@@ -0,0 +1,134 @@
...
+        S(\{f_k\}, m; \alpha) = \sum_{k \neq m}
+            \exp\!\left(\frac{\alpha}{2}(f_k(x) - f_m(x))\right)
+
+    The unique minimiser satisfies
+    :math:`f_k^*(x) = \frac{1}{\alpha} \log K_{k,0}`,
+    so network outputs are multiplied by :math:`\alpha` to recover log Bayes
+    factors. Setting :math:`\alpha = 1` (default) gives the plain rule where
+    the network directly estimates log Bayes factors.
+
+    Optionally, a leaky parity-odd power (l-POP) head link
```
han-ol: Links can be passed directly to `ExponentialScore(self, ..., links=bf.links.Leaky(lambda=2))`, so we do not need to couple links to specific score classes.

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
```diff
@@ -0,0 +1,134 @@
...
+    targets = keras.ops.convert_to_tensor(targets)
+    zeros = keras.ops.zeros_like(f[..., :1])
+    f_full = keras.ops.concatenate([zeros, f], axis=-1)  # (..., M)
+    m = keras.ops.cast(keras.ops.argmax(targets, axis=-1), dtype="int32")
+    m_idx = keras.ops.expand_dims(m, axis=-1)  # (..., 1)
+    f_m = keras.ops.take_along_axis(f_full, m_idx, axis=-1)  # (..., 1)
+    return f_full - f_m  # (..., M), broadcast
+
+
+class _LeakyLink(keras.Layer):
```
han-ol: Is this link uniquely useful (or necessary) in the `ExponentialScore` or do you have otherwise a reason to keep it private here? We could add it to the bayesflow.links submodule.

bayesflow/networks/inference/scoring/scoring_rules/logistic_score.py
```diff
@@ -0,0 +1,98 @@
...
+        A positive float uses power mode (:math:`\exp(\alpha \cdot \mathrm{softplus})` loss).
+    """
+
+    NOT_TRANSFORMING_LIKE_VECTOR_WARNING = ("log_bayes_factors",)
+    # Small-stddev init keeps initial log-odds near zero, preventing exp() overflow at the start of training.
+    _head_kernel_initializer = keras.initializers.TruncatedNormal(mean=0.0, stddev=0.01)
+
+    def __init__(self, alpha: float | None = None, **kwargs):
+        super().__init__(**kwargs)
+        self.alpha = alpha
```
han-ol, line 53: alpha must be positive, correct? I'd explicitly check and raise a ValueError if not None or positive.

bayesflow/networks/inference/scoring/scoring_rules/logistic_score.py
```diff
@@ -0,0 +1,98 @@
...
+
+    **Log mode** (``alpha=None``, default):
+
+    .. math::
+
+        S(\{f_k\}, m) = \sum_{k \neq m} \log\!\left(1 + e^{f_k(x) - f_m(x)}\right)
+
+    The unique minimiser is :math:`f_k^*(x) = \log K_{k,0}(x)`.
+
+    **Power mode** (``alpha`` is a positive float):
```
han-ol: I am not sure why we'd want to have both of these modes in one class. Is there a reason for combining?

bayesflow/networks/inference/scoring/scoring_rules/polynomial_score.py
```diff
@@ -0,0 +1,73 @@
...
+        S(p, m; \alpha)
+        = \frac{\alpha - 1}{\alpha}\sum_k p_k^\alpha - p_m^{\alpha - 1}
+
+    where :math:`p = \mathrm{softmax}(\hat y)` and :math:`m` is the true model index.
+    The unique minimiser of the expected score is the true posterior :math:`p_k^* = P(\mathcal{M}_k \mid x)`
+    for any :math:`\alpha > 1`.
+
+    For :math:`\alpha = 2` this is proportional to the :class:`BrierScore`
+    (same gradient direction, same minimiser).  Larger :math:`\alpha` sharpens the
+    penalty for confidently wrong predictions.
```
han-ol: I'd drop the "confidently" here, as I think there is nothing particularly doubtful or confident about a prediction, be it correct or wrong.

bayesflow/networks/inference/scoring/scoring_rules/polynomial_score.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Parameters
+    ----------
+    alpha : float, optional
+        Exponent (default: 2.0).  Must satisfy :math:`\alpha > 1`.
+    """
+
+    def __init__(self, alpha: float = 2.0, **kwargs):
+        super().__init__(**kwargs)
+        self.alpha = alpha
```
han-ol, line 40: I think the score method silently accepts alpha <=1, so we should check and raise a ValueError here.

bayesflow/networks/inference/scoring/scoring_rules/scoring_rule.py
```diff
@@ -159,11 +164,30 @@ def get_head(self, key: str, output_shape: Shape) -> keras.Sequential:
...
 
         return keras.Sequential([subnet, dense, reshape, link])
 
+    @property
+    def is_pmp_rule(self) -> bool:
+        """True for PMP rules (output logits), False for Bayes factor rules (output log Bayes factors)."""
+        return "logits" in self.get_head_shapes_from_target_shape((1, 2))
+
+    def to_bayes_factors(self, f: Tensor) -> Tensor:
+        """Convert raw network outputs to log Bayes factors.
```
han-ol: The docstring reads as if there are only model comparison scores.

We should either 
(1) explicitly mention this method applies to those only
(2) introduce a simple `DiscreteScoringRule` (or similar name), that inherits from `ScoringRule`, adds specific methods and is the base method of all the model comp scores.

(2) might also be good to allow easy validation of scores passed to the model comparison workflow: just check `isinstance(score, DiscreteScoringRule)`

bayesflow/simulators/model_comparison_simulator.py
han-ol: It is probably not intentional that this ignores callable classes and partial expressions.

han-ol: `callable(shared_simulator) and not isinstance(shared_simulator, Simulator)` would be more robust.

bayesflow/simulators/model_comparison_simulator.py
```diff
@@ -83,18 +82,17 @@ def __init__(
...
-                logits = np.log(p) - np.log(1 - p)
-            case _:
-                raise ValueError("Received conflicting arguments. At most one of `p` or `logits` must be provided.")
+        if logits is not None and p is not None:
+            raise ValueError("Received conflicting arguments. At most one of `p` or `logits` must be provided.")
+        elif p is not None:
+            p = np.array(p, dtype=float)
+            if not np.isclose(np.sum(p), 1.0):
+                raise ValueError("Probabilities must sum to 1.")
+            if np.any(p <= 0):
```
han-ol, line 91: I'd check for positivity first, then normalization.

Elsewhere in the library, we are not enforcing input normalization and instead always normalize to p_i/sum(p_i). Check out `ScoringRuleApproximator.sample(..., score_weights=...)` for example. Just for consistency we could also do that here - or would you prefer explicit checks here specifically / generally?

han-ol (changes requested):
Very happy to see extension of the scoring rule module. Goes in a good direction. I have a few comments about argument validation, distinction of scores and links and some minor things and some questions.

I focused this first review on the scoring rules, as the workflow will still change to allow passing and co-learning multiple scoring rules.

bayesflow/simulators/model_comparison_simulator.py
han-ol:
```python
>>> from types import FunctionType
>>> from functools import partial
>>> foo = lambda x,y: x*y
>>> isinstance(partial(foo, y=14), FunctionType)
False
```

bayesflow/approximators/model_comparison_approximator.py
stefanradev93: Please, let's not use `stuff | None` anywhere in the lib, as it adds loads of redundant clutter if the default is `None`.

bayesflow/networks/inference/scoring/scoring_rules/scoring_rule.py
jerrymhuang: Yeah, that's the purpose. Link function is also less favorable here because it can be overly constraining on the range of outputs.

bayesflow/networks/inference/scoring/scoring_rules/logistic_score.py
jerrymhuang: Done.

bayesflow/networks/inference/scoring/scoring_rules/polynomial_score.py
jerrymhuang: Done.

bayesflow/networks/inference/scoring/scoring_rules/polynomial_score.py
jerrymhuang: Done.

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
jerrymhuang: Done.

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
jerrymhuang: Done.

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
jerrymhuang: Calling it lambda would create confusion with Python's lambda function. To disambiguate, I changed this to `scale` for ExponentialScore's `alpha`, and `power` for l-POP's `alpha`.

bayesflow/networks/inference/scoring/scoring_rules/scoring_rule.py
jerrymhuang: Yeah, great idea! I decided to go with 2) and introduced `CategoricalScoringRule`.

bayesflow/networks/inference/scoring/scoring_rules/logistic_score.py
jerrymhuang: It was originally two different classes: `LogisticScore` and `PowerLogisticScore`. But I combined them because they share the same core function `1 + \exp(f_k - f_m)`, just with different outer functions associated with each mode.

I am open to splitting them up again, though.

han-ol: I'd split it, if you are on the fence.

han-ol: I'd split it if you are on the fence.

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
han-ol: Looks good, thanks!

jerrymhuang:
All points of feedback for the scoring rules have now been addressed. I also improved code coverage along the way.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -22,12 +23,18 @@
 @serializable("bayesflow.approximators")
 class ModelComparisonApproximator(Approximator):
```
stefanradev93: Why did you decide to inherit from `Approximator` and not `ScoringRuleApproximator` after all?

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -251,40 +268,70 @@ def get_config(self):
...
+        else:
+            raise RuntimeError(
+                f"Unrecognized scoring rule output keys: {list(raw.keys())}. "
+                "Expected 'logits' (PMP rules) or 'log_bayes_factors' (Bayes factor rules)."
+            )
 
-        if probs:
-            logits = keras.ops.softmax(logits)
+        if summary_outputs is not None:
+            result["summaries"] = keras.ops.convert_to_numpy(summary_outputs)
```
stefanradev93: I think we settled for `_summaries` to avoid rare cases where `summaries` is an actual variable.

bayesflow/diagnostics/metrics/model_comparison_accuracy.py
```diff
@@ -0,0 +1,65 @@
...
+    num_models = true_models.shape[-1]
+    true_labels = true_models.argmax(axis=-1)
+
+    if predictions.shape[-1] == num_models - 1:
+        # Bayes factor mode: prepend f_0 = 0 and take argmax
+        f0 = np.zeros((predictions.shape[0], 1), dtype=predictions.dtype)
+        predictions = np.concatenate([f0, predictions], axis=-1)
+
+    predicted_labels = predictions.argmax(axis=-1)
+    return float(np.mean(predicted_labels == true_labels))
```
stefanradev93: Avoid redundant AI overcaution pls. `np.mean` returns a float.

bayesflow/approximators/model_comparison_approximator.py
han-ol: I think that part is still in the works.

bayesflow/approximators/model_comparison_approximator.py
jerrymhuang: Done.

bayesflow/approximators/model_comparison_approximator.py
jerrymhuang: It's now inherited from `ScoringRuleApproximator` and allows passing and co-learning multiple scoring rules at once.

han-ol:
I think we can further harmonize the `ModelComparisonApproximator` with the other approximators:

They always take an inference_network in the constructor, but here such an inference network is not passed but constructed internally.

I'd propose moving from 
```python
appr = ModelComparisonApproximator(
    num_models=3,
    classifier_network=MLP(),
    scoring_rules=CrossEntropyScore(),
    summary_network=SetTransformer(),
)
```

to 

```python
appr = ModelComparisonApproximator(
    num_models=3,
    inference_network=ScoringRuleNetwork(CrossEntropyScore(), subnet=MLP()),
    summary_network=SetTransformer(),
)
```

For reference the situation with a continuous approximator.

```python
appr = ContinuousApproximator(
    inference_network=ScoringRuleNetwork(MeanScore(), subnet=MLP()),
    summary_network=SetTransformer(),
)
```


This unifies the interface and simplifies the code. For example, no need anymore to bypass ContinuousApproximator.get_config.

han-ol:
I think we can further harmonize the `ModelComparisonApproximator` API with the other approximators, while slimming it down.
`ModelComparisonApproximator` should be closer to `ScoringRuleApproximator`.

All approximators other than current `ModelComparisonApproximator` expect an `inference_network` constructor argument, but here such an inference network is not passed but constructed internally.

I'd propose moving from 
```python
appr = ModelComparisonApproximator(
    num_models=3,
    classifier_network=MLP(),
    scoring_rules=CrossEntropyScore(),
    summary_network=SetTransformer(),
)
```

to 

```python
appr = ModelComparisonApproximator(
    num_models=3,
    inference_network=ScoringRuleNetwork(CrossEntropyScore(), subnet=MLP()),
    summary_network=SetTransformer(),
)
```

For reference the situation with a continuous scoring rule approximator.

```python
appr = ScoringRuleApproximator(
    inference_network=ScoringRuleNetwork(MeanScore(), subnet=MLP()),
    summary_network=SetTransformer(),
)
```


This unifies the interface and simplifies the code: `ModelComparisonApproximator.__init__` and `get_config`, ...(?) become much simpler / obsolete.

By the way: Is the `num_models` argument necessary and for what?

jerrymhuang:
This interface is indeed cleaner! I slimmed down the `ModelComparisonApproximator`, which also eliminated `num_models` since it is deemed no longer necessary.

LarsKue:
Unrequesting my review on this, please re-request when you deem the PR ready

bayesflow/links/leaky.py
```diff
@@ -0,0 +1,29 @@
...
+    r"""Leaky parity-odd power (l-POP) transform :math:`J_\lambda(x) = x(1 + |x|^{\lambda - 1})`.
+
+    Expands large magnitudes super-linearly while remaining odd and smooth,
+    improving numerical recovery of extreme log Bayes factors without affecting
+    properness of the scoring rule.
+
+    Parameters
+    ----------
+    power : float
+        Exponent :math:`\lambda`. Must be greater than 1. Recommended value: ``2.0``.
```
stefanradev93: Why not have a default of 2 then?

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
```diff
@@ -0,0 +1,116 @@
...
+from bayesflow.types import Shape, Tensor
+from bayesflow.utils import weighted_mean
+from bayesflow.utils.serialization import serializable
+
+from .categorical_scoring_rule import CategoricalScoringRule
+
+
+def _pairwise_diff(f: Tensor, targets: Tensor) -> Tensor:
+    """Prepend f_0=0 and compute f_k - f_m for all k, where m is the true model."""
+    targets = keras.ops.convert_to_tensor(targets)
```
stefanradev93: needed?

bayesflow/networks/inference/scoring/scoring_rules/exponential_score.py
```diff
@@ -0,0 +1,116 @@
...
+    zeros = keras.ops.zeros_like(f[..., :1])
+    f_full = keras.ops.concatenate([zeros, f], axis=-1)  # (..., M)
```
stefanradev93: Pls get rid of AI shape hints

bayesflow/networks/inference/scoring/scoring_rules/power_logistic_score.py
```diff
@@ -0,0 +1,75 @@
...
+            One-hot encoded true model labels of shape ``(..., M)``.
+        weights : Tensor, optional
+            Per-sample weights for a weighted mean.
+
+        Returns
+        -------
+        Tensor
+            (Optionally weighted) mean power logistic score over the batch.
+        """
+        targets = keras.ops.convert_to_tensor(targets)
```
stefanradev93: Needed?

bayesflow/networks/inference/scoring/scoring_rules/power_logistic_score.py
```diff
@@ -0,0 +1,75 @@
...
+        diff = _pairwise_diff(estimates["log_bayes_factors"], targets)
+        mask = 1.0 - targets
+        M = keras.ops.cast(keras.ops.shape(diff)[-1], dtype="float32")
+        clip_max = 88.0 - keras.ops.log(keras.ops.maximum(M - 1.0, 1.0))
```
stefanradev93: Magic number?

bayesflow/diagnostics/metrics/model_comparison_accuracy.py
stefanradev93: Bump

bayesflow/networks/inference/scoring/scoring_rules/brier_score.py
```diff
@@ -0,0 +1,56 @@
...
+            One-hot encoded target labels of shape ``(..., num_models)``.
+        weights : Tensor, optional
+            Per-sample weights for a weighted mean.
+
+        Returns
+        -------
+        Tensor
+            (Optionally weighted) mean Brier score over the batch.
+        """
+        targets = keras.ops.convert_to_tensor(targets)
```
stefanradev93: Is this needed?

stefanradev93 (changes requested):
Almost there, see remaining comments.

stefanradev93:
At this point, I am becoming positively accustomed to the codebase and interface but would very much like for @paul-buerkner to go over the new tutorial and approve / request final changes. For example, I am not sure whether:
1. We should have the option for analytic Bayes factors (practically never available) in `plot_default_diagnostics`; I would still keep the function as a standalone diagnostic comparison against Bayes factors from another method (be it analytic or approximate);
2. We should be showing a matrix of **mean** log Bayes factors. These can be skewed easily by extreme Bayes factors, which is why I suspect the matrix does not correspond to the confusion matrix shown later. @paul-buerkner what do you think? Perhaps we should always convert to a confusion matrix to keep outputs comparable regardless of the scoring rule.
3. We should not also demonstrate numeric diagnostics from the workflow.
4. We should not train the ARCH models using offline training to demonstrate good performance in that mode in addition to online training.

bayesflow/networks/inference/scoring/scoring_rule_network.py
```diff
@@ -183,6 +183,7 @@ def call(
     def compute_metrics(
         self, x: Tensor, conditions: Tensor = None, sample_weight: Tensor = None, stage: str = "training", **kwargs
     ) -> dict[str, Tensor]:
+        x = keras.ops.convert_to_tensor(x)
```
han-ol: Is there ever a situation where x is not Tensor?

stefanradev93: Nope, this seems to be an artefact.

paul-buerkner (changes requested):
I have reviewed specifically the example notebook. I made some direct edits myself to improve clarify and flow of the notebook. 

My further comments relate both to notebook and to aspects of the code (user interface and output).

Notebook comments:
- This is indeed a deep dive that starts right with scoring rules, a super technical aspect of model comparison. As a beginner of Bayesflow just wanting to learn about model comparison, this would immedietaly scare me. If we have a deep dive, can we also have a "shallow dive"? So, can we perhaps keep the t-test example notebook and just update it to the new workflow? No to need to talk about scoring rules etc there. Just estimate a bunch of PMPs an be happy. 
- in the notebook, `plot_default_diagnostics` produces 4 plots, but explained are only 3 in the 4 normals case study. Can you add explanation for the missing one (the loss history)?
- Speaking of loss history, why is the loss flat? The model seems to learn something but the loss does not improve. how can this be?
- please inteprete the output of plot_default_diagnostics with respect to the models in questions. Is what we see good/not so good, does it make sense etc.? 
- Same with the PMP diagnostic plots.
- The BF scatter plots colors to quadrants. Explain why this is done when introducing the plot.
- The argument for using PMP rather than BF scoring rules in the GARCH model is lacking. With bayesflow we can easily compute both, independent of whether the likelihood or marginal likelihood is analytical. I have simplified the argumentation to reflect this. please double check.
- Explain what the `model_names` argument does. I assume it is just for pretty printing?
- Also in the GARCH model, I am mssing substantive interpretation of the diagnostic results. Please add. 

Code:
- in the Pairwise log Bayes factors plot, what is log K? Do you mean log BF, or something else? K is maximally uninformative.
- The workflow argument is named `scoring_rules` (emphasize on **s**). does this mean I could pass multiple scoring rules at once (as a list of scoring rules)? If yes, how will they play together in the output?
- Can we change the confusion matrix to alternatively output percentages rather than raw counts? At least via a (non-default) argument. Perhaps it is already possible?

jerrymhuang:
@stefanradev93 @han-ol @paul-buerkner I have incorporated all the changes you suggested. Please feel free to merge this PR if everything looks good.

bayesflow/networks/inference/scoring/scoring_rules/categorical_scoring_rule.py
```diff
@@ -0,0 +1,64 @@
...
+
+        The default implementation is the identity, correct for BF rules whose
+        minimiser is directly the log Bayes factor (e.g.
+        :class:`ExponentialScore`, :class:`LogisticScore`, :class:`PowerLogisticScore`). Subclasses whose
+        minimiser lives in a transformed space must override this method.
+        """
+        return f
+
+
+def check_categorical_rules_not_mixed(scoring_rules) -> None:
```
stefanradev93: Make this a static method of `CategoricalScoringRule`.

bayesflow/networks/inference/scoring/scoring_rules/logistic_score.py
```diff
@@ -0,0 +1,56 @@
...
+            One-hot encoded true model labels of shape ``(..., M)``.
+        weights : Tensor, optional
+            Per-sample weights for a weighted mean.
+
+        Returns
+        -------
+        Tensor
+            (Optionally weighted) mean logistic score over the batch.
+        """
+        targets = keras.ops.convert_to_tensor(targets)
```
stefanradev93: Why is a conversion to tensor needed here?

bayesflow/networks/inference/scoring/scoring_rules/power_logistic_score.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Largest integer x such that exp(x) does not overflow in float32.
+FLOAT32_EXP_MAX = np.floor(np.log(np.finfo(np.float32).max))
+
+
+@serializable("bayesflow.scoring_rules", disable_module_check=True)
+class PowerLogisticScore(CategoricalScoringRule):
+    r""":math:`S(\{f_k\}, m; \alpha) = \sum_{k \neq m} \left(1 + e^{f_k - f_m}\right)^\alpha`
+
+    The unique minimiser is :math:`f_k^*(x) = \tfrac{1}{\alpha+1}\log K_{k,0}(x)`,
```
stefanradev93: "minimizer"

bayesflow/networks/inference/scoring/scoring_rules/power_logistic_score.py
```diff
@@ -0,0 +1,78 @@
+import keras
```
stefanradev93: Unsure if this class is needed. The `LogisticScore` should support powers. It is not a household name compared to Brier or Cent.

bayesflow/networks/inference/scoring/scoring_rules/categorical_scoring_rule.py
```diff
@@ -0,0 +1,64 @@
...
+        """True for PMP rules (output logits), False for Bayes factor rules (output log Bayes factors)."""
+        return "logits" in self.get_head_shapes_from_target_shape((1, 2))
+
+    def to_bayes_factors(self, f: Tensor) -> Tensor:
+        """Convert raw network outputs to log Bayes factors.
+
+        The default implementation is the identity, correct for BF rules whose
+        minimiser is directly the log Bayes factor (e.g.
+        :class:`ExponentialScore`, :class:`LogisticScore`, :class:`PowerLogisticScore`). Subclasses whose
+        minimiser lives in a transformed space must override this method.
```
stefanradev93: "minimizer"

stefanradev93 (changes requested):

stefanradev93:
I think we are almost there. I have a few points remaining besides the concrete review remarks in the code:
- Since one can pass multiple scores, why can't PMP and BF scores be mixed. Each head has its own output dimensionality, so I don't quite understand the rule in place for not mixing scores. Can we perhaps show this in the tutorial (see below)?
- I suggest (@paul-buerkner pls agree or disagree) to not maintain two model comparison tutorial notebooks, but instead organize the current one around three section: 1) directly showing model comparison with default PMPs on the Gaussians, minimal scoring rule vernacular, from zero-to-hero in a few lines; 2) unpacking the various scoring rules [current one] and how multiple ones can be used, including the math behind them (can be another entry in the table); 3) the GARCH demo which is non-trivial, the network actually needs to learn across epochs, and there are no ground truths.
- This occurred to me while writing the comment: do we have an equivalent of `compute_default_diagnostics` that returns numerical summaries (e.g., accuracy, ECE, and Brier score per model)?
- Minor: I am personally not a fan of the green colormap for the plots and would like something closer to our default (blueish) color map.

bayesflow/networks/inference/scoring/scoring_rules/categorical_scoring_rule.py
jerrymhuang: This is no longer needed as PMP and BF rule mixing is now enabled.

jerrymhuang:
@stefanradev93 I have addressed all your feedback except for the notebook. To this end, I side with @paul-buerkner's comments earlier that there should be two separate notebooks, as scoring rules are more in-depth topic. Another argument for this separation is that a very long notebook would be as daunting to newcomers of BayesFlow (and model comparison as a concept) as introducing scoring rules off the bat.

We can, however, refer to the shorter notebook in the deep dive notebook and say that the former is a prerequisite. Alternatively, if we have to merge everything into one notebook, then I would suggest clearly demarcate the sections so that people know that the first section is introductory and suffices for most MC tasks.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -0,0 +1,623 @@
...
+        return adapter
+
+    def estimate(
+        self,
+        *,
+        conditions: dict,
+        return_summaries: bool = False,
+        merge_scores: bool = True,
+        **kwargs,
+    ) -> dict[str, np.ndarray]:
```
han-ol: fix output type (it currently assumes merge_score=True)

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -0,0 +1,623 @@
...
+            **filter_kwargs(kwargs, ModelComparisonApproximator.__init__),
+        )
+
+        self._init_optimizer(initial_learning_rate, optimizer, **kwargs.get("optimizer_kwargs", {}))
+        self._init_checkpointing(checkpoint_filepath, checkpoint_name, save_weights_only, save_best_only)
+        self.history = None
+        self._needs_compile = True
+
+    @property
+    def classifier_network(self) -> keras.Layer:
```
han-ol: (Why) is this necessary? Didn't see it used anywhere.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -0,0 +1,623 @@
...
+        - ``optimizer_kwargs``  : dict — passed to ``_init_optimizer``.
+        - ``simulator_kwargs``  : dict — passed to :class:`~bayesflow.simulators.ModelComparisonSimulator`.
+        - Other keys forwarded to :class:`~bayesflow.approximators.ModelComparisonApproximator`.
+    """
+
+    def __init__(
+        self,
+        simulator: Sequence[Simulator] | ModelComparisonSimulator | None = None,
+        adapter: Adapter | None = None,
+        classifier_network: keras.Layer | str = "mlp",
```
han-ol: It is not obvious, albeit also not really wrong, to call this `classifier_network`, elsewhere in the library we would call it `subnet`. (e.g. coupling_flow, scoring_rule_network, ...).

Historically we had a classifier network argument, but that included what are now the heads (corresponding to scoring rules). Now it is just the shared base of a scoring rule network (gets forwarded).

I would propose to rename `classifier_network` -> `subnet` and drop the workflow level `find_network` resolution. THat is done already inside the scoring rule network anyway. Behaves the same because `find_network` is idempotent, but just unnecessary to call it twice.

That would mean `ModelComparisonWorkflow(scoring_rules=, ..., subnet="mlp", subnet_kwargs=dict(dropout=0.5))`, which in the direct object construction pattern would be `ModelComparisonWorkflow(scoring_rules=, ..., scoring_subnet=MLP(dropout=0.5,...)`.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -0,0 +1,623 @@
...
+            raise KeyError(
+                "test_data must contain 'model_indices' (raw simulator output) or "
+                "'inference_variables' (adapted output). Neither key was found."
+            )
+
+        rules = self.approximator.inference_network.scoring_rules.values()
+        has_pmp_rules = any(rule.is_pmp_rule for rule in rules)
+        has_bf_rules = any(not rule.is_pmp_rule for rule in rules)
+
+        # Diagnostics operate on merged 'model_probs'; force it regardless of any user estimate_kwargs.
```
han-ol: If I am interested in diagnosing a particular scoring rule, or realistically all of them separately, what is the intended way to get them?

Instead of silently dropping a `merge_score=False` kwarg, we could also create diagnostics in a for loop over the estimates dict.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -0,0 +1,623 @@
...
+
+        Parameters
+        ----------
+        conditions : dict[str, np.ndarray]
+            Conditioning data as produced by the simulator (or real observations).
+        return_summaries: bool, optional
+            If set to True and a summary network is present, will return the learned summary statistics for
+            the provided conditions.
+        merge_scores : bool, optional
+            Only relevant when the workflow was configured with more than one scoring rule
```
han-ol: Rough note, hopefully I'll come back to this with a more concrete proposal: this could be less confusing.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -0,0 +1,623 @@
...
+        self,
+        *,
+        conditions: dict,
+        return_summaries: bool = False,
+        merge_scores: bool = True,
+        **kwargs,
+    ) -> dict[str, np.ndarray]:
+        """
+        Return posterior model probabilities (and raw network outputs) for the given conditions.
+
```
han-ol: For the sphinx autodoc generated API reference it would be good UX to link to the more in depth information in the ModelComparisonApproximator.estimate method.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -230,61 +220,93 @@ def fit(
...
         -------
-        outputs: np.ndarray
-            Predicted posterior model probabilities given `conditions`.
+        dict[str, np.ndarray]
+            Single scoring rule: always contains ``"model_probs"`` of shape
+            ``(num_datasets, num_models)``. PMP rules additionally contain ``"logits"``;
+            Bayes factor rules additionally contain ``"log_bayes_factors"``.
+            If a summary network is present, also contains ``"_summaries"``.
+
+            Multiple scoring rules with ``merge_scores=True`` (default): the flat single-rule
```
han-ol: I can feel the LLM talking to me in this docstring, which is not what we should want in our docs.

More concretetely, we should make a clear decision about how and why the docstrings of approximator and workflow methods differ, as it is the same method plus a progress bar.

bayesflow/approximators/model_comparison_approximator.py
```diff
@@ -230,61 +220,93 @@ def fit(
...
-        resolved_conditions, adapted, _ = self._prepare_conditions(conditions, **kwargs)
+        raw = self.inference_network(xz=None, conditions=resolved_conditions, **inference_kwargs)
 
-        inference_kwargs = self._collect_mask_kwargs(self._INFERENCE_MASK_KEYS, adapted)
+        per_rule = {
+            rule_key: self._rule_output_to_model_probs(self.inference_network.scoring_rules[rule_key], rule_output)
+            for rule_key, rule_output in raw.items()
+        }
+
+        # Single rule → flat dict (merge_scores is a no-op here).
```
han-ol: "→"

please avoid characters that nobody can type with a keyboard. I'd like us to stick to ASCII only.

han-ol: more concretely to that comment, I'd just delete the line.

bayesflow/diagnostics/plots/mc_calibration.py
```diff
@@ -23,21 +23,23 @@ def mc_calibration(
     metric_fontsize: int = 14,
     tick_fontsize: int = 12,
     epsilon: float = 0.02,
-    figsize: Sequence[int] = None,
+    fig_size: Sequence[int] = None,
```
han-ol: why not `figsize` like in matplotlib?

bayesflow/workflows/model_comparison_workflow.py
han-ol: done

bayesflow/workflows/model_comparison_workflow.py
han-ol: I went ahead and pushed the proposed change in 2b90deee56ef444b21b6febd379a3c83ee557de7, not to force the proposal, but to make it easier to see what I mean. Feel free to amend

han-ol (changes requested):
I reviewed the workflow so far and made some of the requested changes.
The integration of scoring rule networks looks good to me, but I am thinking about whether the distinction of PMP scores and BF scores could be more elegant by having the BF scores also actually output a fixed 0 as the reference model's logit, so that all of the scores predict logits. Haven't reached a conclusion yet.

bayesflow/workflows/model_comparison_workflow.py
stefanradev93: Your marked four lines. Which do you mean? `needs_compile` is an essential flag for all workflows, if that's what you mean.

bayesflow/workflows/model_comparison_workflow.py
stefanradev93: Great!

bayesflow/workflows/model_comparison_workflow.py
stefanradev93: Agreed. Please, make sure the tests are passing.

bayesflow/approximators/model_comparison_approximator.py
stefanradev93: Yes please. Let's try to avoid bamboozling, oversophisticating LLM slop lingo throughout.

bayesflow/workflows/model_comparison_workflow.py
stefanradev93: I don't think the method is even necessary (same signature as the one in `BasicWorkflow` and same defaults as in `ModelComparisonApproximator`. Removed in 72caa6ad8.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -0,0 +1,598 @@
...
+        # When a shared_simulator provides context variables used as inference_conditions,
+        # those variables are scalars (one value per batch, not per sample) and must be
+        # broadcast to the batch dimension before the adapter can concatenate them.
+        # We auto-detect the broadcast target as the first summary variable.
+        broadcast_ref = None
+        if shared_simulator is not None and inference_conditions is not None:
+            sv = [summary_variables] if isinstance(summary_variables, str) else (summary_variables or [])
+            if sv:
+                broadcast_ref = sv[0]
```
stefanradev93: This is also based on a wrong assumption. The context variables need not be scalars.

stefanradev93 (changes requested):

bayesflow/workflows/model_comparison_workflow.py
han-ol: I meant the `classifier_network` being a property of the workflow.

han-ol: you already removed it :+1:

han-ol:
Summary of offline discussion (thanks @stefanradev93, @jerrymhuang, @Kucharssim):

1. Naming: "log_bayes_factor" is renamed to "log_odds", since those learned quantities are actually log posterior odds. They only equal log BFs under uniform model priors, which we do support deviating from.
2. Unified output: All `CategoricalScoringRule` subclasses have estimation head output of length M (collapsing the M vs. M-1 split). Since scores disagree on their natural input space (polynomial needs probs, logistic and exponential score need raw log-odds, cross-entropy takes either), a `Softmax` link absorbs the transformation of the logits from the head pre-activation to probabilities, or the default identity link passes on the raw unnormalized logits.
3. Derived quantities: PMPs are computed from logits; log BFs from logits + model priors ("enrichment" of the estimate dict by approximator). Prior model weights come from the simulator when available; otherwise equal priors are assumed.

I can take over the logit+link unification of all scores.

han-ol:
The unification of categorical scoring rules is in place. I have relied on the testing suite and have not trained anything myself. The diagnostics are untouched, but still are shaped in a way path-dependent on the pmp rule split. In particular they request BFs to be M-1 in length. Probably other things too, have not looked too closely.

The docstrings are verbose in some places. I only quickly edited them to reflect the changes. I would like the output description of the estimate methods to be more consistent (including terser) to ScoringRuleApproximator.estimate.

Not sure what the failing doc tests are about.

Just so we do not forget, I believe some of the earlier review comments are not addressed yet as well.

stefanradev93:
Took a while to cleanup the AI slop and fix some rough / edge cases. I still want to add some changes on the deep dive notebook, but feel free to quickly double-check and approve @paul-buerkner @han-ol (once the tests are passing).

Note, that I also had to change the behavior of standardize to always lazy initialize and deal with missing keys.

bayesflow/diagnostics/metrics/accuracy.py
```diff
@@ -0,0 +1,102 @@
...
+def accuracy(
+    predictions: np.ndarray,
+    true_models: np.ndarray,
+    model_names: Sequence[str] = None,
+    per_model: bool = True,
+) -> dict[str, any] | float:
+    """
+    Compute classification accuracy for model comparison networks.
+
+    Works with both PMP and Bayes factor scoring rules:
```
han-ol: The distinction is gone now. We can simplify these.

bayesflow/workflows/model_comparison_workflow.py
han-ol: answer: users pass estimates from some score directly to the method and get the corresponding figures. that is nice.

bayesflow/networks/inference/scoring/scoring_rules/brier_score.py
```diff
@@ -1,41 +1,49 @@
...
+        For one-hot targets the Brier score is an exact affine map of the
+        :math:`\alpha = 2` Tsallis polynomial score,
+        :math:`S_{\mathrm{Brier}} = 2\,S_{\mathrm{poly}} + 1`. The heavy lifting is
+        therefore inherited from :class:`PolynomialScore`; :meth:`score` only applies
+        this transform so that the reported value is the true Brier score. (The scale
+        and offset are immaterial for Bayes-risk minimization — same minimizer and
+        gradient direction — but keep the metric interpretable.)
     """
 
+    #: Affine map from the alpha=2 Tsallis polynomial score to the Brier score.
```
han-ol: We use attribute docstrings elsewhere for such a comment (scoring_rule.py). Could we match that here? They also have wider tool support, #: is Sphinx-specific I think.

bayesflow/networks/inference/scoring/scoring_rules/logistic_score.py
```diff
@@ -23,9 +23,9 @@ class LogisticScore(CategoricalScoringRule):
 
     for :math:`\alpha = 0` (default, the plain logistic rule).
 
-    The unique minimizer is :math:`f_k^*(x) = \tfrac{1}{\alpha+1}\log K_{k,0}(x)`,
-    so network outputs are multiplied by :math:`\alpha + 1` to recover log Bayes
-    factors. For :math:`\alpha = 0` the network directly estimates log Bayes factors.
+    The network outputs logits :math:`f` whose softmax is the posterior over
+    models; pairwise differences :math:`f_k - f_j` estimate log Bayes factors
+    :math:`\log K_{k,j}`.
```
han-ol: the pairwise difference point is a slip from previous times when we didn't distinguish between bayes factors and posterior odds.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -43,16 +39,17 @@ class ModelComparisonWorkflow(BasicWorkflow):
...
         :func:`~bayesflow.utils.find_network` (e.g. ``"mlp"``).
     scoring_rules : CategoricalScoringRule or Sequence or dict[str, CategoricalScoringRule] or str, optional
         Scoring rule(s) used to train the classifier. Accepts, in increasing generality:
 
         - a single :class:`~bayesflow.scoring_rules.CategoricalScoringRule` instance;
         - a string name recognised by :func:`~bayesflow.utils.find_scoring_rule`;
-        - a **list/tuple** of rules (or names) to co-learn several rules simultaneously,
+        - a list/tuple of rules (or names) to co-learn several rules simultaneously,
           which are auto-named after their class (e.g. ``"cross_entropy_score"``);
         - a mapping of explicit names to rules.
```
han-ol: The term "scoring rule" is sometimes shortened to "rule" and sometimes to "score". The code reflects mostly (or exclusively?) the latter way, e.g. `BrierScore`, `merge_scores`, ..., but in the new docstrings we often use "rule".

I think we should always use "score".

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -61,14 +58,13 @@ class ModelComparisonWorkflow(BasicWorkflow):
...
 
-        - **PMP rules** (``"cross_entropy"``, ``"brier"``, ``"polynomial"``): network
-          outputs softmax probabilities over all ``num_models`` models.
-        - **Bayes factor rules** (``"exponential"``, ``"scaled_exponential"``,
-          ``"leaky_exponential"``, ``"logistic"``, ``"power_logistic"``): network outputs
-          ``num_models - 1`` log Bayes factors relative to model 0.
+        - **PMP rules** (``"cross_entropy"``, ``"brier"``, ``"polynomial"``): estimate
+          the posterior over all ``num_models`` models.
+        - **Bayes factor rules** (``"exponential"``, ``"leaky_exponential"``,
+          ``"logistic"``, ``"power_logistic"``): additionally report ``num_models - 1``
```
han-ol: Should we keep such a distinction still?

All scores now produce estimates for all of these: model_probs, log_odds and potentially log_bf.

Also the `num_models - 1` is out of date.

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -266,89 +233,22 @@ def default_adapter(
...
+            inference_conditions=inference_conditions,
         )
-        elapsed = time.perf_counter() - start_time
-        logging.info(f"Estimation completed in {format_duration(elapsed)}.")
-        return estimates
 
     def plot_default_diagnostics(
         self,
         test_data: Mapping[str, np.ndarray] | int,
+        estimates: Mapping[str, np.ndarray] | None = None,
```
han-ol, line 242: love this addition :)

bayesflow/workflows/model_comparison_workflow.py
```diff
@@ -363,20 +263,18 @@ def plot_default_diagnostics(
 
         Always produced:
 
+        - ``"confusion_and_bayes_factors"`` — the posterior model probability
+          confusion matrix (PMP view) and the pairwise Bayes factor heatmap (BF view)
+          drawn side by side in a single figure. Both views are shown regardless of the
+          scoring-rule family: merged estimates always carry ``"model_probs"``, and the
+          pairwise log odds are taken from the pooled ``"log_odds"``
+          when a Bayes factor rule contributed, otherwise derived from the pooled
```
han-ol: outdated "BF rule" exceptionalism

han-ol: text is a mixture of updated and old

han-ol (commented):
Great! Added a few comments that concern naming and doc correctness.

stefanradev93 (approved):

paul-buerkner (approved):

han-ol (approved):
There are still some issues I raised earlier with the diagnostics, but this PR has gone long enough so I will approve it at this point.

PR #700, opened 2026-05-28
Fix/conv net super init kwargs

aayush2505:
## Problem

`ConvolutionalNetwork.__init__` calls `super().__init__(**layer_kwargs(kwargs))` on line 86.
`layer_kwargs()` filters kwargs to only those accepted by Keras layers, which silently
drops arguments (e.g. `base_distribution`) that are intended for parent classes
in the inheritance chain.

All other summary networks in the same directory use `super().__init__(**kwargs)`,
which passes the full kwargs through correctly.

## Fix

Replace `**layer_kwargs(kwargs)` with `**kwargs` in the `super().__init__` call,
making `ConvolutionalNetwork` consistent with the rest of the summary networks.


stefanradev93:
Nice spot, Aayush!

PR #701, opened 2026-05-28
Semi supervised approximator with self-consistency

Kucharssim:
First implementation of the self consistency loss in the new version of bayesflow. It still has some rough edges and we need to test it more properly, but it should be fine for the basic use cases outlined in the `SelfConsistency.ipynb` example notebook. For now we only provide it in the experimental module, but the changes also touch on some other things, mainly to make it possible to run adapter, draw posterior samples, and evaluate log prob using keras operations only (so that we can sample and evaluate log prob during training.). I am also not sure about some of the naming, but first I wanted to get your impression.


Kucharssim:
the tests are not looking great :D I will check that out

Kucharssim:
I need to add some tests yet to improve coverage, but can you let me know what you think of this so far, @paul-buerkner, @stefanradev93? Thank you!

I recommend looking into the jupyter notebook under `examples/experimental/SelfConsistency.ipynb`

paul-buerkner:
Thanks! I made some edits to the notebook. Here are some more comments that perhaps you can address:

- Provide links to our papers on this method so people can read more about it. 
- Explain the high-level idea of the SC loss earlier.
- Explain why we call it "SemiSupervised".
- Results of the posterior-likelihood approximator are not discussed. The notebook just ends with training. Does it make sense to add a bit more so the notebook doesn't end up that sudden?
- rename the notebook to Self_Consistency (with _) for naming consistency

paul-buerkner (changes requested):
See my comments on the notebook above.

paul-buerkner (approved):
Thank you! Notebook looks good to me now.

paul-buerkner:
@stefanradev93 can you please review too so we can then merge?

bayesflow/experimental/self_consistency/semi_supervised_approximator.py
```diff
@@ -0,0 +1,193 @@
...
+            "composite_metrics": self.composite_metrics,
+        }
+        return base_config | serialize(config)
+
+    def build(self, data_shapes: Mapping[str, tuple[int] | Mapping[str, Mapping]]):
+        for name, appr in self.approximators.items():
+            if hasattr(appr, "built") and not appr.built:
+                appr.build(data_shapes[name])
+
+        self.built = True
```
stefanradev93: Not needed.

bayesflow/experimental/self_consistency/semi_supervised_approximator.py
```diff
@@ -0,0 +1,193 @@
...
+    def compute_metrics(self, stage: str = "training", **data: Mapping[str, Tensor]) -> dict[str, Tensor]:
+        """Compute the total loss and all sub-metrics for one training step.
+
+        Parameters
+        ----------
+        **data : Mapping[str, Tensor]
+            Named list of input datasets.
+        stage : str, optional
+            Current training stage (e.g., "training", "validation", "inference"). Controls
+            the behavior of standardization and some metric computations (default is "training").
```
stefanradev93: Not the correct order of args.

bayesflow/experimental/self_consistency/semi_supervised_approximator.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        if hasattr(self.approximators[approximator], "sample"):
+            samples = self.approximators[approximator].sample(**kwargs)
+        else:
+            raise AttributeError(f"approximator `{approximator}` does not have a `sample` method.")
+
+        summary_outputs = samples.pop("_summaries", None)
+
+        if self.adapter is not None:
+            samples = self.adapter(samples, inverse=True)
```
stefanradev93: Adapter cannot be `None` given the existing signature.

bayesflow/experimental/self_consistency/semi_supervised_approximator.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if summary_outputs is not None:
+            samples["_summaries"] = summary_outputs
```
stefanradev93: This, together with the above `pop` is unnecessary. I would just return the result from `sample`.

bayesflow/experimental/self_consistency/semi_supervised_approximator.py
```diff
@@ -0,0 +1,193 @@
...
+        self.build(keras.tree.map_structure(keras.ops.shape, adapted_data))
+
+    def _adapt_data(self, data: Mapping[str, Any]) -> Mapping[str, Any]:
+        adapted_data = {}
+
+        for name, appr in self.approximators.items():
+            if hasattr(appr, "adapter"):
+                adapted_data[name] = appr.adapter(data)
+            else:
+                adapted_data[name] = data
```
stefanradev93: An approximator always has an adapter, no?

bayesflow/approximators/approximator.py
```diff
@@ -153,9 +153,9 @@ def _prepare_conditions(
         summary_outputs : Tensor or None
             Raw summary network outputs, or ``None`` if no summary network.
         """
-
         if data is not None:
-            adapted = self.adapter(data, strict=False, **kwargs)
+            data = keras.tree.map_structure(keras.ops.convert_to_tensor, data)
+            adapted = self.adapter(data, strict=False, keras=True, **kwargs)
             adapted = keras.tree.map_structure(keras.ops.convert_to_tensor, adapted)
```
stefanradev93: Adds one unnecessary call for non-keras adapters.

bayesflow/approximators/helpers/conditions.py
```diff
@@ -94,24 +94,31 @@ def resolve(
...
-                    }
-
-                    batch_outputs = summary_network(
-                        batch_variables, **filter_kwargs(batch_kwargs, summary_network.call)
-                    )
-                    batches.append(batch_outputs)
 
-                summary_outputs = tree_concatenate(batches, axis=0)
+                if num_conditions is None:
+                    # Graph mode: shapes are symbolic, cannot iterate over batches
```
stefanradev93: Not sure I understand this comment. Can you please clarify?

bayesflow/adapters/adapter.py
stefanradev93, line 41: General remark. I think the fact that we can support keras ops in the adapter is great and I commend the change. However wouldn't it be easier to leave it to the caller to simply create the `Adapter` with a flag `use_keras` (or similiar) such that the adapter then internally knows which op to use.  From my POV, this would remove the need to shuffle a `keras` kwarg throughout the entire pipeline, and it is then very easy to reconfigure the whole adapter by doing something like `adapter.use_keras=False` if ever needed. @paul-buerkner @Kucharssim What do you think?

pyproject.toml
```diff
@@ -61,6 +61,7 @@ test = [
   "pytest-cov>=7.0.0",
   "pytest-rerunfailures>=16.1",
   "pytest-xdist>=3.8.0",
+  "tensorflow-probability>=0.25.0"
```
stefanradev93, line 65: Not sure I like having tf prob as a dependency. If anything, it should be an optional dependency. By the same token, we can argue for `numpyro` or similar as a dep.

stefanradev93 (changes requested):

bayesflow/adapters/adapter.py
Kucharssim: Yes, that was the other way to do it. I am generally fine implementing it either way!

pyproject.toml
Kucharssim: Is it used only for testing, and it is specified among the `test` dependencies. Are these not optional?

I am happy to rewrite the tests without relying on this, but I thought it would be nice to cover the use-case of users specifying prior/likelihood using tensorflow distributions, as it makes using self-consistency much easier...

paul-buerkner:
About adapter design, both is fine by me. If you two agree the other one is
nicer, feel free to change it

Simon Kucharsky ***@***.***> schrieb am Mi. 8. Juli 2026 um
21:59:

pyproject.toml
stefanradev93: Missed that on the github view, happy to keep it as is.

bayesflow/approximators/helpers/conditions.py
Kucharssim: I've expanded on the comment, is it now clearer?

bayesflow/adapters/transforms/to_dict.py
```diff
@@ -28,12 +29,12 @@ def forward(self, data, **kwargs) -> dict[str, np.ndarray]:
                 if value.dtype == "category":
                     value = pd.get_dummies(value)
 
-                value = np.asarray(value).astype("float32", copy=False)
+                value = ops.convert_to_tensor(value, dtype="float32")
```
stefanradev93: Shouldn't we use floatx here?

bayesflow/adapters/adapter.py
```diff
@@ -100,7 +103,7 @@ def get_config(self) -> dict:
         return serialize(config)
 
     def forward(
-        self, data: dict[str, Any], *, log_det_jac: bool = False, **kwargs
```
stefanradev93, line 103: `Any` from `typing` is the recommended hint, afaik.

bayesflow/adapters/adapter.py
```diff
@@ -60,14 +59,18 @@ class Adapter(MutableSequence[Transform]):
     ----------
     transforms : Sequence[Transform], optional
         The sequence of transforms to execute.
+    device_name: str, optional
+        Passed to keras.device to specify on which device to apply the adapter.
     """
 
-    def __init__(self, transforms: Sequence[Transform] | None = None):
+    def __init__(self, transforms: Sequence[Transform] | None = None, device_name: str = "cpu"):
```
stefanradev93: Should we just go with `device` as in canonical in torch?

bayesflow/approximators/continuous_approximator.py
```diff
@@ -222,6 +222,7 @@ def sample(
         sample_shape: Literal["infer"] | Tuple[int] | int = "infer",
         return_summaries: bool = False,
         seed: int | keras.random.SeedGenerator | None = None,
+        numpy: bool = True,
```
stefanradev93: Missing in docstring and can be dangerously mistaken for the library. Perhaps `to_numpy` / `as_numpy`?

bayesflow/approximators/continuous_approximator.py
```diff
@@ -299,7 +302,7 @@ def sample(
 
         return samples
 
-    def log_prob(self, data: Mapping[str, np.ndarray], **kwargs) -> np.ndarray:
+    def log_prob(self, data: Mapping[str, np.ndarray], numpy: bool = True, **kwargs) -> np.ndarray:
```
stefanradev93: Same as above.

bayesflow/experimental/self_consistency/semi_supervised_approximator.py
```diff
@@ -0,0 +1,190 @@
...
+        Generates samples from one of the approximator (specified by its name) given input conditions.
+        **kwargs are passed to the approximator.sample method.
+        """
+        if hasattr(self.approximators[approximator], "sample"):
+            samples = self.approximators[approximator].sample(**kwargs)
+        else:
+            raise AttributeError(f"approximator `{approximator}` does not have a `sample` method.")
+
+        samples = self.adapter(samples, inverse=True)
+        if numpy:
```
stefanradev93: Same comment regarding naming.

bayesflow/utils/tensor_utils.py
```diff
@@ -37,6 +37,14 @@ def concatenate_valid_shapes(tensor_shapes: Sequence[Shape | None], axis: int =
     return output_shape
 
 
+def repeat_valid(x: Tensor, repeats: int, axis: int = 0) -> Tensor | None:
```
stefanradev93: type hint should be `Tenor | None`

tests/test_adapters/test_adapters.py
```diff
@@ -61,14 +62,16 @@ def test_constrain():
 
     ad = (
         Adapter()
+        .to_array()
+        .convert_dtype("float64", "float32")
         .constrain("x_lower_cont", lower=0)
         .constrain("x_upper_cont", upper=0)
         .constrain("x_both_cont", lower=0, upper=1)
-        .constrain("x_lower_disc1", lower=0, inclusive="lower")
+        .constrain("x_lower_disc1", lower=0, inclusive="lower", epsilon=1e-6)
```
stefanradev93: Why is this not default if it's important for stability?

bayesflow/adapters/transforms/random_subsample.py
```diff
@@ -26,20 +29,19 @@ def __init__(
...
-    def forward(self, data: np.ndarray, **kwargs) -> np.ndarray:
-        max_sample_size = data.shape[self.axis]
+    def forward(self, data: Tensor, **kwargs) -> Tensor:
+        max_sample_size = ops.shape(data)[self.axis]
 
         if isinstance(self.sample_size, int):
             sample_size = self.sample_size
         else:
-            sample_size = np.round(self.sample_size * max_sample_size)
+            sample_size = int(ops.round(self.sample_size * max_sample_size))
```
stefanradev93, line 38: Why does this even need an `ops` call?

stefanradev93 (changes requested):
Great work, Simon! I think we are almost there. Did you do some runtime benchmarks with the new (CPU / GPU) adapter vs. the old numpy one?

bayesflow/adapters/adapter.py
Kucharssim: I am fine with that, but since we are only using it to pass it forward to keras.device, I figured I can keep consistent with that

bayesflow/adapters/transforms/random_subsample.py
Kucharssim: We can keep the numpy call, I just removed it so that we can remove the numpy import. I don't mind either way...

bayesflow/adapters/transforms/to_dict.py
Kucharssim: Yes, thanks!

tests/test_adapters/test_adapters.py
Kucharssim: Good point, will change that

bayesflow/adapters/adapter.py
Kucharssim: Ok, changed throughout the adapter

Kucharssim:
> Did you do some runtime benchmarks with the new (CPU / GPU) adapter vs. the old numpy one?

I've only run benchmarks comparing CPU vs GPU on the new adapter. I will compare to the old numpy as well

Issue #714, opened 2026-07-03
Time2Vec sequence length dependent spacing


han-ol:
This concerns `TimeSeriesTransformer` internals when no explicit time stamps are passed (`time_axis=None`).
then `Time2Vec.call` builds times as:

```python
t = keras.ops.linspace(0, keras.ops.shape(x)[1], keras.ops.shape(x)[1], dtype=x.dtype)
```

`linspace(0, T, T)` spans [0, T] with T points, so the spacing is T/(T-1) and the time
assigned to a given step changes with the total length `T`:

- `T=5` -> `[0, 1.25, 2.5, 3.75, 5.0]`
- `T=9`, first 5 -> `[0, 1.125, 2.25, 3.375, 4.5]`

`arange(T)` -> `[0, 1, ..., T-1]` makes step `i` map to time `i` regardless of `T`, so a
right-padded sequence keeps the same times for its real steps as the unpadded one.
The reason I care for this little detail is that I am looking at masking correctly when a time-series has varying length within a single batch: #713

**Is there a good reason not to use the simple thing (`arange`) here?**

stefanradev93:
`arange` is definitely the better function to use here. Feel free to make the change and fiddle with the time embeddings.

PR #715, opened 2026-07-03
arange in time2vec, lstm and gru

han-ol:
Addresses #714.


stefanradev93 (approved):
The change is good! The tests are failing due to a preexisting problem with `jit_compile` it seems.

han-ol:
I added a test for the change that passes, checking basically whether #714 is fixed. If the CI tests are failing because of upstream issues, this is ready for review/merge.

han-ol:
thank you for the quick review!

PR #717, opened 2026-07-05
Diffusion Transformer

arrjon:
Due to popular request, I added support for a new `TimeTransformer` subnet and introduced significant improvements to the masking and sampling logic in the diffusion models.

We can now replace the MLP with a transformer and employ masking on targets. The transformer constructs the correct attention masks, which is the reason a transformer should be used instead of an MLP. I validated the new subnet on the 10 low-dim benchmarks, and it is at least as good as the MLP for enough simulation budget, but roughly 3 times slower.

I also found a minor bug in the sampling of the Stable Consistency Model, where time was updated too late.


stefanradev93:
Great addition, Jonas! I cleaned up and removed some redundant code. Additionally:

- Since the new subnet uses mostly DiT logic, why not call it `DiffusionTransformer` (with a slight abuse of terminology, since we will be using it for all free-form nets)?
- Can the internal logic handle both `(B, T, C)` (not summarized) and `(B, C)` (pre-summarized) conditions?
- Can we add a small usage example to the diffusion tutorial)?

See some small questions in the code too. Will try to fix https://github.com/bayesflow-org/bayesflow/pull/718 before merging here.

bayesflow/networks/helpers/blocks/diffusion_transformer_block.py
```diff
@@ -0,0 +1,193 @@
...
+        mod = keras.ops.expand_dims(mod, axis=1)
+        shift_attn, scale_attn, gate_attn, shift_ffn, scale_ffn, gate_ffn = keras.ops.split(mod, 6, axis=-1)
+
+        attn_in = self.modulate(self.attn_norm(x, training=training), shift_attn, scale_attn)
+        kv = attn_in if conditions is None else keras.ops.concatenate([attn_in, conditions], axis=1)
+        attn_out = self.attn(attn_in, kv, attention_mask=attention_mask, training=training)
+        h = x + gate_attn * self.attn_dropout(attn_out, training=training)
+
+        ffn_in = self.modulate(self.ffn_norm(h, training=training), shift_ffn, scale_ffn)
+        h = h + gate_ffn * self.ffn(ffn_in, training=training)
```
stefanradev93, line 157: Is the `gate_ffn` really needed given that our ffn is already gated?

bayesflow/networks/subnets/transformer/time_transformer.py
```diff
@@ -0,0 +1,406 @@
...
+        for block in self.blocks:
+            h = block(
+                (h, base_mod),
+                conditions=cond_h,
+                attention_mask=attention_mask,
+                update_mask=update_mask,
+                training=training,
+            )
+
+        final_mod = keras.ops.expand_dims(self.final_ada_ln(keras.ops.silu(t_emb)), axis=1)
```
stefanradev93: Is another `silu` really needed here given that `t_emb` has already passed through a `silu` (L234) and an mlp (L235)?

bayesflow/networks/inference/consistency/stable_consistency_model.py
```diff
@@ -95,9 +100,8 @@ def __init__(
         self.p_mean = float(kwargs.get("p_mean", -1.0))
         self.p_std = float(kwargs.get("p_std", 1.6))
         self.c = float(kwargs.get("c", 0.1))
-        self.drop_cond_prob = drop_cond_prob
-        self.unconditional_mode = False
         self.drop_target_prob = float(kwargs.get("drop_target_prob", 0.0))
+        self.drop_missing_prob = float(kwargs.get("drop_missing_prob", 0.0))
```
stefanradev93: Why not make them regular constructor arguments instead of hiding them (as in diffusion model)? Why the float conversion (I know AI loves it, but it's unnecessary clutter)? And also give them meaningful names (as opposed to `c`, `p_mean`, etc).

bayesflow/networks/inference/consistency/consistency_model.py
```diff
@@ -119,9 +125,8 @@ def __init__(
         self.c_huber = None
         self.c_huber2 = None
         self.unique_n = None
-        self.drop_cond_prob = drop_cond_prob
-        self.unconditional_mode = False
         self.drop_target_prob = float(kwargs.get("drop_target_prob", 0.0))
+        self.drop_missing_prob = float(kwargs.get("drop_missing_prob", 0.0))
```
stefanradev93: Same as with stable. Why not regular, user-exposed kwargs? Why the float conversion?

bayesflow/networks/inference/consistency/consistency_model.py
stefanradev93: Same as with stable? Why not regular user-facing kwargs? Why is the float conversion of a float necessary? (applies throughout).

stefanradev93 (commented):
Left a few more comments.

bayesflow/networks/helpers/blocks/diffusion_transformer_block.py
arrjon: Not sure: the reasoning is that at init the FFN's contribution to the residual stream is near zero

bayesflow/networks/subnets/transformer/time_transformer.py
arrjon: If we removed it, we would also need to drop `time_mlp_out` as we would just concatenate Dense layers without activation in between. So I would keep it.

bayesflow/networks/inference/consistency/stable_consistency_model.py
arrjon: I am hiding them here, because dropping and masking with consistency models is still experimental, and I am not sure how good it is.

arrjon: changing the names of `c`, `p_mean` etc. would be a breaking change. Should we still do it?

arrjon:
> * Since the new subnet uses mostly DiT logic, why not call it `DiffusionTransformer` (with a slight abuse of terminology, since we will be using it for all free-form nets)?

Agreed!

> * Can the internal logic handle both `(B, T, C)` (not summarized) and `(B, C)` (pre-summarized) conditions?

The internal logic is made for `(B, T, C)` or `(B, T)`, so with and without condition, but always with a diffusion time

> * Can we add a small usage example to the diffusion tutorial)?

I will add it.

arrjon:
I added the example to the tutorial on dIffusion models and simplified how to pass fixed targets so users do not have to standardize themselves.

bayesflow/networks/subnets/transformer/time_transformer.py
stefanradev93: Ok, I see you dropped it. I guess it didn't matter for performance?

bayesflow/networks/subnets/transformer/diffusion_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+                )
+
+        # DiT-style shared time MLP on top of the Fourier features
+        self.time_mlp = keras.Sequential(
+            [
+                keras.layers.Dense(self.width, activation="silu", kernel_initializer=kernel_initializer),
+                keras.layers.Dense(self.width, activation="silu", kernel_initializer=kernel_initializer),
+            ]
+        )
+        self.ada_ln_shared = keras.layers.Dense(6 * self.width, kernel_initializer="zeros", bias_initializer="zeros")
```
stefanradev93: Not a layer norm, so naming could be a tad confusing.

bayesflow/networks/subnets/transformer/diffusion_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+                    include_identity=True,
+                )
+
+        # DiT-style shared time MLP on top of the Fourier features
+        self.time_mlp = keras.Sequential(
+            [
+                keras.layers.Dense(self.width, activation="silu", kernel_initializer=kernel_initializer),
+                keras.layers.Dense(self.width, activation="silu", kernel_initializer=kernel_initializer),
+            ]
+        )
```
stefanradev93: Maybe use our `MLP` class with res?

bayesflow/networks/subnets/transformer/diffusion_transformer.py
```diff
@@ -0,0 +1,409 @@
...
+            cond_h = self.condition_proj(keras.ops.expand_dims(conditions, axis=-1))
+            cond_h = cond_h + self.condition_id[None]
+            cond_h = cond_h + self.state_embedding(None, condition_mask, num_conditions, batch_size, latent=False)
+            cond_h = self.condition_norm(cond_h, training=training)
+
+        t = keras.ops.reshape(t, (keras.ops.shape(t)[0], -1))[:, :1]
+        t_emb = self.time_mlp(self.time_emb(t))
+
+        # Shared adaLN-single modulation, computed once and reused by every block.
+        base_mod = self.ada_ln_shared(t_emb, training=training)
```
stefanradev93: No need for `training` arg, since not a layer norm (unless it's a but and it was supposed to be a layer norm).

bayesflow/networks/defaults.py
```diff
@@ -25,6 +25,20 @@
     }
 )
 
+DIFFUSION_TRANSFORMER_DEFAULTS = MappingProxyType(
+    {
+        "widths": (128, 128, 128, 128, 128),
```
stefanradev93, line 30: The default looks pretty beefy with a depth of 5. Necessary for decent performance?

arrjon: with 4 layers, we still have a bit more parameters than the current TimeMLP, so that might be okay

arrjon: but as it will be used for learning lots of conditions etc maybe a beefy network is a good default. Otherwise one should use the MLP anyway

stefanradev93: you are right.

PR #718, opened 2026-07-05
Fix torch compiler not found in CI workflow

stefanradev93:
Attempts a fix for the compiler not found on path error in torch CIs.


stefanradev93:
@LarsKue Any ideas why `jit_compile` is suddenly failing in the CIs? Seems to be passing locally.

LarsKue:
@stefanradev93 I think torch.compile is just not that compatible with the autograd utilities yet.

PR #719, opened 2026-07-08
fix masking in compositional

arrjon:
I minor issue we missed yesterday with the transformer backbone is that in compositional mode, masks need to be broadcasted to the correct shape. Adapted the tests to verify the new implementation works now.


arrjon:
I was unhappy with the names of the masks and probabilities of dropping conditions etc. Now they feel more intuitive.

stefanradev93 (approved):
I agree, these changes are very welcome.

Issue #709, opened 2026-06-12
[BUG] C2ST not showing in API docs


maltelueken:
**Describe the bug**
In the online API docs (version 2.0.12), the C2ST is not showing under `bayesflow.diagnostics.metrics`.

han-ol:
Thank you for raising this issue!

PR #727, opened 2026-07-23
Submodule pages for diagnostics and coupling flow components 

han-ol:
Addresses #709 and makes the diagnostics API pages cleaner by separating out module pages for `metrics` and `plots`.

Also adds auto-generated API references for the  coupling flow components, i.e. transforms and couplings.


stefanradev93:
Thanks, Hans! I don't see any additions to coupling flow files.

han-ol:
I force-pushed, because I had previously added some local experimentation files by accident. Undid that.

Now this is actually ready for review.

Issue #726, opened 2026-07-23
Standardization is not padding invariant


han-ol:
Context: In time series the second to last axis holds the trajectory. In standardization all but the last axis are treated as sample dimensions to reduce over to compute running moments during training.

Since the moment computation currently does not take any masks into account, it is not invariant to pads. I think it should be!

stefanradev93:
Also relates to an older https://github.com/bayesflow-org/bayesflow/issues/415.

PR #728, opened 2026-07-23
Padding mask invariance of Standardize

han-ol:
Addresses #726.

### DIFF
--- a/bayesflow/networks/helpers/standardization/standardize.py
+++ b/bayesflow/networks/helpers/standardization/standardize.py
@@ -26,7 +26,9 @@ class Standardize(keras.Layer):
         self.count = None
 
     def compute_mask(self, inputs, mask=None):
-        # The mask is consumed here (moment computation). Nothing is propagated downstream.
+        # `mask` (magic keyword in Keras) is terminated here by `return None`
+        # to prevent warnings about inability to inject it downstream.
+        # We explicitly pass mask and do not rely on having it travel with as a tensor attribute.
         return None
 
     def moving_std(self, index: int) -> Tensor:
--- a/bayesflow/networks/helpers/standardization/standardize.py
+++ b/bayesflow/networks/helpers/standardization/standardize.py
@@ -21,11 +21,13 @@ class Standardize(keras.Layer):
         """
         super().__init__(**layer_kwargs(kwargs))
 
-        self.supports_masking = True
-
         self.moving_mean = None
         self.moving_m2 = None
         self.count = None
+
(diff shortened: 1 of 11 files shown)


bayesflow/networks/helpers/standardization/standardize.py
```diff
@@ -145,14 +151,21 @@ def _update_moments(self, x: Tensor, index: int):
...
-        batch_count = keras.ops.cast(keras.ops.prod(keras.ops.shape(x)[:-1]), self.count[index].dtype)
-
-        # Compute batch mean and M2 per feature
-        batch_mean = keras.ops.mean(x, axis=reduce_axes)
-        batch_m2 = keras.ops.sum((x - expand_left_as(batch_mean, x)) ** 2, axis=reduce_axes)
+        if mask is None:
+            mask = keras.ops.ones(keras.ops.shape(x)[:-1], dtype=x.dtype)
+        else:
+            mask = keras.ops.cast(keras.ops.broadcast_to(mask, keras.ops.shape(x)[:-1]), x.dtype)
+        mask = mask[..., None]
```
stefanradev93: Multiplying with ones seems wasteful for large tensors unless we make a leap of faith that the compiler will optimize it away.

stefanradev93:
That would work, but the mask needs to be passed by approximators to standardizer too.

bayesflow/networks/helpers/standardization/standardize.py
```diff
@@ -25,6 +25,10 @@ def __init__(self, **kwargs):
         self.moving_m2 = None
         self.count = None
 
+    def compute_mask(self, inputs, mask=None):
+        # The mask is consumed here (moment computation). Nothing is propagated downstream.
+        return None
+
     def moving_std(self, index: int) -> Tensor:
```
arrjon, line 34: why do we need this?

han-ol: The mask argument is magic in keras and this is the way to tell it to not warn about not being able to inject mask arguments in later keras.layers.

han-ol: I added a commit so the inline comment explains it better. The keras docs are lacking on this. It is however a real behavior. I found out about it by seeing the warning and then looking at keras source code.

The alternatives to defining such methods in each class are:

* excepting the confusing warning
* use a different keyword arg name other than `mask`. But not sure what would be a good candidate.
* outsource it to a `MaskConsuming` mixin class. Then, this is not repeated across multiple classes. This might be my favorite.

What do you think of all this?

PR #738, opened 2026-07-25
add output projector option to diffusion models

arrjon:
The output projection layer was incompatible with the transformer backbone in the diffusion type models when employed together with masking of targets. I added an optional `output_projector` parameter to the constructors of these models, allowing users to specify a custom output projection layer. If not provided, a default dense layer is used as before, except when the subnet is a `"diffusion_transformer"`, in which case an identity layer is used.

### DIFF
--- a/bayesflow/networks/inference/diffusion/diffusion_model.py
+++ b/bayesflow/networks/inference/diffusion/diffusion_model.py
@@ -291,7 +291,8 @@ class DiffusionModel(InferenceNetwork):
         self.subnet.build((xz_shape, time_shape, conditions_shape))
         out_shape = self.subnet.compute_output_shape((xz_shape, time_shape, conditions_shape))
 
-        self.output_projector.build(out_shape)
+        if not self.output_projector.built:
+            self.output_projector.build(out_shape)
 
     def get_config(self):
         base_config = super().get_config()
--- a/bayesflow/networks/inference/diffusion/diffusion_model.py
+++ b/bayesflow/networks/inference/diffusion/diffusion_model.py
@@ -83,10 +83,6 @@ class DiffusionModel(InferenceNetwork):
         Probability of marking each condition as missing during training, so the network learns
         to handle missing conditions. Only takes effect when the subnet accepts
         ``infer_target_mask`` (e.g. ``diffusion_transformer``). Default is 0.0.
-    output_projector : keras.layers.Layer, optional
-        The output projector is applied to the subnet output. Default is a dense layer to map
-        the subnets output to the inference variable dimension.
-        Projector is omitted for ``"diffusion_transformer"``.
     **kwargs
         Additional keyword arguments passed to the base ``InferenceNetwork``.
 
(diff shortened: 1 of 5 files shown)


stefanradev93:
Since we are now more aware of breaking changes and this is breaking, wouldn't it be better to check in `build()` if the subnet is a diffusion transformer and select identity there? So far, we don't have other incompatible layers.

arrjon:
I removed the `output_projector` and instead check it in the built function now. Trained transformer backbones will still break, but they were only introduced with the last version. Is this acceptable? The way I implemented it now, a user can define a flag in their own subnets to avoid having a dense layer if this is undesired.

stefanradev93:
What if we checked if the subnet output shape is already the same as xz_shape and only if it's not, we add an output projector? We still check for diffusion transformer incompatibility. Should cover all existing cases and most future cases.

arrjon:
Agreed, we can do this. Will still be a breaking change for the transformer backbone (and also on inference tasks where the output of the MLP has the same width as the parameter dimension, unlikely but can happen).

arrjon:
Okay, now it is non-breaking, but I needed to add a `from_config` method.

stefanradev93:
Do we even need the `self._project_output` flag? Why not simply choose between a projector or an identity?

arrjon:
`self._project_output` is only there for backward compatibility and makes it a non-breaking change.

stefanradev93:
But the attribute is not there in `x.13`. A network with an output projector from `x.13/12...` will still load properly. What's the use case for the extra check? Or do you wanna have the extra kwarg as a hidden flag for a special use case?

arrjon:
Older networks that do not have the attribute stored will default to True:
``` return super().from_config({"project_output": True} | config ```

Newer networks will either have the projector or not depending on whether the subnet has the correct dimensions or not. This is the only purpose for the flag. If we removed it and only check for the dimensions, this would be a breaking change, as old networks would just drop their projection layer.

**Comment by arrjon** [comment]
I reverted to mostly non-breaking to keep the code more readable. Diffusion transformer backbone might fail silently and if the MLP width is exactly the parameter dimension.

File: bayesflow/networks/inference/diffusion/diffusion_model.py [hunk 1]
```diff
@@ -268,14 +268,20 @@ def build(self, xz_shape: Shape, conditions_shape: Shape = None):
...
-        units = 1 if self._prediction_type == "potential" else xz_shape[-1]
-        self.output_projector = keras.layers.Dense(units=units, bias_initializer="zeros")
-
         # construct input shape for subnet and subnet projector
         time_shape = (xz_shape[0], 1)
         self.subnet.build((xz_shape, time_shape, conditions_shape))
         out_shape = self.subnet.compute_output_shape((xz_shape, time_shape, conditions_shape))
 
+        units = 1 if self._prediction_type == "potential" else xz_shape[-1]
+        if out_shape[-1] != units:
```
stefanradev93, line 277: Isn't this true for potential and creates a projector anyways?

arrjon: Most likely yes, but I think there is no harm in doing it like this. Maybe the subnet already has singluar output for a specific reason...

Issue #737, opened 2026-07-25
[BUG] RatioApproximator fails on incomplete batches


stefanradev93:
`RatioApproximator.fit()` crashes when the last batch results in $B < K$ (batch size smaller than requested number of contrastive samples).

For example, 100 simulations with `batch_size=32` leaves a final batch of 4 and raises:

```text
TypeError: Cannot concatenate arrays with shapes that differ in dimensions other than the one being concatenated: concatenating along dimension 2 for shapes (4, 3, 2), (4, 16, 500).
```

For the final batch, `_sample_from_batch()` reduces the number of contrastive samples to batch_size - 1, but compute_metrics() still uses the configured K when tiling the conditions and calculating log_K. This produces incompatible tensor shapes.

Incomplete batches should be supported by using `min(K, current_batch_size - 1)`. Alternatively, the last (incomplete) batch should be dropped like PyTorch does.

PR #743, opened 2026-08-01
Fix ratio approximator bug when batch size is smaller than num contrastive samples

stefanradev93:
Addresses #737 by using `min(K, batch_size - 1)` in `_sample_from_batch`.


stefanradev93:
@LarsKue We should consider letting our datasets drop the last batch from each epoch if `epochs % batch_size != 0`, as PyTorch data loaders do to avoid situations like this.

PR #746, opened 2026-08-05
Add Autoregressive Approximators and Decoders (WIP)

stefanradev93:
This PR adds a new `AutoregressiveApproximator` that can be used to learn the full joint distribution of time-varying targets. It is compatible with all of our inference networks, but required some additional machinery for (efficient) autoregressive sampling and log prob estimation. Currently, the approximator can be minimally trained like so:

```
encoder = bf.networks.RecurrentNetwork(return_sequences=True)
decoder = bf.networks.RecurrentDecoder()

inference_network = bf.networks.CouplingFlow()

approximator = bf.AutoregressiveApproximator(
    inference_network=inference_network,
    encoder_network=encoder,
    decoder_network=decoder,
    adapter=adapter,
    standardize="all"
)

approximator.compile(optimizer="adamw")

h = approximator.fit(
    dataset=train_dataset,
    validation_data=val_dataset,
    epochs=epochs
)
```

it is compatible with any many-to-many summary network and comes with two "decoder networks": rnn-based and transformer-based which are both trained using teacher forcing and know how to cache computations for sampling.

Let's discuss some of the naming. For example, the `encoder_network` can just as well be called a `summary_network` with the caveat that it does "summarize" the time dimension. Additionally, you may have a better name for the `decoder_network`. The remaining TODOs are"

- [x] Add an option to make the recurrent decoder deep
- [x] Add a few more tests (especially in terms of masking guarantees) and a few personal benchmarks
- [x] Polish the `Autoregressive_Smoothing` notebook to become an actual tutorial
- [x] Add an entry to the user guide
~- [ ] Add an option for EOS tokens for targets with an unknown time horizon~


bayesflow/approximators/autoregressive_approximator.py
```diff
@@ -0,0 +1,265 @@
...
+        )
+
+        inference_kwargs = {key: value for key, value in kwargs.items() if key != "batch_size"}
+        step_log_prob = self.inference_network.log_prob(
+            inference_variables,
+            conditions=conditions,
+            **inference_kwargs,
+        )
+        log_det = keras.ops.convert_to_tensor(adapter_log_det.get("inference_variables", 0.0))
+        step_log_prob = step_log_prob + log_det + standardizer_log_det
```
LarsKue: This leaves a shape error if the adapter yields a log_det, since the `adapter_log_det` is of shape `(B,)` but the `step_log_prob` is of shape `(B, T)`.

bayesflow/approximators/helpers/samplers.py
```diff
@@ -152,3 +159,90 @@ def get_config(self) -> dict:
...
+    """Sample trajectories with an incrementally cached autoregressive decoder."""
+
+    def _sample_batch(
+        self,
+        *,
+        inference_network: keras.Layer,
+        decoder_network: keras.Layer,
+        num_samples: int,
+        conditions: Tensor | None,
+        sample_shape,
```
LarsKue: Seems the `sample_shape` is ignored in this function?

LarsKue (commented):
I did a rough scan and from my side, everything looks good from a design standpoint. Some easily fixable points in the comments.

bayesflow/approximators/helpers/samplers.py
stefanradev93: Fixed.

bayesflow/approximators/autoregressive_approximator.py
stefanradev93: Fixed.

PR #748, opened 2026-08-07
Update `resolve_seed` for improved logic with seed generators

LarsKue:


bayesflow/approximators/ensemble_approximator.py
```diff
@@ -270,11 +276,20 @@ def sample(
...
         weights = self._resolve_member_weights(member_weights)
-        counts = multinomial_allocation(weights, num_samples, seed=seed)
-        alloc = {name: count for name, count in counts.items() if count > 0}
+        names = tuple(weights.keys())
+        probs = np.array(list(weights.values()))
+
+        # Sample counts from multinomial
+        K = len(probs)
+        logits_broadcast = keras.ops.broadcast_to(keras.ops.expand_dims(keras.ops.log(probs), axis=0), (num_samples, K))
+        cat_indices = keras.ops.squeeze(keras.random.categorical(logits_broadcast, num_samples=1, seed=seed), axis=-1)
```
stefanradev93: This should use `seed_generator`.

bayesflow/approximators/ensemble_approximator.py
```diff
@@ -270,11 +276,20 @@ def sample(
                 fn=lambda name, a: a.sample(num_samples=num_samples, conditions=conditions, split=split, **kwargs),
```
stefanradev93: `seed_generator` needs to be passed here as well, so it should be resolved before this call.

bayesflow/distributions/diagonal_normal.py
```diff
@@ -99,7 +95,7 @@ def log_prob(self, samples: Tensor, *, normalize: bool = True) -> Tensor:
 
     @allow_batch_size
     def sample(self, batch_shape: Shape, seed: int | keras.random.SeedGenerator | None = None) -> Tensor:
-        sg = resolve_seed(seed) or self.seed_generator
+        sg = resolve_seed(seed, self.seed_generator)
```
stefanradev93: We are not abbreviating it anywhere else, why here?

bayesflow/distributions/diagonal_normal.py
```diff
@@ -43,9 +42,6 @@ def __init__(
             Default is 1.0.
         trainable_parameters : bool, optional
             Whether to treat the mean and standard deviation as learnable parameters. Default is False.
-        seed_generator : keras.random.SeedGenerator, optional
```
stefanradev93, line 46: Why disallow this and then still create an attribute? We may as well then not entertain an attribute at all, if the only way to control seeding would be through `sample`. Applies to all distros.

bayesflow/distributions/diagonal_normal.py
```diff
@@ -111,7 +107,6 @@ def get_config(self):
             "mean": self.mean,
             "std": self.std,
             "trainable_parameters": self.trainable_parameters,
-            "seed_generator": self.seed_generator,
```
stefanradev93, line 115: Also kills backward compatibility for a reason that doesn't seem good enough to me.

stefanradev93: I understand that we may want to keep this in line with how we initialize approximators, in which case I would simply add a backward compatibility `from_config` that pops out a seed gen.

stefanradev93 (changes requested):
The cleanup is helpful. I've requested a few changes. Can we expand our test suite to cover  verify sample determinism for all approximators supporting `.sample()`?

bayesflow/distributions/diagonal_normal.py
LarsKue: I agree that we can remove this inconsistency.

bayesflow/distributions/diagonal_normal.py
LarsKue: Agreed that backward compatibility is a concern here that I did not address adequately. I disallowed the constructor argument for the sake of consistency, but we could add it everywhere as well.

bayesflow/approximators/ensemble_approximator.py
LarsKue: Fixed

bayesflow/approximators/ensemble_approximator.py
LarsKue: Fixed

bayesflow/distributions/diagonal_normal.py
LarsKue: Fixed through a config pop for now.

bayesflow/distributions/diagonal_normal.py
LarsKue: We only need to seed where we have randomness, so that is only in `sample`. Do you have a concrete use case where seeding from the constructor is much more ergonomic than seeding in sample?

stefanradev93 (approved):

PR #750, opened 2026-08-12
Replace hardcoded float32 values with floatx()

stefanradev93:
Replaces hardcoded `float32` values (e.g., in distribution parameter inits), integrators to use `keras.backend.floatx()`. Does not touch adapter transforms yet, as these will be superseded by #747.


LarsKue (approved):
I'm in favor of this change. @stefanradev93 Is there a reason to use `keras.backend.floatx` instead of `keras.config.floatx`? I'm ok with either, but it seems that `keras.config` is the intended exported API because usually we should avoid using `keras.backend` for backend-agnostic code.

PR #755, opened 2026-08-21
GraphicalApproximator: enumerate all factorizations to infer the best inverse

daniel-habermann:
This PR changes how the default inverse factorization is selected. Previously, we used the "inverse Stuhlmüller" method, where latent nodes were added to the inverse graph in topological order. This is a reasonable default for many models, but there are some situations in which it selects a non-optimal inverse factorization.

Consider the following three-level simulator, modelling bird counts within regions within squares over several years:

```python
import numpy as np
from bayesflow.experimental.graphical_simulator import GraphicalSimulator

def sample_globals():
    mu = np.random.normal(0, 5)
    sigma_region = np.abs(np.random.normal(0, 1))
    sigma_square = np.abs(np.random.normal(0, 1))
    sigma_year = np.abs(np.random.normal(0, 1))

    return {
        "mu": mu,
        "sigma_region": sigma_region,
        "sigma_square": sigma_square,
        "sigma_year": sigma_year,
    }


def sample_regions(mu, sigma_region):
    alpha_region = np.random.normal(mu, sigma_region)

    return {"alpha_region": alpha_region}


def sample_squares(alpha_region, sigma_square):
    alpha_square = np.random.normal(alpha_region, sigma_square)

    return {"alpha_square": alpha_square}


def sample_years(sigma_year):
    beta_year = np.random.normal(0, sigma_year)

    return {"beta_year": beta_year}


def sample_obs(alpha_square, beta_year):
    count_offset = np.random.choice([0, np.log(2)])
    count = np.random.poisson(np.exp(alpha_square + beta_year + count_offset))

    return {"count": count, "count_offset": count_offset}


def meta_fn():
    N_regions = 82
    N_squares = np.random.randint(1, 251)
    N_years = 31

    return {"N_regions": N_regions, "N_squares": N_squares, "N_years": N_years}


def breeding_bird_survey_simulator():
    r"""

              globals
             /    |   \
           /      |    \
       regions    |   years    
           \      |     |
             \    |     |
             squares    |
                   \    |
                     \  /
                     obs

    """

    simulator = GraphicalSimulator(meta_fn=meta_fn)

    simulator.add_node("globals", sample_fn=sample_globals)
    simulator.add_node("regions", sample_fn=sample_regions, reps="N_regions")
    simulator.add_node("squares", sample_fn=sample_squares, reps="N_squares")
    simulator.add_node("years", sample_fn=sample_years, reps="N_years")
    simulator.add_node("obs", sample_fn=sample_obs)

    simulator.add_edge("globals", "regions")
    simulator.add_edge("globals", "squares")
    simulator.add_edge("globals", "years")
    simulator.add_edge("regions", "squares")
    simulator.add_edge("squares", "obs")
    simulator.add_edge("years", "obs")

    return simulator

```

Using the inverse Stuhlmüller heuristic, the inverse graph is constructed by adding nodes in the follwing order: `globals`, `regions`, `squares`, `years`. In this factorization, only `globals` and `years` are independently amortizable.

However, if nodes were added to the inverse graph in this ordering: `globals`, `years`, `squares`, `region`, the resulting factorization allows `globals`, `regions` and `squares` to be independently amortizable, which is much better as only the 31 years need to be inferred sequentially.

This PR adds an `factorizations.py` module that enumerates all possible factorizations. This requires `N!` graph inversions, where `N` is the number of latent nodes in the simulation graph. Even for complex models, `N` is typically well below 10, so enumerating all is computationally negligible. 

The `invert()` method of `ExpandedGraph` has a new `node_ordering` argument that allows to compute the inverse graph for any order of latent nodes. If this argument is empty (the default), all possible factorizations are enumerated internally and the one selected that minimizes the expected number of groups that need to be inferred sequentially. The expected number of groups for each node is known from the specified simulator, i.e, in the example above, `N_regions=82, N_squares=Uniform(1, 250) and N_years=31`. This means that leaving `years` to be inferred sequentially is much better than leaving `regions` and `squares` to be inferred sequentially.


paul-buerkner (approved):
Looks good to me. merging now.

Issue #757, opened 2026-08-26
[FEATURE] Pairs recovery plots


paul-buerkner:
Today I was thinking of a new kind of diagnostic plot that we may want to add, which shows recovery plots in a matrix (pairs) format.

On the diagonal are our standard recovery plots (ground-truth vs. estimate). Above the diagonal are pairwise plots of estimates (estimate of parameter 1 vs. estimate of parameter 2). Below the diagonal are pairwise plots of ground truths (truth of parameter 1 vs. truth of parameter 2). This enables quick understanding of dependencies of parameters in both prior and posterior and helps identifiying issues in estimation of parameter X propagating to issues in estimation of parameter Y.

stefanradev93:
I'd second that!

paul-buerkner:
Implemented in #765

PR #765, opened 2026-09-06
feat(diagnostics): add pairs recovery plot

emecii:
Draft implementation proposal for #757. The requested matrix is implemented; the API choices below are proposals, not previously agreed decisions.

### Behavior

Add `bayesflow.diagnostics.pairs_recovery(estimates, targets)` with one point per simulated dataset:

- Diagonal: ground truth versus posterior point estimate, uncertainty intervals and identity line.
- Upper triangle: pairs of posterior point estimates.
- Lower triangle: pairs of ground truths.

For row i and column j, x is variable j and y is variable i. Off-diagonal estimates describe variation across datasets, not within-dataset posterior dependence. Dictionary selection, derived quantities and point/uncertainty aggregators use existing helpers; the default point summary is the median.

```python
import numpy as np
from bayesflow.diagnostics import pairs_recovery

rng = np.random.default_rng(757)
targets = rng.normal(size=(40, 3))
estimates = targets[:, None, :] + rng.normal(scale=0.3, size=(40, 100, 3))
fig = pairs_recovery(estimates, targets, variable_names=["alpha", "beta", "gamma"])
fig.savefig("pairs-recovery.png")
```

### API choices for discussion

- Return a Matplotlib `Figure`, as `recovery` does, rather than a Seaborn `PairGrid`. Each row/column mixes data sources, so axes are independent; the axes are available in row-major order for customization.
- Apply `add_corr=True` only on diagonal recovery panels, matching `recovery`. Off-diagonal correlation annotations are not added.
- Use square axes and constrained layout, with each cell labeled by variable and source. Diagonal limits include uncertainty endpoints. Styling uses explicit arguments; arbitrary PairGrid kwargs are not forwarded.

Would these choices fit the intended interface, or should the return type/correlation options follow another convention? The draft stays open for that approach discussion.

### Validation

All 36 existing plotting tests plus 19 new tests pass (55 total); the new tests also pass under JAX. Tests inspect actual Matplotlib artists for every panel's coordinates/orientation, uncertainty geometry, labels, singleton grids, dictionary/derived-quantity selection, and input validation. Deliberate substitutions of wrong data, wrong variables, clipped uncertainty limits and shared scales each fail assertions.

The example was rendered and visually inspected. Ruff checks pass for the four changed files. Full tox/OS/backend matrix, TensorFlow and documentation build were not run. No adapter, dependency or existing plotting API changes.

Prepared with Codex assistance. No prior maintainer approval of this implementation is asserted.

### CI follow-up

The macOS Torch job currently fails in the existing ActNorm numerical-Jacobian test at `keras.ops.logdet` → `torch.logdet`, with a Metal compiler `unsupported deferred-static-alloca-size` error ([job log](https://github.com/bayesflow-org/bayesflow/actions/runs/34016608740/job/101441361633)). The same failure appears on the separate NumPy inverse-softplus PR #764. The traceback does not involve the plotting changes, but this runner failure has not been reproduced locally, so its cause remains unconfirmed. No tests have been skipped or changed to hide the failure.


.prototype/linear-regression-pairs-recovery.png
paul-buerkner, line 1: Should be removed before merging

tests/test_diagnostics/test_pairs_recovery.py
paul-buerkner, line 1: This test is located in a separate file rather than in tests_diagnostics_plots. Given the extend of the tests, I guess it makes sense but it still a little inconsistent. @stefanradev93 thoughts on this?

bayesflow/diagnostics/plots/pairs_recovery.py
```diff
@@ -0,0 +1,185 @@
...
+        If variable names do not match the selected variables, errors are
+        negative, or uncertainty bounds do not enclose the point estimates.
+
+    Notes
+    -----
+    Pairwise estimates describe variation of posterior point summaries across
+    datasets, not posterior dependence within a dataset. Posterior summaries
+    are heuristic and can be misleading for multimodal distributions.
+    """
+    if test_quantities is not None:
```
paul-buerkner, line 117: Looks like a lot of custom code for validating inputs is used. Can you please reuse the validation code in plots/recovery.py? It's using the same data after all. Or is there a reason all of this custom validation code is needed?

paul-buerkner (changes requested):
Plot looks good now but the code requires a bit of cleaning before we can merge this PR.

.prototype/linear-regression-pairs-recovery.png
emecii: Removed in e914d43c.

bayesflow/diagnostics/plots/pairs_recovery.py
emecii: Refactored this to use `prepare_plot_data`, the same preprocessing and validation path used by `recovery`, in e914d43c.

bayesflow/diagnostics/plots/pairs_recovery.py
```diff
@@ -0,0 +1,191 @@
...
+        if not isinstance(estimates, Mapping) or not isinstance(targets, Mapping):
```
paul-buerkner: If you compare the input checks done here with the input checks done in recovery you notice there is still quite a bit of more input checks here. Are they all necessary? And if they are should we somehow bundle them perhaps to resuse them also in other plots in the future?

paul-buerkner (commented):
Getting there, thank you! I still have questions about all the input checks

bayesflow/diagnostics/plots/pairs_recovery.py
emecii: Agreed. I removed the pairs-local mapping/name checks by moving that normalization into `compute_test_quantities`, and extracted the posterior-summary/error-bar contract into `compute_recovery_estimates`, which is now used by both `recovery` and `pairs_recovery`. The remaining checks cover the shared 3D posterior-draw and aggregator-output contract; only the square layout remains pairs-specific. I also parametrized the aggregate/error validation tests across both plots. Verified with Ruff and the full diagnostics suite (85 passed).

bayesflow/diagnostics/plots/pairs_recovery.py
```diff
@@ -0,0 +1,182 @@
...
+        estimates=estimates,
+        targets=targets,
+        variable_keys=variable_keys,
+        variable_names=variable_names,
+        pairwise=True,
+    )
+    estimates, targets = plot_data["estimates"], plot_data["targets"]
+    variable_names = plot_data["variable_names"]
+    try:
+        points, errors = compute_recovery_estimates(
```
paul-buerkner: I am sorry about being nitpicky about this, but for consistency with the other "prepare" functions, I would like compute_recovery_estimates to return a dict with two elements, not two elements directly. 

(Perhaps it would be better also for prepare_plot_data and friends to return several elements directly, but this is not a change I would like to be made in this PR.)

paul-buerkner (changes requested):
Thank you! Almost done :-)

bayesflow/diagnostics/plots/pairs_recovery.py
emecii: Updated in 15b6e29c: `compute_recovery_estimates` now returns a dict with `point_estimates` and `uncertainty` entries. Both recovery plot callers use the named entries, and the return contract is documented and covered by a regression test. Verified with the focused test file, the full diagnostics suite (86 passed), and Ruff.

paul-buerkner (approved):
Excellent, thank you! Will merge as soon as I can confirm tests passing locally for me too

PR #770, opened 2026-09-11
Fix node expansion in GraphicalSimulator

daniel-habermann:
The new factorization enumeration exposed a bug in the graph expansion algorithm that previously did not matter: For the three-level example model (schools -> classrooms -> students), the expanded graph only contains a single student within each classroom, that is, classroom_1 -> student_1 and classroom_2 -> student_2.

For the posterior factorization, that means that students are exchangeable even when not conditioning on the classroom parameters. This situation could not arise with the previous inverse Stuhlmüller default, but since we are now enumerating all possible node orderings, each expanded node must always include two copies of each child, otherwise we get an incorrect posterior factorization.

This PR ensures that graph expansion always fully expands all nodes, so that classroom_1 has two children student_11 and student_12 and classroom_2 has two children student_21 and student_22. By doing so, graph inversion can then factorize the posterior correctly, independent of the node ordering.

### DIFF
--- a/bayesflow/experimental/graphical_approximator/example_approximators/three_level_approximator.py
+++ b/bayesflow/experimental/graphical_approximator/example_approximators/three_level_approximator.py
@@ -16,6 +16,7 @@ def three_level_approximator():
         DeepSet(summary_dim=30),
         DeepSet(summary_dim=40),
         DeepSet(summary_dim=50),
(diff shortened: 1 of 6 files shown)


paul-buerkner (approved):
Thanks. Fix looks good to me. Tests are failing for unrelated reasons, so I think you can merge this.

PR #781, opened 2026-09-13
Single legend and optional loss components in loss plot

LarsKue:
Builds on #780. Once that is merged, rebase with:

```bash
git fetch origin dev
git rebase --onto origin/dev fix/loss-plot-columns fix/loss-plot-single-legend
git push --force-with-lease
```

Keras' `History` stores one scalar per epoch per key, so `history.history["loss"]` is always one-dimensional; loss components (e.g. `layer_loss`) live under separate keys. `loss` now plots only the total loss by default. With `show_components=True`, every other non-`val_` key gets its own panel below the total, labeled by its title-cased key (overridable via `component_names`), with the matching `val_<key>` curve overlaid if present. A non-1-D loss history uses its first column.

The panels share one figure legend below the plot instead of one per panel; `tight_layout` reserves space for it, and on figures narrower than 12 inches the legend wraps to two columns.

Before (two-column history, no longer supported):

legend_before

After, default (`loss` only):

legend_after_default

After, `show_components=True` (`loss` + `layer_loss`):

legend_after_components

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014rsUdKsZMN4xrrJ4CWPy3X


stefanradev93:
bump

